PR #5, opened 2015-03-24
Add endpoint tests

kitcambridge:
Other small fixes...

* Use the current time if the `version` param is omitted, not a number, or negative to match the Go server behavior.
* Pass along the `node_id` for the "client busy" case, not the "routing failure" case.
* Handle empty `node_id` fields in `_process_jumped_client`.

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -1,0 +1,486 @@
+import json
+import functools
+
+from cyclone.web import Application
+import twisted.internet.base
+from twisted.internet.defer import Deferred
+from txstatsd.metrics.metrics import Metrics
+from cryptography.fernet import Fernet, InvalidToken
+import requests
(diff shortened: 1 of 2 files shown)


autopush/tests/test_endpoint.py
```diff
@@ -0,0 +1,486 @@
...
+from cyclone.web import Application
+import twisted.internet.base
+from twisted.internet.defer import Deferred
+from txstatsd.metrics.metrics import Metrics
+from cryptography.fernet import Fernet, InvalidToken
+import requests
+from boto.dynamodb2.exceptions import (
+    ProvisionedThroughputExceededException,
+)
+
```
bbangert: No space here. Also, imports should be alphabetical in order, and imports/from's should be grouped.

bbangert:
r+

PR #8, opened 2015-03-24
Add optional CORS support to Autopush

jrconlin:
This is a late fix request from releng. CORS support is only required for DEV and STAGE environments (thus why it defaults to "off")

Operations:
Note that CORS is only enabled if you run `autoendpoint --cors` or with environment `ALLOW_CORS`

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -43,19 +43,11 @@ class EndpointHandler(cyclone.web.RequestHandler):
         self.version = version
         self.data = data
 
-    @cyclone.web.asynchronous
     def options(self, token):
         self._addCors()
-        self.set_status(200)
-        self.write("")
-        self.finish()
 
-    @cyclone.web.asynchronous
     def head(self, token):
         self._addCors()
-        self.set_status(200)
-        self.write("")
-        self.finish()
 
     @cyclone.web.asynchronous
     def put(self, token):
@@ -223,11 +215,8 @@ class EndpointHandler(cyclone.web.RequestHandler):
 
     def write_error(self, code, exception=None):
         """ Write the error (otherwise unhandled exception when dealing with
-        unknown protocol specifications.) """
-        reason = "No reason"
-        if exception is not None and exception.reason is not None:
-            reason = exception.reason
-        error = "%d: %s" % (code, reason)
+        unknown method specifications.) """
         self.set_status(code)
-        log.err("Endpoint write_error: %s" % error)
+        if exception is not None:
+            log.err(exception)
         self.finish()
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -14,7 +14,7 @@ from twisted.python import log
 
 class EndpointHandler(cyclone.web.RequestHandler):
     def initialize(self):
-        self.metrics = self.settings.metrics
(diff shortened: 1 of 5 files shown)


autopush/endpoint.py
```diff
@@ -200,3 +220,14 @@ def _finish_missed_store(self, result=None):
...
+
+    def write_error(self, code, exception=None):
+        """ Write the error (otherwise unhandled exception when dealing with
+        unknown protocol specifications.) """
+        reason = "No reason"
+        if exception is not None and exception.reason is not None:
+            reason = exception.reason
+        error = "%d: %s" % (code, reason)
+        self.set_status(code)
+        log.err("Endpoint write_error: %s" % error)
```
bbangert: log.err can take a plain exception, it'll properly wrap it, and then raven will properly extract all the exception details for reporting

autopush/endpoint.py
```diff
@@ -43,11 +44,26 @@ def _load_params(self):
         self.data = data
 
     @cyclone.web.asynchronous
+    def options(self, token):
+        self._addCors()
+        self.set_status(200)
+        self.write("")
+        self.finish()
```
bbangert: These don't need to be asynchronous since when your function returns, its actually done. So the decorator can be dropped, along with the calls to write and finish, and set_status (200 is default, finish isn't needed without asynchronous, and I don't believe you need to write anything).

autopush/settings.py
```diff
@@ -74,9 +75,19 @@ def __init__(self,
...
             if key == "crypto_key":
                 self.fernet = Fernet(val)
             else:
                 setattr(self, key, val)
+
+    def get(self, key):
+        """ cyclone/web.py is calling self.settings.get("debug") where
+        self == EndpointHandler. this is causing an unhandled exception.
+        """
+        import pdb; pdb.set_trace()
```
bbangert: This shouldn't be stuck in here.

autopush/tests/test_endpoint.py
```diff
@@ -452,6 +455,32 @@ def test_process_jumped_client_error(self):
...
+        endpoint.head(None)
+        eq_(endpoint._headers[acrm], "*")
+
+    def test_cors_options(self):
+        acrm = "Access-Control-Request-Method"
+        endpoint = self.endpoint
+        endpoint.settings.cors = True
+        endpoint.options(None)
+        eq_(endpoint._headers[acrm], "*")
+
```
bbangert: There's a new write_error, a test should trigger the error writing.

autopush/settings.py
```diff
@@ -74,9 +75,18 @@ def __init__(self,
...
     def update(self, **kwargs):
         for key, val in kwargs.items():
             if key == "crypto_key":
                 self.fernet = Fernet(val)
             else:
                 setattr(self, key, val)
+
+    def get(self, key):
+        """ cyclone/web.py is calling self.settings.get("debug") where
+        self == EndpointHandler. this is causing an unhandled exception.
```
bbangert: It is? If cyclone already attaches its own 'settings' object to the handler, we'll need to rename ours to keep them separate.

bbangert:
r+

PR #10, opened 2015-03-25
Add some sane protocol defaults.

bbangert:


jrconlin:
r+ (particularly setting FailByDrop=False)

PR #12, opened 2015-03-25
Add metrics to provision errors.

### LINKED ISSUE #2: Add provisioned error metrics.
Several parts of the code catch and retry provisioned throughput exceptions. These parts of code should emit metrics indicating that along with context.

i.e. 'error.provisioned.new_connection', 'error.provisioned.store_notification'.

bbangert:
Closes #2.

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -57,13 +57,18 @@ def get_storage_table():
 
 
 class Storage(object):
-    def __init__(self, table):
+    def __init__(self, table, metrics):
         self.table = table
+        self.metrics = metrics
 
     def fetch_notifications(self, uaid):
-        notifs = self.table.query_2(consistent=True, uaid__eq=uaid,
-                                    chid__gt=" ")
-        return list(notifs)
+        try:
+            notifs = self.table.query_2(consistent=True, uaid__eq=uaid,
+                                        chid__gt=" ")
+            return list(notifs)
+        except ProvisionedThroughputExceededException:
+            self.metrics.increment("error.provisioned.fetch_notifications")
+            raise
 
     def save_notification(self, uaid, chid, version):
         conn = self.table.connection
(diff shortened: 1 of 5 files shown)


autopush/db.py
```diff
@@ -57,13 +57,18 @@ def get_storage_table():
...
 
     def fetch_notifications(self, uaid):
-        notifs = self.table.query_2(consistent=True, uaid__eq=uaid,
-                                    chid__gt=" ")
-        return list(notifs)
+        try:
+            notifs = self.table.query_2(consistent=True, uaid__eq=uaid,
+                                        chid__gt=" ")
+            return list(notifs)
+        except ProvisionedThroughputExceededException:
```
jrconlin: I'm glad we're noting exceptions like ProvisionedThroughput, but there are other exceptions we might want to be aware of (ResourceInUse, ItemCollectionSizeLimitExceeded, etc.) Do we know if there's enough info in these to analyze the cause if they're handled by the default catch mechanism?

autopush/tests/test_db.py
```diff
@@ -11,40 +10,37 @@
...
+
+class StorageTestCase(unittest.TestCase):
+    def setUp(self):
+        table = get_storage_table()
+        self.real_table = table
+        self.real_connection = table.connection
 
-        get_storage_table()
+    def tearDown(self):
+        self.real_table.connection = self.real_connection
```
jrconlin: If table could be corrupted and needs to be recovered, does it make sense to have self.real_table just do a copy of table and store that?

bbangert: So, only the connection needs to be restored from my experiments, because mock_dynamodb2 will otherwise retain the one I messed up. The table is created fresh each instantiation, but mock_dynamodb2 retains the connection and re-assigns it to new table objects.

autopush/db.py
bbangert: I went through for each function and looked at what the failure handler does. And for other exceptions, its going to return a 503 and ask the user to try again, which seems like the appropriate response. The sentry logger should log the context of the exception with details about it.

autopush/tests/test_db.py
jrconlin: Fair, it's in a unit test, so I was curious why you're restoring something anyway. Good to know for mine, thanks!

Issue #13, opened 2015-03-25
Cyclone handlers in websocket.py overloads cyclone's settings


bbangert:
@jrconlin already fixed the endpoint to use ''ap_settings'', this should be fixed in websocket.py's handler as well.

bbangert:
Closed by #13.

PR #14, opened 2015-03-26
Fix for #13 websocket routing and notif handlers

jrconlin:
* These also need to use ap_settings to avoid property stomping

@bbangert r?


bbangert:
r+

PR #15, opened 2015-03-26
Add table naming and provisioning throughput options.

### LINKED ISSUE #4: Make storage/router DynamoDB table names configurable
The router/storage table names used must be configurable, as 'router' and 'storage' could cause name conflicts.

bbangert:
Closes #4.

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -10,10 +10,12 @@ from boto.dynamodb2.table import Table
 from boto.dynamodb2.types import NUMBER
 
 
-def create_router_table():
-    return Table.create("router",
+def create_router_table(tablename="router", read_throughput=5,
+                        write_throughput=5):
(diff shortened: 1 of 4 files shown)


jrconlin:
r+ (mostly)

I'm ok with using the env/CLI approach to app configuration, but it makes for some pretty unwieldy starts. I don't think we'll hit 262K command line length any time soon, and it's certainly possibly to offload to the environment. I worry that there might be a risk of op error. We may want to check with Ops to see if there's a more preferred means of specifying configuration.

kitcambridge:
`configargparse` can read from a config file too, right?

bbangert:
That's correct, we could add a config file option, and configargparse can use that as well.

jrconlin:
Cool, concern withdrawn.

PR #18, opened 2015-03-31
Feature/pping

### LINKED ISSUE #16: Add GCM and APNS Bridges

jrconlin:
Closes #16, sort of.

This pull has the basic requirements for using 3rd party bridges like GCM and APNS. However, while constructing these bridges, I realized that it would be far more efficient to have these as separate queue processors, so incoming requests would get handed off to a batch of machines who's sole job is to feed and maintain the notifications for APNS and GCM. (Ideally, these could be spun up or down independent on real load.)

These will require some design work to most efficiently deal with this, so this PR contains the same level of completion as the pushgo interfaces.

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -16,7 +16,6 @@ from twisted.python import log
 def MakeEndPoint(fernet, endpoint_url, uaid, chid):
     """ Create an endpoint (used both by websocket handler and
     the /register endpoint """
-    import pdb;pdb.set_trace()
     token = fernet.encrypt((uaid + ":" + chid).encode('utf8'))
     return endpoint_url + "/push/" + token
 
@@ -110,24 +109,47 @@ class EndpointHandler(cyclone.web.RequestHandler):
             self.write("Invalid")
             return self.finish()
 
+        # Is there a proprietary ping associated with this uaid?
+        pping = result.get("proprietary_ping")
+        if pping is not None:
+            d = deferToThread(
+                self.pinger.ping,
+                self.uaid,
+                self.version,
+                self.data,
+                pping)
+            d.addCallback(self._process_pping, result)
+            d.addErrback(self._error_response)
+            return
+        self._process_route(result)
+
+    def _process_pping(self, result, routeinfo):
+        if not result:
+            log.msg("proprietary ping failed, falling back to routing")
+            return self._process_route(routeinfo)
+        # Ping handoff succeeded, no further action required
+        self.metrics.increment("router.pping.hit")
+        self.write("Success")
+        self.finish()
+
+    def _process_route(self, routeinfo):
         # Determine if they're connected at the moment
-        node_id = result.get("node_id")
 
         # Indicator if we got a node_id, but the node won't handle
         # delivery at this moment later.
         self.client_check = False
 
-        if node_id:
+        if routeinfo.get("node_id"):
             # Attempt a delivery if they are connected
             payload = json.dumps([{"channelID": self.chid,
                                    "version": self.version,
                                    "data": self.data}])
             d = deferToThread(
                 self.ap_settings.requests.put,
-                node_id + "/push/" + self.uaid,
+                routeinfo.get("node_id") + "/push/" + self.uaid,
                 data=payload
             )
-            d.addCallback(self._process_routing, result)
+            d.addCallback(self._process_routing, routeinfo)
             d.addErrback(self._error_response)
         else:
             self._save_notification()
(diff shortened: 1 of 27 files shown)


autopush/endpoint.py
```diff
@@ -12,6 +13,13 @@
 from twisted.python import log
 
 
+def MakeEndPoint(fernet, endpoint_url, uaid, chid):
```
kitcambridge: What do you think about making this a method on `AutopushSettings`, since all callers use `ap_settings.fernet` and `ap_settings.endpoint_url`?

autopush/endpoint.py
```diff
@@ -220,3 +256,115 @@ def write_error(self, code, exception=None):
...
+        self.set_status(500)
+        self.write("Error processing request")
+        self.finish()
+
+    def _load_params(self):
+        tags = {'chid': 'channelid',
+                'conn': 'connect',
+                }
+        chid = conn = None
+        if len(self.request.body) > 0:
```
kitcambridge: `EndpointHandler` uses this pattern, too...I wonder if it's worth breaking out into a generic "parse body with defaults" function.

autopush/endpoint.py
jrconlin: That it's a pretty good idea, reduces the argument list, and makes things lemony fresh?

autopush/endpoint.py
jrconlin: I dunno, there's a lot in common, but subtle differences (e.g. Endpoint does int conversion and checking on "version") Not saying that it's not impossible to build a generic handler, but I think we'd be introducing more complexity than it's probably worth.

autopush/endpoint.py
```diff
@@ -220,3 +256,115 @@ def write_error(self, code, exception=None):
...
+        self._registered(True)
+
+    @cyclone.web.asynchronous
+    def put(self, uaid=None):
+        self.metrics = self.ap_settings.metrics
+        self.start_time = time.time()
+
+        self.add_header("Content-Type", "application/json")
+
+        if uaid is None:
```
kitcambridge: What if we included the ID in the body, instead of the URL—and just made the URL `/register`? It seems a bit odd to make a request to `/register/abc` but receive a response with `"useragentid": "xyz"`.

autopush/endpoint.py
```diff
@@ -220,3 +256,115 @@ def write_error(self, code, exception=None):
...
+        else:
+            try:
+                uuid.UUID(uaid)
+            except ValueError:
+                log.msg("Invalid UAID [%s], swapping for valid one" % uaid)
+                uaid = str(uuid.uuid4())
+
+        self.uaid = uaid
+        if not self._load_params():
+            log.msg("Invalid parameters")
```
kitcambridge: Will Twisted's error logger trigger Sentry?

autopush/endpoint.py
jrconlin: Hmm. let me think about this. I agree that specifying a disposed uaid is odd, but I think there was goofiness in cyclone handling both a /register/ and /register/abc... url, but that may just need regex goofing.

autopush/pinger/apns_ping.py
```diff
@@ -0,0 +1,42 @@
...
+from twisted.python import log
+import apns
+
+
+# https://github.com/djacobs/PyAPNs
+class APNSPinger(object):
+    apns = None
+
+    def __init__(self, config):
+        self.apns = apns.APNs(use_sandbox=config.get("sandbox", False),
```
kitcambridge: Should these live in `settings`?

kitcambridge:
This looks great, @jrconlin! Just noted a few pedantic nits. :wink:

autopush/endpoint.py
```diff
@@ -220,3 +256,115 @@ def write_error(self, code, exception=None):
...
+
+    def options(self, token):
+        self._addCors()
+
+    def head(self, token):
+        self._addCors()
+
+    def _addCors(self):
+        if self.ap_settings.cors:
+            self.set_header("Access-Control-Request-Method", "*")
```
kitcambridge: This doesn't work as-is, sadly. I think the minimum required header set is:

```python
self.set_header("Access-Control-Allow-Origin", "*")
self.set_header("Access-Control-Allow-Methods", "PUT")
```

autopush/endpoint.py
jrconlin: Ok, we need to have `PUT /register/uaid` because clients will need to send new Connect info. If the UAID is invalid (because it's bogus) we'll re-issue the UAID. So, most of the time, the UAID shouldn't change. It's also RESTful to say that you're PUTting to a resource handle, so in this case we're PUTing to the `/register/uaid`.

That said, I should allow for now UAID for PUT for initial connections, so will update to handle that.

autopush/endpoint.py
bbangert: Yup, my sentry addition tacked on a log observer to twisted's.

bbangert:
This does add a dozen or two untested lines, though not as much as originally. I'm still trying to determine how that makes me feel.

jrconlin:
Well, 11 newly untested lines according to the latest coveralls. The four in endpoint.py are probably the more slightly interesting ones, although that was just rolling some repeated code into a func. I'm a bit surprised that they weren't triggered from other test cases, but it's easy enough to fix.

bbangert:
cool beans

autopush/endpoint.py
kitcambridge: *[/me nods](https://www.youtube.com/watch?v=wlMwc1c0HRQ)*

kitcambridge:
I think this is about ready...just fix up the CORS headers, and we can merge it in. :shipit: Your call on whether to add pinger configs to `AutopushSettings`, or leave it as-is. I think it's handy to have a centralized location for defaults, but that's a personal preference. Foolish consistency and all...

jrconlin:
Applied fixes for CORs headers. Going to leave the ppinger stuff separate since it's more a plugin than a component, and how that bit works may alter in the future.

PR #23, opened 2015-04-03
Use Twisted's HTTP client for internal routing

### LINKED ISSUE #21: Switch internal routing from requests lib to twisted's http client
Two blocking operations hit the thread-pool in the endpoint, decryption, and internal delivery to a connection node. We should switch to using twisted's http client to reduce the use of the thread-pool.

http://twisted.readthedocs.org/en/latest/web/howto/client.html

kitcambridge:
Closes #21.


autopush/endpoint.py
```diff
@@ -110,29 +113,38 @@ def _process_uaid(self, result):
...
-    def _process_routing(self, result, item):
+    def _send_notification(self, node_id):
+        payload = json.dumps([{"channelID": self.chid,
+                               "version": self.version,
+                               "data": self.data}])
+        return self.ap_settings.agent.request(
+            "PUT",
+            node_id + "/push/" + self.uaid,
+            bodyProducer=FileBodyProducer(StringIO(payload)),
+        ).addCallback(IgnoreBody.ignore)
```
jrconlin: No error handler?

kitcambridge: Not quite...if there's a request error, `IgnoreBody.ignore` won't be called and the error will be forwarded to the next registered errback (`self._error_response`), just like JavaScript promises. I can add a test.

autopush/main.py
```diff
@@ -196,6 +196,7 @@ def make_settings(args, **kwargs):
             pingConf = None
 
     return AutopushSettings(
+        reactor,
```
bbangert: reactor is a global, it doesn't need to be passed around.

kitcambridge: Updated. I passed it around per the [Twisted docs](https://twistedmatrix.com/documents/current/api/twisted.internet.reactor.html)...

> New application code should prefer to pass and accept the reactor as a parameter where it is needed, rather than relying on being able to import this module to get a reference.

But we're not using the reactor in the tests.

PR #25, opened 2015-04-03
Adds sample configs

### LINKED ISSUE #22: Example stage/prod configs.
I'll need example stage and prod configs. If they are close to the same, one will suffice.

kitcambridge:
Closes #22.

I also added separate `router_ssl_{key, cert}` options, since I think we'll want to expose a TLS listener for WebSocket connections, but just a plaintext TCP listener for internal routing requests. Access to the routing port can be restricted via security groups.

@oremj r?


shared_settings.ini
```diff
@@ -0,0 +1,48 @@
+; Settings shared by the connection and endpoint nodes. Overrides
+; should be specified in `autopush_settings.ini` and
+; `autoendpoint_settings.ini`.
+
+; Enable verbose logging for WebSocket connections and HTTP requests.
+; This should be disabled in production.
+debug = false
+
+; The public hostname. Defaults to the system hostname if unspecified.
+hostname =
```
oremj: Will this pull the ec2 hostname, like simplepush?

autopush_settings.ini
```diff
@@ -0,0 +1,17 @@
...
+; Settings for the internal routing listener, used to deliver routed
+; notifications to clients connected to this node.
+router_hostname =
+router_port = 8081
+router_ssl_key =
+router_ssl_cert =
+
+; The endpoint hostname and port, used to construct the push endpoint
+; URL for each registered channel.
+endpoint_hostname =
```
oremj: Can I get an example hostname.

autopush_settings.ini
```diff
@@ -0,0 +1,17 @@
+; Connection node settings.
+
+[connection]
+; The port on which to listen for WebSocket connections.
+port = 8080
+
+; Settings for the internal routing listener, used to deliver routed
+; notifications to clients connected to this node.
+router_hostname =
```
oremj: Example hostname here too.

shared_settings.ini
kitcambridge: This just uses `socket.gethostname()`...so I think that'll return the internal hostname? I can clarify.

oremj:
I just need clarification about what the different host names are referring to. Otherwise, lgtm.

shared_settings.ini
```diff
@@ -0,0 +1,48 @@
...
+; exposed by the connection node uses separate keys, specified in the
+; `[connection]` section. If omitted, TLS will not be used.
+ssl_key =
+ssl_cert =
+
+; The push endpoint token encryption key. This should be identical for
+; connection and endpoint nodes in the same cluster; otherwise,
+; endpoints won't be able to decrypt tokens generated by the
+; connection node.
+crypto_key = i_CYcNKa2YXrF_7V1Y-2MFfoEl7b6KX55y_9uvOKfJQ=
```
oremj: Need docs on how to generate a key.

configs/autopush_shared.ini
```diff
@@ -0,0 +1,43 @@
+; Settings shared by the connection and endpoint nodes.
+
+; Enable verbose logging for WebSocket connections and HTTP requests.
+; This should be disabled in production.
+debug = false
```
oremj: autopush: error: debug is a flag but is being set to 'false'

Getting that with `debug = false`.

PR #26, opened 2015-04-06
Note required PyPy version; add OS X instructions

kitcambridge:
Mostly for selfish reasons, since @jrconlin and @bbangert both use Linux...

* `cffi@0.9.2` requires PyPy >= 2.5.1.
* OS X doesn't ship with `wget` by default, but `curl -O` is equivalent for our purposes. :crying_cat_face:
* `ARCHFLAGS`, `LDFLAGS`, and `CFLAGS` should be set, to avoid linking against the ancient system-bundled OpenSSL.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ BUILD_DIRS = bin build deps include lib lib64 lib_pypy lib-python site-packages
 all:	build
 
 $(BIN)/pip: $(BIN)/pypy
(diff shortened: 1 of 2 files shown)


kitcambridge:
Not sure what happened with Travis...`DirtyReactorAggregateError: Reactor was unclean` doesn't look good, but this didn't change the code at all.

PR #30, opened 2015-04-06
Smoke test fixes

kitcambridge:
* Fix missing `settings` attribute errors.
* Fix encoding errors by converting URLs to bytes.
* Only respond with a 202 for routing misses.

/me needs to run the smoke tests next time...

@bbangert r?


jrconlin:
I'm curious what triggered the error for this. Since tests were passing, do we need to check for that?

kitcambridge:
I hadn't run the smoke tests in a while. Maybe it makes sense to run them on Travis, too?

bbangert:
A test that results in a unicode string requiring this fix should be added, so we don't accidentally break it this way again.

Issue #20, opened 2015-04-01
Log the websocket user-agent


bbangert:

bbangert:
Fixed by #31.

PR #31, opened 2015-04-07
Feature/logging for ops and ua

### LINKED ISSUE #11: Ensure logging stdout matches ops logging format

bbangert:
Fixes #11, #20.


autopush/logging.py
```diff
@@ -0,0 +1,85 @@
...
+    def start(self):
+        """Start capturing Twisted logs."""
+        startLoggingWithObserver(self, setStdout=False)
+
+
+def stdout(message):
+    msg = {}
+    ts = message.pop("timestamp")
+    del message["task_level"]
+    msg["Hostname"] = HOSTNAME
```
kitcambridge: Could we use `AutopushSettings.hostname` here?

autopush/main.py
```diff
@@ -254,8 +238,11 @@ def connection_main(sysargs=None):
     # Internal HTTP notification router
     site = cyclone.web.Application([
         (r"/push/([^\/]+)", r),
-        (r"/notif/([^\/]+)", n),
-    ], default_host=settings.router_hostname)
+        (r"/notif/([^\/]+)", n)
+    ],
+        default_host=settings.router_hostname, debug=args.debug,
+        log_function=skip_request_logging
```
kitcambridge: Is this because Cyclone's default logger writes to standard out?

autopush/endpoint.py
```diff
@@ -253,6 +253,7 @@ def _finish_missed_store(self, result=None):
         # since we're handing off, return 202
         self.set_status(202)
         self.write("Success")
+        self.set_status(202)
```
kitcambridge: I think we had a mid-air collision in #30. :smile_cat:

autopush/logging.py
bbangert: Not sure, hostname is the host to bind to, but might not be the preferred name so I went with fqdn.

autopush/main.py
bbangert: It's mainly because cyclone wants to log every http request, which strikes me as dumb, and the only way to disable it was to null op the log function. If we want to log something, we can with twisted's log function.

PR #37, opened 2015-04-07
Datadog features

oremj:
Adding the `host` argument to each metric allows it to aggregate the metric with everything else it knows about that host.

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -27,22 +27,30 @@ class TwistedMetrics(object):
 
 
 class DatadogMetrics(object):
-    def __init__(self, api_key, app_key, flush_interval=10):
+    def __init__(self, api_key, app_key, flush_interval=10,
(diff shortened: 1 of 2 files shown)


oremj:
@bbangert r?

autopush/metrics.py
```diff
@@ -26,20 +27,30 @@ def timing(self, name, duration):
...
+                 namespace="autopush"):
+
         datadog.initialize(api_key=api_key, app_key=app_key)
         self._client = ThreadStats()
         self._flush_interval = flush_interval
+        self._host = get_hostname()
+        self._namespace = namespace
+
+    def _prefix_name(self, name):
+        return "%s.%s" % (self.namespace, name)
```
bbangert: Should be self._namespace

bbangert:
Looks fine, waiting for travis.

PR #38, opened 2015-04-07
Feature/producer pausing

### LINKED ISSUE #28: Register a producer for the websocket protocol for more intelligent pausing

bbangert:
Closes #28.

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -9,8 +9,10 @@ from twisted.internet.defer import (
     DeferredList,
     CancelledError
 )
+from twisted.internet.interfaces import IProducer
 from twisted.internet.threads import deferToThread
 from twisted.python import log
+from zope.interface import implements
 
 
 def ms_time():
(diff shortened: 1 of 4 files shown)


autopush/websocket.py
```diff
@@ -219,6 +241,11 @@ def process_notifications(self, tries=0):
         if self._should_stop:
             return
 
+        # Are we paused? Try again later.
+        if self.paused:
+            reactor.callLater(1, self.process_notifications)
```
kitcambridge: Could we call `process_notifications` in `resumeProducing`? (Setting short-lived timers for polling variables like this gets expensive in JavaScript, but I've seen this used elsewhere in the code. I assume this is the Pythonic way to do it?)

bbangert: It's possible that process_notifications was already scheduled, so that wouldn't reduce the need for this check.

autopush/websocket.py
```diff
@@ -351,55 +390,44 @@ def force_delete(self, result, chid):
...
-                continue
+        # If its a direct update, remove it and return
+        if self.direct_updates.get(chid) == version:
+            del self.direct_updates[chid]
+            return
 
-            if skip:
-                continue
+        # Remove the update if version matches
+        if self.updates_sent.get(chid) == version:
```
kitcambridge: Maybe `<= version`, if the client is coalescing acknowledgements?

autopush/websocket.py
```diff
@@ -351,55 +390,44 @@ def force_delete(self, result, chid):
...
+            del self.updates_sent[chid]
+        else:
+            return
 
-            # Attempt to delete this notification from storage
-            storage = self.ap_settings.storage
+        # If we ack'd a notification that wasn't direct, delete it
+        d = deferToThread(self.ap_settings.storage.delete_notification,
+                          self.uaid, chid, version)
+        d.addCallback(self.check_ack, self.uaid, chid, version)
```
kitcambridge: Will this retry indefinitely if `delete_notification` returns a `ConditionalCheckFailedException`?

bbangert: Nope, only an uncaught exception by the function results in the retry. If it fails the conditional check, we move on, since that means it was already deleted or is newer than what we're trying to delete.

autopush/websocket.py
```diff
@@ -23,13 +25,33 @@ def periodic_reporter(settings):
 
 
 class SimplePushServerProtocol(WebSocketServerProtocol):
+    implements(IProducer)
```
kitcambridge: So it's both a protocol and a push producer. Neat!

PR #39, opened 2015-04-08
Feature/redeliver disconnected

### LINKED ISSUE #36: Bug: Send un-ackd direct delivery notifications to storage
If a notification is directly delivered but not ack'd, and the client drops, we drop the notification entirely. Per the todo in websocket.py:

        # TODO: Any notifications directly delivered but not ack'd need
        # to be punted to an endpoint router

We should add this code so that the connection node fires off a notification delivery to the router to redeliver these un-ack'd messages.

bbangert:
Closes #36.


kitcambridge:
Looks great; just waiting for Travis...

PR #40, opened 2015-04-08
Add unified and Moto server binaries

kitcambridge:
Closes #33.

* `autonode` starts connection and endpoint nodes on the same machine. It's identical to running `autopush` and `autoendpoint` side-by-side, but uses `--{connection, endpoint}_port` and `--{connection, endpoint}_ssl_{key, cert}` options to avoid environment variable collisions.
* `automock` starts a unified server that proxies all AWS calls to a Moto server. Using the `@mock_dynamodb2` decorator with Twisted didn't work; see gabrielfalcao/HTTPretty#239.
* Add a `RESOLVE_HOSTNAMES` option for Docker. I found this helpful testing locally.
* Allow setting Boolean flags via environment variables and config files. Setting the flag via `RESOLVE_HOSTNAMES=1` (or `RESOLVE_HOSTNAMES=`) resulted in the error, `RESOLVE_HOSTNAMES is a flag but is being set to 1`.

PR #41, opened 2015-04-09
Add an `automock` script

kitcambridge:
Alternative implementation of #40, using Supervisor instead of a unified `autonode`. This requires mounting a config file into the container, since `autopush` and `autoendpoint` rely on shared environment variables (e.g., `port`, `ssl_key`, `ssl_cert`).

### DIFF
--- a/automock/supervisord.conf
+++ b/automock/supervisord.conf
@@ -1,0 +1,16 @@
+[supervisord]
+nodaemon=true
+
+[program:moto]
+command=/home/autopush/pypy/bin/moto_server dynamodb2 -p 5000
+stdout_logfile=/var/log/moto_server.log
+
+[program:autopush]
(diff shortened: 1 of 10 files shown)


automock/supervisord.conf
```diff
@@ -0,0 +1,16 @@
+[supervisord]
+nodaemon=true
+
+[program:moto]
+command=/home/autopush/pypy/bin/moto_server dynamodb2 -p 5000
```
kitcambridge: I think we can specify `--config_{shared, connection, endpoint}` as environment variables, and have Supervisor substitute them.

bbangert:
r+

automock/supervisord.conf
kitcambridge: On second thought, since we can just mount the appropriate config files, we don't need the extra arguments.

Issue #24, opened 2015-04-03
Ensure we're using the recommended TLS settings


kitcambridge:
[The Wiki](https://wiki.mozilla.org/Security/Server_Side_TLS) has a list of recommended settings, as does the [Go server](https://github.com/mozilla-services/pushgo/blob/1.5rc1/src/github.com/mozilla-services/pushgo/simplepush/net.go).

PR #42, opened 2015-04-09
ssl: add AutopushSSLContextFactory

oremj:
Adds the ability to use a certificate chain file and turns off SSLv3.


oremj:
@bbangert r?

This custom context factory will also be needed for #24.

bbangert:
r+

PR #44, opened 2015-04-09
logging: EnvVersion -> 2.0

oremj:
EnvVersion for logs needs to be 2.0.


oremj:
@whd tells me heka decoding won't work until this is set to 2.0.

oremj:
@bbangert r?

PR #45, opened 2015-04-09
Feature/log core protocol errors

bbangert:
The Websocket Protocol only chooses to log exceptions in the methods it calls if debug is enabled. However, debug enables a *ton* of logging we don't want. This add's exception catching to the main methods the Protocol calls (the rest are deferreds that we already log errors on).


kitcambridge:
r+

PR #49, opened 2015-04-16
Add a health handler with DynamoDB connection status

### LINKED ISSUE #34: Add /status/health or /heath
The /status endpoint is a good starting point. I'd also like an endpoint that does a deeper check. Off the top of my head:
* make sure dynamodb is working

kitcambridge:
Closes #34.

I thought about reusing `preflight_check` for this...but it would be really inconvenient if folks could use up our provisioned throughput with a simple `while true; do curl https://autopush.server/health; done`.

### DIFF
--- a/autopush/health.py
+++ b/autopush/health.py
@@ -1,6 +1,66 @@
 import cyclone.web
 
+from boto.dynamodb2.exceptions import (
+    InternalServerError,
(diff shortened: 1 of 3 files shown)


jrconlin:
Looks good. r+

PR #51, opened 2015-04-27
Feature/pping

jrconlin:
* Add full reply handling for GCM/APNs 
* Add secondary index for token lookups (used by GCM & APNs)
* change name to "bridge"


autopush/bridge/apns_ping.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _connect(self):
+        self.apns = apns.APNs(use_sandbox=self.config.get("sandbox", False),
+                              cert_file=self.config.get("cert_file"),
+                              key_file=self.config.get("key_file"),
+                              enhanced=True)
+
+    def __init__(self, config, storage):
+        self.config = config
+        self.default_title = config.get("default_title", "SimplePush")
```
kitcambridge: APNS requires a tile and body for silent notifications, too?

autopush/bridge/bridge.py
```diff
@@ -0,0 +1,67 @@
...
+                log.err("Failed to register connection for %s:%s" %
+                        (uaid, connect))
+                raise BridgeFailEx("Unable to register connection")
+        except Exception, e:
+            log.err(e)
+            raise
+        return True
+
+    def ping(self, uaid, version, data, connectInfo):
+        if connectInfo is None or connectInfo is False:
```
kitcambridge: `if not connectInfo`?

autopush/bridge/gcm_ping.py
```diff
@@ -0,0 +1,95 @@
...
+    def _error(self, err, **kwargs):
+        log.err(err, **kwargs)
+        return
+
+    def _send(self, payload, iter=0):
+        try:
+            result = self.gcm.send(payload)
+            return self._result(result, iter)
+        except Exception, e:
+            self._error(e)
```
kitcambridge: Are exceptions intentionally logged twice here?

autopush/bridge/gcm_ping.py
```diff
@@ -0,0 +1,95 @@
...
+            raise
+
+    def _result(self, reply, iter=0):
+        # handle reply content
+        # acks:
+        #  for reg_id, msg_id in reply.success.items():
+        # updates
+        for old_id, new_id in reply.canonical.items():
+            self.storage.byToken('UPDATE', new_id)
```
kitcambridge: Would it make sense to use conditional updates and deletes here, in case the app server sends a batch of notifications to the same registration ID?

autopush/bridge/apns_ping.py
jrconlin: not always. https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/ApplePushService.html#//apple_ref/doc/uid/TP40008194-CH100-SW15 The title and body of the message are user displayable items and only appear if there's no content_available flag set. It may be a bug (because of the loc-* fields), but as of the latest 8.1 it works.

autopush/bridge/bridge.py
jrconlin: More trapping if other falsy values happen to show up, but collapsing to if not connectInfo should suffice.

autopush/bridge/gcm_ping.py
jrconlin: no, crap, debugging detritus.

autopush/bridge/gcm_ping.py
jrconlin: Not sure I follow. As I understand the spec, canonical.items() returns replacement tokens. The token is the one that we're using to send out, so for us, the "old_id" would always be the same for this call. (the id being the recipient process we're tied to, not the remote app that may be the eventual target.) Are you thinking about having the bridge handle multiple outbound requests? In that case, yes, we could have the byToken handle the old and new values.

kitcambridge: I'm spouting more nonsense than usual today. :smile_cat: For some reason, I thought the ID was the remote app, not the replacement token for the recipient—and we'd end up clobbering the token for concurrent requests to the same ID.

autopush/tests/test_bridge.py
```diff
@@ -0,0 +1,231 @@
...
+                         {"type": "gcm", "token": "abcd123"}))
+
+        # APNs test
+        apns.gateway_server = Mock()
+        self.f_apns.gateway_server = Mock()
+        self.f_apns.gateway_server.send_notification = Mock()
+        reply = tping.ping("uaid", 123, 'abcd',
+                           {"type": "apns", "token": "abcd123"})
+        self.assertTrue(reply)
+        # Need to find a better way to parse the second param of this
```
kitcambridge: This ought to do it... :wink:

```py
_, (_, payload, _), _ = self.f_apns.gateway_server.send_notification.mock_calls[0]
self.assertEqual(payload.custom, {'Msg': 'abcd', 'Ver': 123})
```

kitcambridge:
Looks great, @jrconlin! :tada:

autopush/bridge/apns_ping.py
```diff
@@ -0,0 +1,86 @@
...
+                return False
+            token = connectInfo.get("token")
+            if token is None:
+                return False
+            payload = apns.Payload(alert=connectInfo.get("title",
+                                                         self.default_title),
+                                   content_available=1,
+                                   custom={"Msg": data,
+                                           "Ver": version})
+            id = time.time()
```
kitcambridge: I think this should be `int(time.time())`. PyAPNs will truncate it automatically, but then `self.messages.get(err.get('identifier'))` won't work.

autopush/websocket.py
```diff
@@ -270,21 +270,26 @@ def process_hello(self, data):
...
             d.addCallback(self._check_router, True)
             d.addErrback(self.err_hello)
         else:
             self._check_router(False)
 
-    def _check_router(self, paused=False):
+    def _check_router(self, paused=False, bridge_register=None):
         if paused:
             self.transport.resumeProducing()
+        if bridge_register is not None and bridge_register is False:
```
kitcambridge: More of a philosophical question, but should we let the connection proceed even if bridge registration fails? This makes sense for some more than others: UDP wake-up is an optimization, and it's not critical if the `/netinfo/v1` endpoint is unavailable.

On the other hand, if bridge registration fails and the client can only use GCM/APNS, then it has a problem. But I think the client should decide that, and close the connection if it can't continue.

jrconlin: I've argued about this with myself a fair bit. Push can fail back to maintaining it's own socket if bridge registration fails. It's not as efficient, but it's possible. Bridges are a bit of a cost savings/efficiency thing both for the client and for the server. That said, it's certainly possible for the bridge to be broken for any number of reasons (registration failure, provider termination, etc.) All of which should probably be factored somehow.

I think that registration failure needs to be communicated back to the client, but that's also why I don't terminate the socket connection on a bridge failure. The client is certainly free to, at that point. 

The way I view it, the client is really the source of truth about how critical a given bridge really is. Granted, this is purely a subjective ruling so I'm happy to change things if there's good logic otherwise.

kitcambridge: > I think that registration failure needs to be communicated back to the client, but that's also why I don't terminate the socket connection on a bridge failure.

Agreed. Should we remove the `return` on line 289 and let the normal handshake take over, then, or wait for the client to retry?

jrconlin: My preference is to be explicit. If the bridge registration fails, the client can retry without bridge registration if so desired. Otherwise we risk the potential of "eating" the bridge registration failure and the client falsely believing that it can break the web-socket connection.

kitcambridge: Sounds good. Thanks for clarifying!

PR #53, opened 2015-04-29
Work around a Moto bug in `get_connection`

kitcambridge:
Easy to reproduce by running the smoke tests with a mock Moto server. @bbangert already trapped `JSONResponseError` in `get_uaid`; this just adds the same logic to `get_connection`.


jrconlin:
WFM: And gets rid of the weird error comment. We might want to expand the test cases to note the extra exception case, but it's for a testing system anyway, so not critical.

PR #63, opened 2015-05-13
Add a `pong_delay` option for excessive pings

kitcambridge:
`pong_delay` waits to respond to client pings, to work around clients that enter a ping loop.

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -51,6 +51,7 @@ class AutopushSettings(object):
                  pingConf=None,
                  resolve_hostname=False,
                  min_ping_interval=20,
+                 pong_delay=0,
                  max_data=4096,
                  enable_cors=False):
 
@@ -73,6 +74,7 @@ class AutopushSettings(object):
         key = crypto_key or Fernet.generate_key()
         self.fernet = Fernet(key)
 
+        self.pong_delay = pong_delay
         self.min_ping_interval = min_ping_interval
         self.max_data = max_data
         self.clients = {}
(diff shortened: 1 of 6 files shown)


autopush/settings.py
```diff
@@ -51,6 +51,7 @@ def __init__(self,
                  pingConf=None,
                  resolve_hostname=False,
                  min_ping_interval=20,
+                 pong_delay=0,
                  max_data=4096,
```
jrconlin: nit, we should make max_data a config option as well. It's a number and we may want to change it.

bbangert:
Needs config example updated. And if you could add min_pong_interval too, I missed that.

autopush/settings.py
```diff
@@ -51,6 +51,7 @@ def __init__(self,
                  pingConf=None,
                  resolve_hostname=False,
                  min_ping_interval=20,
```
jrconlin: I thought we were going to not make this hard-coded?

kitcambridge: This is the default, no?

jrconlin: Yes. I'm a dope. Nevermind.

autopush/websocket.py
```diff
@@ -392,15 +393,24 @@ def finish_notifications(self, notifs):
             self._check_notifications = False
             reactor.callLater(1, self.process_notifications)
 
+    def _send_ping(self):
+        self.last_ping = time.time()
+        self.metrics.increment("updates.client.ping")
```
bbangert: Needs to include tags=self.base_tags.

bbangert:
r+

PR #64, opened 2015-05-13
Fix `{min_ping_interval, pong_delay}` setting

kitcambridge:
Ensure we pass both options to the `AutopushSettings` constructor.


bbangert:
r+

PR #68, opened 2015-05-14
Track deferreds and remove them on client shutdown.

### LINKED ISSUE #66: Bugfix: Track deferred's, for cancelling
All deferred's generated in a client should be tracked, so they can be cancelled if a client suddenly disconnects.

bbangert:
Fixes #66.

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -45,6 +45,33 @@ class SimplePushServerProtocol(WebSocketServerProtocol):
     # Testing purposes
     parent_class = WebSocketServerProtocol
 
+    # Defer helpers
+    def deferToThread(self, func, *args, **kwargs):
+        d = deferToThread(func, *args, **kwargs)
+        self._callbacks.append(d)
+
+        def f(result):
+            if d in self._callbacks:
+                self._callbacks.remove(d)
+            return result
+        d.addBoth(f)
+        return d
+
+    def deferToLater(self, when, func, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


autopush/websocket.py
```diff
@@ -45,6 +45,33 @@ class SimplePushServerProtocol(WebSocketServerProtocol):
...
+    # Defer helpers
+    def deferToThread(self, func, *args, **kwargs):
+        d = deferToThread(func, *args, **kwargs)
+        self._callbacks.append(d)
+
+        def f(result):
+            if d in self._callbacks:
+                self._callbacks.remove(d)
+            return result
+        d.addBoth(f)
```
kitcambridge: Does `addBoth` re-raise failures, too?

bbangert:
Don't merge yet, waiting to update CHANGELOG.

autopush/websocket.py
bbangert: It gets called whether its a failure passed in, or result, and since I return it, its passed regardless to the next in the chain.

kitcambridge: Ah, [I see](https://twistedmatrix.com/documents/current/core/howto/defer.html#visual-explanation) returning failures is equivalent to throwing. */me notes for the future*

bbangert: Yea, if you look at the deferToLater I made, I call failure.Failure() inside an exception handling block, and Failure() will find it to initialize itself with the current error being handled.

Issue #67, opened 2015-05-14
Feature: Add structured log output for endpoint


bbangert:
The endpoint should log more details on every hit, using the structured log output that is setup, to track:

- User-Agent
- Status code
- Hash of UAID
- Remote-IP

PR #70, opened 2015-05-14
Feature/endpoint logging

bbangert:
Fixes Issue #67.

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -10,7 +10,7 @@ from boto.dynamodb2.exceptions import (
 from cryptography.fernet import InvalidToken
 from StringIO import StringIO
 from twisted.internet.threads import deferToThread
-from twisted.python import log
+from twisted.python import failure, log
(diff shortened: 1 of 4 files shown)


autopush/endpoint.py
```diff
@@ -103,6 +120,7 @@ def _process_uaid(self, result):
         """Process the result of the AWS call"""
         if not result:
             self.set_status(404)
+            log.msg("UAID not found in AWS.", **self._client_info())
```
kitcambridge: Should we log the `uaid` hash too?

bbangert: Yup, patch incoming.

kitcambridge:
Looks great!

kitcambridge:
That's a very strange, unrelated failure. Looks like it tries to connect to AWS instead of using moto?

bbangert:
Yea, quite odd, kicking the build.

bbangert:
Looks fine now.

PR #71, opened 2015-05-15
Disconnect dupe UAIDs

### LINKED ISSUE #65: Bugfix: Disconnect dupe uaid's, sweep all connections for excessive idle
On connect, we should sendClose any prior client connection of the same UAID.

Also, we should have a separate task scheduled, that sweeps through *all* the connections, and if the last ping was more than N minutes/seconds ago, initiates a server-side ping......

OR.....

We turn on autobahn's websocket auto-ping feature.

jrconlin:
Find duplicate UAIDs and disconnect older connections.
Added Websocket Pings to disconnect dead sockets.

fixes #65

@bbangert r?

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -163,6 +163,21 @@ def _parse_connection(sysargs=None):
         default=20,
         type=int,
         env_var="MIN_PING_INTERVAL")
+    parser.add_argument(
+        '--auto_ping_interval',
+        help="Interval between Websocket pings",
+        default=0,
+        type=float,
+        env_var="AUTO_PING_INTERVAL")
+    parser.add_argument(
+        '--auto_ping_timeout',
+        help="Timeout in seconds for Websocket ping replys",
+        default=4,
+        type=float,
+        env_var="AUTO_PING_TIMEOUT")
+    parser.add_argument('--pong_delay', help=("Time to wait after receiving a "
+                        "pong for clients that ping too frequently"),
+                        default=0, type=int, env_var="PONG_DELAY")
 
     add_bridge_args(parser)
     add_shared_args(parser)
@@ -262,6 +277,10 @@ def connection_main(sysargs=None):
         router_scheme="https" if args.router_ssl_key else "http",
         router_hostname=args.router_hostname,
         router_port=args.router_port,
+        min_ping_interval=args.min_ping_interval,
+        auto_ping_interval=args.auto_ping_interval,
+        auto_ping_timeout=args.auto_ping_timeout,
+        pong_delay=args.pong_delay,
     )
     setup_logging("Autopush")
 
@@ -294,7 +313,6 @@ def connection_main(sysargs=None):
         maxFramePayloadSize=2048,
         maxMessagePayloadSize=2048,
         openHandshakeTimeout=5,
-        failByDrop=True,
     )
 
     settings.metrics.start()
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -278,8 +278,6 @@ def connection_main(sysargs=None):
         router_hostname=args.router_hostname,
         router_port=args.router_port,
         min_ping_interval=args.min_ping_interval,
-        auto_ping_interval=args.auto_ping_interval,
-        auto_ping_timeout=args.auto_ping_timeout,
         pong_delay=args.pong_delay,
     )
     setup_logging("Autopush")
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -306,10 +306,6 @@ def connection_main(sysargs=None):
         debug=args.debug,
         debugCodePaths=args.debug,
     )
-    factory.setProtocolOptions(
-        autoPingInterval=args.auto_ping_interval,
-        autoPingTimeout=args.auto_ping_timeout,
-    )
     factory.protocol = SimplePushServerProtocol
     factory.protocol.ap_settings = settings
     factory.setProtocolOptions(
@@ -317,6 +313,8 @@ def connection_main(sysargs=None):
         maxFramePayloadSize=2048,
         maxMessagePayloadSize=2048,
         openHandshakeTimeout=5,
+        autoPingInterval=args.auto_ping_interval,
+        autoPingTimeout=args.auto_ping_timeout,
     )
 
     settings.metrics.start()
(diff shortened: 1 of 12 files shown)


configs/autopush_connection.ini
```diff
@@ -22,3 +22,12 @@ router_port = 8081
 endpoint_scheme = http
 ; endpoint_hostname = updates.push.services.mozilla.com
 endpoint_port = 8082
+
+; Settings for the websocket ping. If a websocket ping fails or times out
+; the connection is auto-closed. This is useful for detecting dead
+; connections because various routers may artificially keep TCP connections
+; alive. This is because the internet is full of lies.
```
kitcambridge: The Internet is full of lies!

autopush/main.py
```diff
@@ -292,6 +306,10 @@ def connection_main(sysargs=None):
         debug=args.debug,
         debugCodePaths=args.debug,
     )
+    factory.setProtocolOptions(
+        autoPingInterval=args.auto_ping_interval,
+        autoPingTimeout=args.auto_ping_timeout,
+    )
```
bbangert: Shouldn't these go in the existing setProtocolOptions call 3 lines down?

autopush/websocket.py
```diff
@@ -327,9 +330,27 @@ def _check_router(self, paused=False, bridge_register=None):
             self.sendMessage(json.dumps(msg).encode('utf8'), False)
             return
         # User exists?
+        # Attempt to register the user for this session
+        # First, check to see if this uaid is registered somewhere else
+        record = self.ap_settings.router.get_uaid(self.uaid)
```
bbangert: Blocking call, this needs to defer to thread.

autopush/main.py
jrconlin: yes. I also need new eyes, apparently.

autopush/websocket.py
```diff
@@ -327,9 +330,27 @@ def _check_router(self, paused=False, bridge_register=None):
...
+        if record:
+            killUrl = record.get("node_id")
+            # Only delete foreign registers. We can handle the others.
+            if killUrl != self.ap_settings.router_url:
+                log.msg("Killing duplicate uaid: %s at %s" %
+                        (self.uaid, killUrl))
+                killUrl = killUrl + "/notif/" + self.uaid
+                d = self.ap_settings.agent.request(
+                    "DELETE",
+                    killUrl.encode("utf8"),
```
bbangert: Delete call should probably include timestamp of the record as well, just in case we start executing here, and the client jumps while we're doing this and lands there. That way we only delete the client if its connect time matches the record.

autopush/websocket.py
```diff
@@ -327,9 +330,27 @@ def _check_router(self, paused=False, bridge_register=None):
...
+                ).addCallback(self._reg_user)
+                d.addErrback(self.log_err, extra="Failed to kill old client")
+                return
```
bbangert: So, I don't think we should have connection nodes attempt to nuke dupe clients, for two reasons:

- Any client that is disconnected, then has a notification come in, will have its node entry deleted. So this only helps a wee bit.
- There's some concurrency issues having them try to call back and forth to delete entries, they'll need to include the connect time to ensure they don't accidentally tell a different node to delete a possibly new and active client that has reconnected in the span of this processing.

If the websocket ping frame is low enough (3 mins), and we delete dupe connections on the same node, that should keep connection churn low enough as is.

autopush/websocket.py
```diff
@@ -74,7 +74,10 @@ def f():
 
     @property
     def base_tags(self):
-        return self._base_tags if self._base_tags else None
+        try:
+            return self._base_tags
+        except AttributeError:
+            return None
```
bbangert: Not sure what case can cause this, also, the effect is not the same as before. The other return statement causes the return value to be None if self._base_tags is [], this is because there was a bug with the datadog library requiring it to be None instead of [].

autopush/websocket.py
```diff
@@ -327,9 +330,27 @@ def _check_router(self, paused=False, bridge_register=None):
...
+                log.msg("Killing duplicate uaid: %s at %s" %
+                        (self.uaid, killUrl))
+                killUrl = killUrl + "/notif/" + self.uaid
+                d = self.ap_settings.agent.request(
+                    "DELETE",
+                    killUrl.encode("utf8"),
+                ).addCallback(self._reg_user)
+                d.addErrback(self.log_err, extra="Failed to kill old client")
+                return
+        self._reg_user()
```
bbangert: Where's the check to see if the client is already connected locally and drop them? This only seems to send a kill if the client is not connected locally, but doesn't force them close if they are a local dupe.

autopush/websocket.py
```diff
@@ -535,7 +556,8 @@ def process_ack(self, data):
         if not updates or not isinstance(updates, list):
             return
 
-        self.metrics.increment("updates.client.ack", tags=self.base_tags)
+        self.metrics.increment("updates.client.ack",
+                               tags=self._base_tags)
```
bbangert: Why isn't this using self.base_tags for the check it needs to ensure it doesn't use [] ?

autopush/websocket.py
```diff
@@ -628,3 +650,9 @@ def put(self, uaid):
         client.process_notifications()
         settings.metrics.increment("updates.notification.checking")
         self.write("Notification check started")
+
+    def delete(self, uaid):
+        client = self.ap_settings.clients.get(uaid)
+        if client:
+            client.onClose(False, 0, "duplication")
```
bbangert: This calls the clients onClose handler, but the onClose func is just something the protocol calls in the course of shutting down the connection, calling it won't actually shutdown the connection. It should call client.sendClose() to start the shutdown of the connection, during the actual close sendClose will eventually call the onClose function.

autopush/settings.py
```diff
@@ -51,6 +51,8 @@ def __init__(self,
                  pingConf=None,
                  resolve_hostname=False,
                  min_ping_interval=20,
+                 auto_ping_interval=0,
+                 auto_ping_timeout=4,
```
bbangert: Why is this passed in over here?

jrconlin: Left over code. removing.

bbangert:
r+

PR #78, opened 2015-05-18
Feature/adaptive pong delay

### LINKED ISSUE #73: Adaptive pong delay
We should pong no more than once every 5 seconds, however our pong delay is a hard-coded value that doesn't consider client latencies. So instead we should track the last time we saw a ping, and if that value is more than 5 seconds, respond immediately, otherwise subtract the value from 10 (to ensure the client doesn't timeout waiting for our pong).

bbangert:
Fixes #73.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,15 @@ Autopush Changelog
 1.1rc3 (**dev**)
 ================
 
+Features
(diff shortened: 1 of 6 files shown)


jrconlin:
r+ (since it's mostly deletion anyway)

PR #80, opened 2015-05-19
Aggressively delete old TCP connections

### LINKED ISSUE #72: Aggressively delete old TCP connections or re-registration
Idle TCP connections continue to slowly grow. We should cull old connections at new device registrations.

jrconlin:
* fixes #72
* Uses dynamodb update return_values to get old node_id, sends
  DELETE request to kill old connection.
* Spent too much time fiddling with trying to Mock the request
  call. Not terribly happy with the "testing" attribute fix.
* Likewise a bit concerned that boto and moto seem to have different
  mock returns for return_value content

@bbangert r?


autopush/websocket.py
```diff
@@ -352,13 +360,31 @@ def finish_hello(self, result):
...
+            attrs = result.get("Attributes")
+            node_id = self._get_aval(attrs.get("node_id", {}), 'S')
+            uaid = self._get_aval(attrs.get("uaid", {}), 'S')
+            connected_at = self._get_aval(attrs.get("connected_at", {}), 'N')
+            url = "%s/notif/%s/%s" % (node_id, uaid, connected_at)
+            if not getattr(self, 'testing', False):
+                d = self.ap_settings.agent.request(
+                    "DELETE",
+                    url.encode("utf8"),
+                ).addCallback(self.finish_hello)
```
bbangert: Why do we wait on hello for the delete call? Can't it just be tossed out with errors captured and ignored?

autopush/websocket.py
```diff
@@ -352,13 +360,31 @@ def finish_hello(self, result):
...
+    def finish_hello(self, *args):
+        self._register = None
         msg = {"messageType": "hello", "uaid": self.uaid, "status": 200}
         self.ap_settings.clients[self.uaid] = self
         self.sendJSON(msg)
         self.metrics.increment("updates.client.hello", tags=self.base_tags)
         self.process_notifications()
 
-    def process_notifications(self):
+    def process_notifications(self, *args):
```
bbangert: Why *args?

autopush/websocket.py
```diff
@@ -615,7 +641,7 @@ def put(self, uaid):
 
 
 class NotificationHandler(cyclone.web.RequestHandler):
-    def put(self, uaid):
+    def put(self, uaid, *args):
```
bbangert: This is needed because of the regexp change?

autopush/websocket.py
jrconlin: I'd prefer to do that, frankly, but I thought that we need to defer this as a blocking call and preserve the callback chain. If I understand (quite possibly incorrectly), calling finish_hello directly would cause the DELETE to go out before we process. It seems cleaner to me to request the delete now.

Is that correct? Am I being too simplistic?

autopush/websocket.py
jrconlin: yes.

autopush/websocket.py
jrconlin: Ah, missed that one. Originally I was calling process_notifications from a defer and was tossing the arg because we didn't need it. Will remove.

autopush/websocket.py
bbangert: My main worry is if that other node is gone, we wait upwards of 10 seconds for this DELETE, and the client times out waiting for us and reconnects. We have to respond to these commands quickly. It can be deferred, we just don't need to wait for it.

autopush/websocket.py
```diff
@@ -368,12 +368,10 @@ def _check_other_nodes(self, result):
...
             url = "%s/notif/%s/%s" % (node_id, uaid, connected_at)
             if not getattr(self, 'testing', False):
-                d = self.ap_settings.agent.request(
+                self.ap_settings.agent.request(
                     "DELETE",
                     url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+                ).addErrback(self.log_err, extra="Failed to delete old node")
```
bbangert: Probably should have two errbacks, one that traps all the connect errors to a dead node, then the self.log_err. Otherwise we'll prolly end up with tons of useless connect errors in sentry again.

bbangert:
Can this patch also delete locally without firing the agent call? i.e, in finish_register, it runs this line:
```
        self.ap_settings.clients[self.uaid] = self
```
So, when it registers, if the node_id is the local node, instead of just running that, look at the other client already connected and if its connect time is newer than the current client, drop the current client, otherwise sendClose the other.

autopush/websocket.py
```diff
@@ -366,14 +369,27 @@ def _check_other_nodes(self, result):
...
-                d = self.ap_settings.agent.request(
-                    "DELETE",
-                    url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+            inSelf = self.ap_settings.clients.get(uaid)
+            if inSelf:
+                if inSelf.connected_at <= connected_at:
+                    inSelf.sendClose()
```
bbangert: If the connected_at is greater, then we should sendClose to this connection and return.

autopush/websocket.py
```diff
@@ -366,14 +369,27 @@ def _check_other_nodes(self, result):
...
+
+                if not getattr(self, 'testing', False):
+                    self.ap_settings.agent.request(
+                        "DELETE",
+                        url.encode("utf8"),
+                    ).addErrback(
+                        _eatConnections
+                    ).addErrback(
+                        self.log_err,
+                        extra="Failed to delete old node")
```
bbangert: For consistency can we stick with the existing style used everywhere else?

autopush/websocket.py
```diff
@@ -470,11 +486,11 @@ def process_ping(self):
 
         """
         now = time.time()
-        last_ping_ago = now-self.last_ping
+        last_ping_ago = now - self.last_ping
```
bbangert: Good catch, I thought flake8 was supposed to whine at me for this.

autopush/websocket.py
```diff
@@ -366,14 +369,27 @@ def _check_other_nodes(self, result):
...
+                    failure.trap(
+                        ConnectError, ConnectionRefusedError, UserError
+                    )
+
+                if not getattr(self, 'testing', False):
+                    self.ap_settings.agent.request(
+                        "DELETE",
+                        url.encode("utf8"),
+                    ).addErrback(
+                        _eatConnections
```
bbangert: Where is the _eatConnections function? It doesn't seem to be in this PR.

autopush/websocket.py
jrconlin: Ah, misunderstood your initial comment.

autopush/websocket.py
jrconlin: line 379?

autopush/websocket.py
jrconlin: I was getting line wrap warnings. will try to trim accordingly.

autopush/websocket.py
jrconlin: Same here. My VIM flake8 caught it, tox did not.

autopush/websocket.py
bbangert: This part is correct, it just needs an else statement in case connected_at for existing is >.

autopush/websocket.py
bbangert: Indeed it is, sorry, didn't see it inline like that. Could you move it out and up closer to the top about the Protocol since it doesn't need anything in closure scope here.

bbangert: And not to nit too much, but since we're back in Python-land, while twisted unfortunately uses camel-case excessively, our own code should generally try to stick with underscore separated lower-case.

autopush/websocket.py
bbangert: The rationale for having each callback/Errback be a separate line is similar to the rule on imports. That is, as separate lines they're easier to see what is changing in a diff. It's why once you import more than a few things from a module, its generally better to go multiline with a (), so that changes to imports are easier to parse in a diff than modifications to a single line.

I.e.
```
from something import foo, smith, george, fred

# instead should be

from something import (
    foo,
    smith,
    george,
    fred,
)
```

autopush/websocket.py
jrconlin: Ok, also going to rework variable label names because I can see that getting really confusing with the current names.

autopush/websocket.py
```diff
@@ -366,14 +369,27 @@ def _check_other_nodes(self, result):
...
             connected_at = self._get_aval(attrs.get("connected_at", {}), 'N')
-            url = "%s/notif/%s/%s" % (node_id, uaid, connected_at)
-            if not getattr(self, 'testing', False):
-                d = self.ap_settings.agent.request(
-                    "DELETE",
-                    url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+            inSelf = self.ap_settings.clients.get(uaid)
```
bbangert: inSelf -> existing_client?

autopush/websocket.py
```diff
@@ -365,15 +368,31 @@ def _check_other_nodes(self, result):
...
-                    url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+            last_connect = self._get_aval(attrs.get("connected_at", {}), 'N')
+            existing = self.ap_settings.clients.get(uaid)
+            if existing:
+                if existing.connected_at <= last_connect:
+                    existing.sendClose()
+                    return
```
bbangert: Shouldn't return here or we won't finish the hello.

autopush/websocket.py
```diff
@@ -365,15 +368,31 @@ def _check_other_nodes(self, result):
...
-            connected_at = self._get_aval(attrs.get("connected_at", {}), 'N')
-            url = "%s/notif/%s/%s" % (node_id, uaid, connected_at)
-            if not getattr(self, 'testing', False):
-                d = self.ap_settings.agent.request(
-                    "DELETE",
-                    url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+            last_connect = self._get_aval(attrs.get("connected_at", {}), 'N')
```
bbangert: So, thing to remember here, last_connect != this_connect, because last_connect is the prior connected_at time, not this one. This makes it somewhat hard to reason about the logic below for me.

autopush/websocket.py
```diff
@@ -365,15 +368,31 @@ def _check_other_nodes(self, result):
...
+            existing = self.ap_settings.clients.get(uaid)
+            if existing:
+                if existing.connected_at <= last_connect:
+                    existing.sendClose()
+                    return
+                if existing.connected_at > last_connect:
```
bbangert: So, the connected_at *could* be greater than the last_connect but *not greater than this connect*, no?

autopush/websocket.py
```diff
@@ -365,15 +368,31 @@ def _check_other_nodes(self, result):
...
-                d = self.ap_settings.agent.request(
-                    "DELETE",
-                    url.encode("utf8"),
-                ).addCallback(self.finish_hello)
-                d.addErrback(self.log_err, extra="Failed to delete old node")
-                return
+            last_connect = self._get_aval(attrs.get("connected_at", {}), 'N')
+            existing = self.ap_settings.clients.get(uaid)
+            if existing:
+                if existing.connected_at <= last_connect:
```
bbangert: Instead of last_connect, this comparison should prolly be done with self.connected_at.

autopush/websocket.py
jrconlin: The `last_connect` is the last, recorded connection we have. 
`existing` is the known existing connection. 
There's also the current connection, which isn't mentioned here, but was just written to the database.

If the existing connection is the same age or older than the last recorded connection, we close it in favor of the current connection; the current connection "wins"
If the existing connection is newer than the last recorded connection, we close the existing connection and return, preventing the current connection from completing as well; the last recorded connection "wins".

Currently, there is no case where the "existing" connection wins, however there's additional logic inside of router to potentially reject connections.

I agree that this is becoming increasingly complicated and that there is a division of logic.

bbangert: Ok, but last_connect was the last connection *before* this update just overwrote it. As of now, the 'last connect' in dynamodb is *this* connection (perhaps). ``existing`` is a current connection to this node for this uaid that *may* be the one overwritten, or *may* be newer than the last_connect but *not* newer than the current connection.

autopush/websocket.py
```diff
@@ -365,15 +368,30 @@ def _check_other_nodes(self, result):
...
+                    existing.sendClose()
+                if existing.connected_at > last_connect:
+                    self.sendClose()
+                    return
+            else:
```
bbangert: No 'else' here, since its possible for an existing to be there *and* the node_id is a different node. So the check here should be to do the delete if the prior node_id is not our own.

bbangert:
r+, waiting on travis

PR #83, opened 2015-05-20
Feature/delete

jrconlin:
* fix bug with hellos with blank UAIDs being rejected. 
* router.register_user returns tuple to deal with python truthiness issues
* Added test to check for definite true when UAID is blank
* fix node_id equivalency check

@bbangert


jrconlin:
We're already doing a test where UAID is "" and it's registering. (see https://github.com/mozilla-services/autopush/blob/master/autopush/tests/test_websocket.py#L304) Let me know if you're looking for a more in-depth check. I'm also not checking the second half of the tuple when we're forcing the router to return {} to work around the moto problem, but I'm happy to add that logic.

bbangert:
r+

bbangert:
Needs a changelog entry for this, and 1.2.1 version bump.

bbangert:
r+ once changelog/version changes are in, looks good

PR #84, opened 2015-05-21
Fix for Sentry Alert trap error

jrconlin:
because i failure.Failure() as fail.


kitcambridge:
:+1: This is such a small change, but do you think it makes sense to add a test?

bbangert:
The code had no test before, which is why it wasn't detected, so I think it makes sense that this should include a test.

PR #87, opened 2015-06-02
WIP: Add detailed logging and metrics for various shutdown paths

bbangert:


autopush/websocket.py
```diff
@@ -366,19 +419,19 @@ def _check_other_nodes(self, result):
             self.sendMessage(json.dumps(msg).encode('utf8'), False)
             return
 
+        existing = self.ap_settings.clients.get(self.uaid)
```
jrconlin: NIT: we may want to use a different term here (preexisting?) to describe what we're doing here. "existing" is a bit vague, since self is essentially also existing. We're looking for older instances on the common node that may share the UAID.

Not critical for this PR.

jrconlin:
r+

PR #90, opened 2015-06-10
Feature/prop ping modularize

### LINKED ISSUE #82: Modularize prop ping further
Right now each prop ping's code is its own module, but the prop ping isn't very modular inside endpoint.py.

For each prop ping, we should probably have a mapping:
```
proprietary code | should_store | func
gcm                false          gcm.some_func
tef                true           tef.other_func
```
Indicating whether we should store the message and/or attempt local delivery, or pass it.

bbangert:
Fixes #82.


CHANGELOG.rst
```diff
@@ -2,16 +2,24 @@
...
+==================
 
 Features
 --------
 
 * Additional logging/metrics on auto-ping and connection aborting.
 
-1.2.2 (May 27, 2015)
-====================
+1.2.2 (2015-05-27)
```
jrconlin: +1 on date format.

autopush/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '1.2.3'
+__version__ = '1.3.0dev'
```
jrconlin: Why are we breaking the version format this way? I thought that we were simply going to increment. (I really don't want to get into the whole OAuth 1.0a v. 2.0 v.s 1.1.4 crap.)

autopush/main.py
```diff
@@ -80,20 +78,16 @@ def add_shared_args(parser):
...
-        '--max_data',
-        help="Max data segment length in bytes",
-        default=4096,
-        env_var='MAX_DATA')
+    parser.add_argument('--max_data', help="Max data segment length in bytes",
+                        default=4096, env_var='MAX_DATA')
 
 
-def add_bridge_args(parser):
+def add_router_args(parser):
```
jrconlin: We're mixing terms here. We should remove "bridge" if this is simply another form of routing, otherwise we'll have too many synonyms.

jrconlin:
autopush/pinger/__init__.py should be removed. (it's cruft at this point)

autopush/__init__.py
bbangert: This was mainly just to reflect that its not tagged yet. We could just use 1.3.0 I suppose even though its not tagged for it yet.

bbangert:
I thought I did remove pinger/init?

autopush/__init__.py
jrconlin: I think I'd prefer that. Less chance of forgetting later, we're already on the next version, and we have a single track for code right now anyway.

jrconlin:
Ah, you did. I misread the diff.

CHANGELOG.rst
bbangert: Per richard's request that we match some other projects.

autopush/endpoint.py
```diff
@@ -1,71 +1,126 @@
...
-from autopush.protocol import IgnoreBody
+from autopush.router.interface import RouterException
+from autopush.utils import (
+    generate_hash,
+    validate_hash,
+)
 
 
-dead_cache = LRUCache(150)
+class Notification(namedtuple("Notification", "version data channel_id")):
```
kitcambridge: :heart: `namedtuple`.

jrconlin:
r+ pending travis.

autopush/websocket.py
```diff
@@ -352,63 +346,40 @@ def process_hello(self, data):
...
-        if connect and self.ap_settings.bridge:
-            self.transport.pauseProducing()
-            d = self.deferToThread(self.ap_settings.bridge.register, uaid,
-                                   connect)
-            d.addCallback(self._check_router, True)
-            d.addErrback(self.err_hello)
-        else:
-            self._check_router(False)
+        # Default router choice
+        router_type = data.get("router_type", "simplepush")
```
jrconlin: Why are we looking for this in the main data? I'd like to keep the main object as simple as possible and push potentially conflicting values into a sub-object.

autopush/websocket.py
```diff
@@ -352,63 +346,40 @@ def process_hello(self, data):
...
-        url = self.ap_settings.router_url
         self.transport.pauseProducing()
-        d = self.deferToThread(router.register_user,
-                               self.uaid, url, self.connected_at)
+        user_item = dict(
+            uaid=self.uaid,
+            node_id=self.ap_settings.router_url,
+            connected_at=self.connected_at,
+            router_type=router_type,
+            router_data={},
```
jrconlin: We're not passing the router_data, meaning that users cannot register anything other than "simplepush" with a websocket interface.

autopush/websocket.py
bbangert: Because there's only two options, simplepush, and soon, webpush. External routers must use the RESTful HTTP API.

autopush/websocket.py
bbangert: Correct, RESTful HTTP API is the *only* way to register external routers or anything other than simplepush, and webpush.

kitcambridge: Ah, cool; I didn't realize that, either. Do you think there's value in supporting UDP wake-up for the WebSocket interface? On the one hand, we already have [clients in the wild](https://github.com/mozilla/gecko-dev/blob/master/dom/push/PushServiceWebSocket.jsm#L1096-1105) that use it, and, unlike our other bridges, it doesn't bypass the socket...but, on the other, a unified interface is *really* nice.

bbangert: I didn't realize we have clients in the wild that we send UDP wake-ups to, I thought that code was for a partner that implements their own system independent of us. My preference would be for a separate RESTful call for a client to add additional router information (such as a UDP wake-up), mainly because the websocket is an implementation detail of this transport, which will eventually be http2. Having a separate scheme devoted to managing custom router/wake data that is outside other specs feels helpful in that regard, since it means we won't bog down the WebPush spec to add custom wake stuff as well (since we'll have a http endpoint independent of the spec for a client to register that).

To that end, it does present the point that for consistency, we should probably keep the websocket simplepush spec for messages like it is now, and if a client wants webpush style independent message delivery, it can make the HTTP call first to get a UAID and ask for it, then do the websocket connection. And no ``router_type`` option will be available during the websocket connection.

jrconlin: On 6/14/2015 3:56 PM, Ben Bangert wrote:
>
> In autopush/websocket.py
> <https://github.com/mozilla-services/autopush/pull/90#discussion_r32385474>:
>
> >          self.transport.pauseProducing()
> > -        d = self.deferToThread(router.register_user,
> > -                               self.uaid, url, self.connected_at)
> > +        user_item = dict(
> > +            uaid=self.uaid,
> > +            node_id=self.ap_settings.router_url,
> > +            connected_at=self.connected_at,
> > +            router_type=router_type,
> > +            router_data={},
>
> Correct, RESTful HTTP API is the /only/ way to register external
> routers or anything other than simplepush, and webpush.
>
>
If that's the case, then Websocket should only register simplepush,
since webpush will exclusively use HTTP2.

I'd also counter that third party routers should be generic. We don't
know what means will be easiest or available for developers, the same
way that we have no idea what routing mechanisms they may want or need.

bbangert: Third party routers are generic, they have a JSON dict of data they can submit in the HTTP request payload for them. Third party routers are not part of the websocket extension, which is non-existent at the moment so it can be whatever we want (and we had all previously said and agreed it was fine for them to be entirely in the RESTful HTTP API). I'm all for dropping it entirely and having it be only SimplePush as you mention.

PR #96, opened 2015-06-15
* Increase test coverage of utils for 100% test coverage.

bbangert:
I previously didn't see the two lines missing since a blanket except statement caught the import error. Python 2.7.7 and above have compare_digest, the rest don't. I've also bumped the pypy dependency to ensure it requires a Python appropriate, and removed the try/except that wasn't needed.

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -46,11 +46,8 @@ def validate_uaid(uaid):
 def validate_hash(key, payload, hashed):
     """Validates that a UAID matches the HMAC value using the supplied
     secret"""
-    try:
-        h = hmac.new(key=key, msg=payload, digestmod=hashlib.sha256)
-        return hmac.compare_digest(hashed, h.hexdigest())
-    except:
-        return False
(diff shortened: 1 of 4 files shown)


autopush/utils.py
```diff
@@ -46,11 +46,8 @@ def validate_uaid(uaid):
 def validate_hash(key, payload, hashed):
     """Validates that a UAID matches the HMAC value using the supplied
     secret"""
-    try:
-        h = hmac.new(key=key, msg=payload, digestmod=hashlib.sha256)
-        return hmac.compare_digest(hashed, h.hexdigest())
-    except:
-        return False
+    h = hmac.new(key=key, msg=payload, digestmod=hashlib.sha256)
```
jrconlin: Is the plan to just let potential exceptions bubble up? (+1 on not eating them, just wondering about logging)

bbangert: It does, if they're not both strings, in which case I'll be very curious how that happened.

jrconlin:
r+

PR #97, opened 2015-06-16
Fix regex to make register have optional end slash

jrconlin:
Fix regex to allow /register and /register/


autopush/main.py
```diff
@@ -323,7 +323,7 @@ def endpoint_main(sysargs=None):
         (r"/push/([^\/]+)", EndpointHandler, dict(ap_settings=settings)),
         # PUT /register/ => connect info
         # GET /register/uaid => chid + endpoint
-        (r"/register/([^\/]+)?", RegistrationHandler,
+        (r"/register/?([^\/]+)?", RegistrationHandler,
```
kitcambridge: Could we also split this up into two handlers—one for `/register/([^\/]+)`, the other for `/register`? Or not really worth it?

bbangert:
r+

PR #99, opened 2015-06-17
s/router/router_type in endpoint

jrconlin:
correct bug where endpoint was not getting the correct router from the stored routing data.

@bbangert r?


bbangert:
r+, looks good, nice test-case (I really should've included something like that, doh). Waiting for Travis.

bbangert:
r+ again :)

Issue #98, opened 2015-06-17
Add sphinx docs for autopush


bbangert:
Sphinx docs for a full doc site for autopush should be added. The restful docs for the registration endpoint should be switched to the sphinx http restful plugin for nice HTTP docs.

bbangert:
Closed via #100.

PR #100, opened 2015-06-17
Feature/sphinx docs

bbangert:
* Add Sphinx docs with ReadTheDocs publishing. Issue #98.
  This change also includes a slight Metrics refactoring with a IMetrics
  interface, and renames MetricSink -> SinkMetrics for naming consistency.


jrconlin:
Seems a bit inconsistent which parameters we name and which ones we don't. Right now, it looks like we're naming "the important" ones, based on function usage and likelihood of someone interacting with them, and that's ok. Do you think we'll need to define all parameters and returns in the future?

autopush/db.py
```diff
@@ -36,34 +39,38 @@ def create_storage_table(tablename="storage", read_throughput=5,
...
-    return Table(tablename)
-
-
-def storage_table(tablename="storage"):
-    return Table(tablename)
-
-
 def get_router_table(tablename="router", read_throughput=5,
                      write_throughput=5):
+    """Get the main router table object
```
jrconlin: Later functions end definitions with punctuation. We should pick a style and stay with it.

autopush/endpoint.py
```diff
@@ -1,8 +1,178 @@
...
+
+If no external routers are configured then the :class:`RegistrationHandler`
+will not be able to perform any additional router data registration.
+
+HTTP API
+========
+
+API methods requiring Authorization must include a HMAC, generated with SHA256
+of the secret key sent on the original register ``POST`` and the message
+payload being the raw content of the body (if any).
```
jrconlin: We should probably note that the payload should be UTF-8 encoded.

bbangert:
I don't think we'll be defining every one of them so explicitly. The database functions seemed most critical in recognizing usage and exceptions raised, so I doc'd them completely. For callbacks, I mainly just focused on trying to make it more obvious how they fit together, reading the code will be required regardless.

autopush/endpoint.py
bbangert: Added.

jrconlin:
r+

autopush/db.py
bbangert: Updating for consistent styling. First line in a multi-line doc string is considered a 'header' and has no period, all other lines must have a period at the end of a sentence.

Issue #104, opened 2015-06-24
Add max connections


bbangert:
Expose the autobahn max connection as a config option so we can cut off excess connections before the server overloads.

crodjer:
If no one has started working on it yet, may I take this up?

bbangert:
Sure, sounds good.

bbangert:
Closed by #105.

PR #105, opened 2015-06-28
Support for limiting maximum concurrent connections

crodjer:
Add max connections as a config argument to avoid server load over a
safe value. In case of a DOS attack, this would avoid a failure of
existing connections.

This PR doesn't include tests, since I couldn't figure a way to test it. I 
understand that a test for this would become a test for Autobahn, 
which would be the library's responsibility?

When I set the limit to 1, this results in a `503` with the message:
`maximum number of connections reached` on a second connection.

When the old connection is closed, a connection spot is open again. The
default  `0`, removes any soft limits on the number of connections.

Ref #104


bbangert:
r+, looks good!

crodjer:
Great!

PR #111, opened 2015-07-10
Feature/udp

### LINKED ISSUE #106: Add a UDP wake-up bridge
Client bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1157696

Now that @bbangert's fantastic routing refactor has landed, we can add a separate bridge for the TEF [UDP wake-up platform](https://github.com/telefonicaid/wakeup_platform). Unlike our other bridges, this one only sends a signal for the client to reconnect if it's offline.

It works like this:

* When the client connects to the push server, it opens a UDP port and includes the IP, port, and carrier info in the [handshake](https://github.com/mozilla/gecko-dev/blob/master/dom/push/PushServiceWebSocket.jsm#L1108-1121): `{"messageType":"hello","uaid":"","channelIDs":[],"wakeup_hostport":{"ip":"127.0.0.1","port":65535},"mobilenetwork":{"mcc":"001","mnc":"01","netid":"001-01.default"}}`. We'll store these params for the wake-up request.
* If the server supports UDP wake-up, and the connection idles for [10 seconds](https://github.com/telefonicaid/notification_server/blob/v1.3/src/ns_ua/connectors/ConnectorUDP.js#L48-50), the server closes the WebSocket with a special close code (4774). The client [won't reconnect](https://github.com/mozilla/gecko-dev/blob/master/dom/push/PushServiceWebSocket.jsm#L1230-1235) when it receives this close code.
* When the push server wants to wake up a client, it sends an HTTP request to the TEF wake-up server (authenticated with a TLS client cert issued by TEF).
* The TEF wake-up server sends a UDP packet to the client, which [prompts the client to reconnect](https://github.com/mozilla/gecko-dev/blob/master/dom/push/PushServiceWebSocket.jsm#L1287-1288) to the push server.

If we see a client isn't online, we can just store the message, and send the ping afterward. The client will get that message the next time it reconnects.

jrconlin:
Fixes #106.

* Add UDP Wake support for SimplePush using router
** Adds timeout to boot connections that specify UDP wake.
* Fix some flake8 issues
* convert tests that use ...**.assert_called()** to use eq_(...**.called**, True) to fix travis incompatibilities
* Fix missing labels for some args.
* convert some worrisome " is " tests to " == "
* bumped pyasn to 0.1.8 to fix install issue.

@bbangert @kitcambridge r?


autopush/db.py
```diff
@@ -247,7 +247,7 @@ def register_user(self, data):
                 for key, value in result["Attributes"].items():
                     try:
                         r[key] = self.table._dynamizer.decode(value)
-                    except AttributeError:
+                    except (TypeError, AttributeError):
```
kitcambridge: What does the `TypeError` catch?

autopush/websocket.py
```diff
@@ -304,6 +309,8 @@ def onMessage(self, payload, isBinary):
             return self.process_ack(data)
         else:
             self.sendClose()
+        # Done processing, start idle.
+        self.set_active()
```
kitcambridge: What do you think about moving `self.check_idle()` below this line, so that you don't need to call it in `process_notifications`, `process_ping`, etc.?

autopush/router/simple.py
```diff
@@ -137,9 +145,27 @@ def route_notification(self, notification, uaid_data):
             returnValue(RouterResponse(response_body="Delivered"))
         else:
             self.metrics.increment("router.broadcast.miss")
+            if self.udp is not None:
+                yield deferToThread(self._send_udp_wake,
+                                    self.udp)
             returnValue(RouterResponse(202, "Notification Stored"))
 
-    #############################################################
+    def _send_udp_wake(self, udp_info):
```
kitcambridge: It might make sense to move this logic into a `SimpleRouter` class, and expose it as a separate router type. On the other hand, that'd actually be more complex than just the one conditional you have here (and YAGNI). What do you think?

autopush/router/simple.py
```diff
@@ -137,9 +145,27 @@ def route_notification(self, notification, uaid_data):
...
+        host = udp_info.get("wakeup_host").get("ip")
+        port = udp_info.get("wakeup_host").get("port")
+        data = json.dumps(udp_info.get("mobilenetwork", {}))
+        if port is not None:
+            host = "%s:%d" % (host, port)
+        response = requests.post(
```
kitcambridge: This isn't quite right, but close!

* The `POST` request is made to the wake-up server (`https://wakeup.tefdigital.com/wakeup/v1`, IIRC), not the device itself. There's a staging server, too; we can make the URL a config option.
* The body contains `mcc`, `mnc`, `netid`, `port`, and `ip`. We should probably just normalize the data before storing it in Dynamo.
* The body needs to be URL-encoded, not JSON. I think you can just pass `data` as-is to `requests.post`; it'll encode it automatically.

autopush/db.py
jrconlin: If value is not proper javascript, it throws a TypeError. If table or _dynamizer aren't initialized, they throw Attribute error. In either case, just take the value as specified and don't be clever.

autopush/router/simple.py
jrconlin: Heh, the first approach I took was to do that, however I realized that i had to break isolation when I did the timeout stuff. Plus the routers tend to think more in "This is a total hand-off including data" where the UDP stuff is more a side-band wake up routine.

autopush/settings.py
```diff
@@ -124,9 +126,15 @@ def __init__(self,
         # CORS
         self.cors = enable_cors
 
+        # Force timeout in idle seconds
+        self.idle_timeout = idle_timeout
```
kitcambridge: Maybe `udp_timeout` is a better name? I'm worried this would cause confusion down the road, particularly if we add a handshake timeout, too.

autopush/router/simple.py
jrconlin: Ah, figured the host was for load balancing on the provider side. Having a single point is fine, but might make doing multiple provider UDP wake support a bit more of a challenge

autopush/logging.py
```diff
@@ -76,7 +76,7 @@ def stdout(message):
     else:
         sys.stdout.write("       %s\n" % message['message'])
     return
-    """
+    # """
```
bbangert: This seems unintentional?

autopush/main.py
```diff
@@ -101,13 +101,25 @@ def add_external_router_args(parser):
...
+        '--udp_timeout',
+        help="UDP: idle timeout before closing socket",
+        type=int, default=0, env_var="UDP_TIMEOUT")
+    parser.add_argument(
+        '--udp_pem',
+        help="UDP: custom TLS PEM file for remote Wake server",
+        type=str, env_var="UDP_PEM")
+    parser.add_argument(
+        '--udp_server', help="UDP: endpoint for UDP wake-up calls",
+        type=str, default='http://example.com', env_var="UDP_SERVER")
```
bbangert: Nit: Inconsistent dropping of arguments. All the existing add_argument calls keep as much on the top line as possible before dropping/indenting.

autopush/main.py
```diff
@@ -189,6 +201,13 @@ def _parse_endpoint(sysargs):
 def make_settings(args, **kwargs):
     """Helper function to make a :class:`AutopushSettings` object"""
     router_conf = {}
+    # Some routers require a websocket to timeout on idle (e.g. UDP)
+    timeout = 0
+    if args.udp_pem is not None and args.udp_timeout is not 0:
```
bbangert: Nit: 'is not 0' should be '!= 0'.

autopush/main.py
```diff
@@ -189,6 +201,13 @@ def _parse_endpoint(sysargs):
 def make_settings(args, **kwargs):
     """Helper function to make a :class:`AutopushSettings` object"""
     router_conf = {}
+    # Some routers require a websocket to timeout on idle (e.g. UDP)
+    timeout = 0
```
bbangert: Why not just set it to args.udp_timeout, since it defaults to 0 anyways?

autopush/logging.py
jrconlin: Actually, sorta intentional. It lets me activate the "human readable logging" by adding a # to the start of line 73 instead of to both lines. The triple quote stops on that line regardless of the starting hash.

autopush/router/simple.py
bbangert: I was actually thinking that there'd be a IWakeup interface, and UDP wake would be one of them. They would be called after routing has completed, before the response returns, I even added a little comment to endpoint.py highlighting where the IRouter check would go to trigger it: https://github.com/mozilla-services/autopush/blob/master/autopush/endpoint.py#L404

The main reason for keeping it 100% separate from the router, is that the current router doesn't do individual messages (webpush style), while I will shortly have a new, separate router, that does. So multiple routers might want a wake-up, and hooking it onto the end of routing (at the spot in the endpoint.py) allows them to remain separate.

autopush/router/simple.py
```diff
@@ -137,9 +145,23 @@ def route_notification(self, notification, uaid_data):
...
+    def _send_udp_wake(self, udp_info):
+        host = self.conf["udp_server"]
+        response = requests.post(
+            host,
+            data=udp_info["data"],
+            cert=self.conf.get("cert"))
+        if response.status_code < 200 or response.status_code >= 300:
+            raise RouterException("Could not send UDP Wakeup",
+                                  status_code=500,
+                                  response_body="Could not send UDP Wakeup")
```
bbangert: If we successfully stored the message for the user, I'm not sure why we'd return a server error of 500 just because maybe a partner's system to wake is not responding at the moment.

autopush/main.py
jrconlin: Counter nit: Flake8 (at least my instance of it) complains about lines not being properly indented unless the line ends with a "(". Additional indenting would put the help message over 79 columns, which would mean that I'd have to break the help line in two. "udp_server" does try to keep the argument line as compact as possible. 

I'm fine breaking the help messages, but I don't think that they'll be quite as readable.

autopush/main.py
jrconlin: Ah, good catch. Missed this during some cleanup.

autopush/websocket.py
```diff
@@ -412,13 +421,35 @@ def process_hello(self, data):
...
+            udp_data = dict(ip=wakeup_host.get("ip"),
+                            port=wakeup_host.get("port"),
+                            mcc=mobilenetwork.get("mcc", ''),
+                            mnc=mobilenetwork.get("mnc", ''),
+                            netid=mobilenetwork.get("netid", ''))
+            self.udp = dict(data=urlencode(udp_data),
+                            timeout=self.ap_settings.udp_timeout)
+        except (AttributeError, KeyError):
+            # Some UDP value is missing, toss it all.
+            pass
```
bbangert: If we maintain a separate list of IWakeup implementations, each could have a 'check_hello' method that is called, to determine what impl is used. This keeps the logic out of websocket.py (where it can easily grow quite a bit if/when we add more wakeup schemes).

autopush/websocket.py
```diff
@@ -550,6 +583,25 @@ def _send_ping(self):
...
+            if self.udp is not None:
+                if ms_time() - self.idle >= self.udp_timeout * 1000:
+                    self.sendClose(code=4774, reason='UDP Wakeup')
+                    return
+                self.idler = self.deferToLater(self.udp_timeout,
+                                               self.check_idle)
+        except (KeyError, AttributeError):
+            # More than likely, this isn't a UDP wake connection.
+            # log.msg("Warn: could not check idle %s" % x)
+            pass
```
bbangert: Should probably be a method on the implementation of IWakeup that indicates whether an 'idle' connection should be closed, and how.

bbangert:
I'd really like to see the wakeup split out of the router/websocket into its own small object with a few methods that implements a Wake interface (and hooked onto the completion of routing). I won't be surprised if in the future we may end up doing something similar to what Google does, where we actually have all active mobile devices connect directly to us, but when 'gone idle', they drop, and we wake them with GCM/APNS/etc. In which case we can easily grow several wakeup methods, and having each one be a custom one-off in this style will impair us.

autopush/main.py
jrconlin: Huh, I stand corrected, the line is exactly 79 characters long.

jrconlin:
Fair. I considered this a "one off" mostly because that's all we at at the moment, and the alternate bridge mechanisms handled their own idle periods. Having a wake interface makes sense. Do we expect there to be other side event interfaces (Has data/ Priority/ etc.)?

bbangert:
Well, the reason I could see copying Google's approach, is that we might find GCM/APNS lacking in some way for optimal message delivery/latency, so FF on mobile when its in-use would actually hold open a connection directly to us, but drop it as soon as the app is backgrounded/killed.

autopush/router/simple.py
jrconlin: I think it's useful to relay the wakeup failure. I'm not sure what response code makes sense. I suppose 202 with body content indicating the warning makes the most sense.

jrconlin:
Note: this may require refactoring if PR #115

autopush/main.py
```diff
@@ -113,19 +113,30 @@ def add_external_router_args(parser):
...
     parser.add_argument('--apns_cert_file',
                         help="%s Certificate PEM file" % label,
                         type=str, env_var="APNS_CERT_FILE")
-    parser.add_argument('--apns_key_file', help="%s Key PEM file",
+    parser.add_argument('--apns_key_file', help="%s Key PEM file" % label,
                         type=str, env_var="APNS_KEY_FILE")
+    # UDP
+    parser.add_argument('--wake_timeout',
+                        help="UDP: idle timeout before closing socket",
+                        type=int, default=0, env_var="wake_TIMEOUT")
```
bbangert: Should be all uppercase.

autopush/router/simple.py
```diff
@@ -48,6 +49,13 @@ def __init__(self, ap_settings, router_conf):
         """Create a new SimpleRouter"""
         self.ap_settings = ap_settings
         self.metrics = ap_settings.metrics
+        self.conf = router_conf
+        self.waker = None
+        if self.conf is not None and self.conf.get("server") is not None:
+            # This should use a Wake resolver dict, but there's only one
+            # right now, so skip a few instructions.
+            self.waker = UDPWake(host=self.conf.get("server"),
+                                 cert=self.conf.get("cert"))
```
bbangert: We create a waker on every single simplepush (and only simplepush?) notification even though only UDP wake clients need it?

autopush/router/simple.py
```diff
@@ -70,6 +78,12 @@ def route_notification(self, notification, uaid_data):
         # Determine if they're connected at the moment
         node_id = uaid_data.get("node_id")
         uaid = uaid_data["uaid"]
+        self.udp = None
+        try:
+            self.udp = json.loads(uaid_data["udp"])
```
bbangert: The common path for 99% of clients is going to be no UDP, exceptions are cheap in Python so we're essentially slow-pathing the common case by adding an exception to it. A 'if udp in uaid_data' test would be fast.

autopush/router/simple.py
```diff
@@ -152,9 +166,21 @@ def route_notification(self, notification, uaid_data):
             returnValue(self.delivered_response(notification))
         else:
             self.metrics.increment("router.broadcast.miss")
+            if self.udp is not None:
+                yield deferToThread(self._send_wake,
```
bbangert: Do we care about waiting for the reply?

autopush/router/simple.py
bbangert: Also, uaid_data is a JSON dict out of DynamoDB. DynamoDB items can store structured nested dicts, so in theory we don't need to JSON encode it on the way in/out like this.

autopush/router/simple.py
```diff
@@ -152,9 +166,21 @@ def route_notification(self, notification, uaid_data):
...
 
-    #############################################################
+    def _send_wake(self, wake_info):
+        if self.waker is not None:
+            try:
+                self.waker.send_wake(wake_info)
+            except WakeException:
+                raise RouterException("Could not send wakeup",
+                                      status_code=500,
+                                      response_body="Could not send Wakeup")
```
bbangert: Why does the appserver need to know this? Client wasn't connected, we couldn't deliver UDP wake, but like any other client it will eventually get its notification so we didn't error. Plus, we already stored the notification at this point, so if the appserver gets a 500, it might retry.

autopush/waker.py
```diff
@@ -0,0 +1,101 @@
...
+class UDPWake(IWake):
+    """ Proprietary UDP wake tool """
+
+    data = {}
+    default_period = 10 * 1000  # default to 10 seconds
+    endpoint = None     # Third party wake endpoint
+    timeout = 0
+    idle = 0
+
+    def _ms_time(self):
```
bbangert: We should prolly stick this in one spot and use it from there. My bad, I'll open a ticket for that.

autopush/websocket.py
```diff
@@ -512,7 +523,26 @@ def process_hello(self, data):
...
+                             port=wakeup_host.get("port"),
+                             mcc=mobilenetwork.get("mcc", ''),
+                             mnc=mobilenetwork.get("mnc", ''),
+                             netid=mobilenetwork.get("netid", '')))
+            # set this so we can store it for the endpoint to call.
+            self.ps.waker = UDPWake(protocol=self,
+                                    timeout=self.ap_settings.wake_timeout,
+                                    kill_func=self.sendClose, code=4774,
+                                    reason="UDP Idle")
+        except (AttributeError, KeyError):
```
bbangert: Same problem here with the common path. Since UDP will usually not be present, this will add CPU overhead to every single connection establishment. Better to do if/else checks before starting that branch.

autopush/waker.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def send_wake(self, wake_info=None):
+        """Send a wake request to the external endpoint."""
+        if wake_info is None:
+            wake_info = self.info
+        response = requests.post(self.endpoint,
+                                 data=urlencode(wake_info["data"]),
+                                 cert=self.cert)
+        if response.status_code < 200 or response.status_code >= 300:
+            raise WakeException()
```
bbangert: The entwined nature of this UDPWake object makes me feel uneasy, but I don't have any good alternative that comes to mind. Would be curious to hear @kitcambridge's thoughts on it, given that this object is created and used by the endpoint, and the websocket instance, and in the case of the websocket this has a ap_settings method given as the kill_func... so there's objects in multiple spots with references back/forth.

autopush/websocket.py
```diff
@@ -658,6 +694,8 @@ def finish_notifications(self, notifs):
             msg = {"messageType": "notification", "updates": updates}
             self.sendJSON(msg)
 
+        if self.ps.waker is not None:
+            self.ps.waker.check_idle()
```
bbangert: Why? Won't the try/finally just reset it at the end of this?

autopush/websocket.py
```diff
@@ -351,16 +352,26 @@ def onMessage(self, payload, isBinary):
...
+            elif cmd == "ack":
+                return self.process_ack(data)
+            else:
+                self.sendClose()
+        finally:
+            # Done processing, start idle.
+            if self.ps.waker is not None:
+                self.ps.waker.set_active()
+                # and start the timer on the connection
+                self.ps.waker.check_active()
```
bbangert: I don't see a check_active method on the UDPWake object.

autopush/websocket.py
jrconlin: Yeah, going to keep the try/except to catch malformed descriptors, but adding the if to prevent needless execution.

autopush/websocket.py
```diff
@@ -520,6 +550,12 @@ def process_hello(self, data):
             connected_at=self.ps.connected_at,
             router_type=router_type,
         )
+        # NOTE: in tests, storing a value of None results in a corrupted
+        # return value of True. (This appears to have to do with a bug in
+        # either moto or boto in how it deals with values encoded to
+        # {"NULL": True}. For now, I'm avoiding storing values of None.
+        if wake_data is not None:
+            user_item['wake_data'] = wake_data
```
bbangert: Why not put this in the branch above that populates wake_data? (And just move the wake_data check after the user_item is made)

autopush/websocket.py
bbangert: port has no fallback so it could be a None as shown, would that cause any problems? Or should port have a default?

jrconlin: Kind of a tricky question. ip and port are both UDP provider defined, and should be included. If they're not, things fail (as they would if set to "None"). I'll add a check for these and fail out, but either way things should not progress.

autopush/waker.py
kitcambridge: It seems this object does two things: maintains an idle timer on the connection, and sends a ping if the client is offline. Maybe it makes sense to break it up into two separate objects?

An idle timer could have other uses, too; at one time, I think we talked about disconnecting clients that don't say hello. Twisted's [`TimeoutMixin`](https://twistedmatrix.com/documents/current/api/twisted.protocols.policies.TimeoutMixin.html) might help here; it manages a timer object, and calls a ` timeoutConnection` method when the timer fires.

Second, instead of the endpoint, we could have `AutopushSettings` create and hold on to the `UDPWake` object (maybe in a `ap_settings.wakers` dict, if we wanted to get fancy). The `IWake` interface could then look like this:

* `sleep(self, proto)`: Called by `timeoutConnection`. This is basically your `kill_func`; it'd just call `proto.sendClose(4774, "UDP Idle")`.
* `wake(self, wake_info)`: Called by the endpoint when it wants to wake the client, after routing is complete.

That's just an idea, though. I'll ponder it some more.

jrconlin: Actually, TimeoutMixin looks like a pretty good option. It does most of what I'm trying to do.

Thanks!

CHANGELOG.rst
```diff
@@ -138,6 +138,9 @@ Features
 * Add Sphinx docs with ReadTheDocs publishing. Issue #98.
   This change also includes a slight Metrics refactoring with a IMetrics
   interface, and renames MetricSink -> SinkMetrics for naming consistency.
+* Add UDP Wake support. Issue #106
+  Some devices which use SimplePush routing offer a feature to wake on
+  a carrier provided UDP ping. See issue #106 for details.
```
bbangert: Needs to be moved to a 1.7.0 section now. Sorry bout that.

autopush/router/simple.py
```diff
@@ -76,6 +75,9 @@ def route_notification(self, notification, uaid_data):
         # Determine if they're connected at the moment
         node_id = uaid_data.get("node_id")
         uaid = uaid_data["uaid"]
+        self.udp = None
+        if "udp" in uaid_data:
+            self.udp = uaid_data["udp"]
```
bbangert:
```python
self.udp = uaid_data.get("udp")?

File: autopush/router/simple.py [hunk 26]
```diff
@@ -159,9 +160,19 @@ def route_notification(self, notification, uaid_data):
...
         else:
             self.metrics.increment("router.broadcast.miss")
-            returnValue(self.stored_response(notification))
-
-    #############################################################
+            retVal = self.stored_response(notification)
+            if self.udp is not None and "server" in self.conf:
+                # Attempt to send off the UDP wake request.
+                try:
+                    requests.post(self.conf["server"],
```
bbangert: Does this block?

autopush/websocket.py
```diff
@@ -351,16 +350,26 @@ def onMessage(self, payload, isBinary):
...
+            elif cmd == "ack":
+                return self.process_ack(data)
+            else:
+                self.sendClose()
+        finally:
+            # Done processing, start idle.
+            self.resetTimeout()
+
+    def timeoutConnection(self):
+        """ UDP conneciton has timed out. """
```
bbangert: No spaces on preceding/following triple-quotes.

autopush/websocket.py
```diff
@@ -512,14 +521,27 @@ def process_hello(self, data):
...
+            if "ip" in wakeup_host and "port" in wakeup_host:
+                mobilenetwork = data.get("mobilenetwork")
+                # Normalize the wake info to a single object.
+                wake_data = dict(data=dict(ip=wakeup_host["ip"],
+                                 port=wakeup_host["port"],
+                                 mcc=mobilenetwork.get("mcc", ''),
+                                 mnc=mobilenetwork.get("mnc", ''),
+                                 netid=mobilenetwork.get("netid", '')))
+                user_item["wake_data"] = wake_data
+                self.setTimeout(self.ap_settings.wake_timeout)
```
bbangert: Rad, TimeoutMixin is slick.

CHANGELOG.rst
jrconlin: Heh, tempted to move it to ∞.0.2 at this point.

autopush/router/simple.py
jrconlin: If "udp" is in uaid_data, would accessing "udp" ever trigger a key error? (honest question, I'd love to be sure about something like this.)

bbangert: Nope, if "udp" is in the dict, you'll never get a key error. Technically the way you have it here is marginally faster than using dict.get as well, because get() invokes more of the function call overhead while the "udp" in uaid_data is a pure key check at a lower level without Python function overhead setup. So either way is fine by me. :)

autopush/router/simple.py
jrconlin: Yes. Although I may have misunderstood the intent of a previous R comment. 
I'll add in a yield.

autopush/router/simple.py
jrconlin: nevermind, missed the obvious.

bbangert:
r+

autopush/router/simple.py
```diff
@@ -159,9 +158,20 @@ def route_notification(self, notification, uaid_data):
...
         else:
             self.metrics.increment("router.broadcast.miss")
-            returnValue(self.stored_response(notification))
-
-    #############################################################
+            retVal = self.stored_response(notification)
+            if self.udp is not None and "server" in self.conf:
+                # Attempt to send off the UDP wake request.
+                try:
+                    yield requests.post(
```
bbangert: Er, I dunno why this works in the tests. yield can only be done on results that are deferred, requests.post is a blocking call that doesn't return a yield. It needs a deferToThread in front.

bbangert:
r+

Issue #50, opened 2015-04-17
Services should self identify


jrconlin:
While working on a different project with a different service it occurred to me that we have a lot of moving parts. There are some services that exist in a "dev", "stage", "prod" or other such environment which are isolated from the others. This may be fine, or it may cause unusual issues due to information isolation (e.g. state existing in "stage" will not exist in "dev" leading to changes potentially not being properly reflected in all calls)

It would be useful for data responses to return the environment they are in as an optional component. This way, systems that may be consuming or processing results can easily confirm the source of the data that they are using and not accidentally "cross the streams".

thoughts?

kitcambridge:
Would this be programmatic (i.e., client sends an `Accept-Environment: stage` header for HTTP and WebSocket requests; server responds with a 400 if it's running in a different environment...though, to an extent, the URI already serves that purpose)...or checked manually via `/health`?

jrconlin:
Huh, sending as a request header is a clever idea, but I was thinking more that it's an element in the response. So since many services we have return JSON in some form, there would be an extra element like 

     {"foo":"bar", "env":"stage"}

The reason to include it in the body rather than as a return header would be because the body is going to be more visible. Headers may not be available in some cases, however the data would be. 

In my original problem, a secondary call was polling a production system. This was anomalous, but I was getting the full data response back for some calls. Having the environment self identify would have saved a good bit of effort and helped quickly determine the problem.

crodjer:
If this isn't urgent, could I try working on it? I'll post here or on the IRC if I have any queries about this.

jrconlin:
@crodjer yep, feel free.

Just to note: this was more the product of some wasted cycles trying to figure out what the hell was going on in a test configuration. Actually, it was a combination of test configurations (between Push, FxA and a few other systems) so it's a bit "edge case". There's no hard or fast requirement for it, but it'd be useful.

PR #112, opened 2015-07-13
[WIP] Provide the service enviornment information

crodjer:
In order for the clients to identify which service they are dealing with, provide the running environment in json responses. Being aware of the environment will help in debugging issues with applications.

This patch may need more tests. I tested it out with simplepush_test locally and added a basic test case.

Ref: #50

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -62,6 +62,7 @@ class AutopushSettings(object):
                  max_data=4096,
                  # Reflected up from UDP Router
                  wake_timeout=0,
+                 env='development',
                  enable_cors=False):
         """Initialize the Settings object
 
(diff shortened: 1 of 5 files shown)


autopush/websocket.py
```diff
@@ -394,6 +394,8 @@ def returnError(self, messageType, reason, statusCode, close=True):
 
     def sendJSON(self, body):
         """Send a Python dict as a JSON string in a websocket message"""
+
+        body['env'] = self.ap_settings.env
```
bbangert: I think it should only go in the hello message. Not every single message.

bbangert:
If you could rebase to the latest, and update, it looks good. Still needs a changelog entry for it.

crodjer:
@bbangert Thanks for the review, I'll incorporate your changes in the PR.

crodjer:
Updated the pull request.

autopush/settings.py
```diff
@@ -151,6 +152,10 @@ def __init__(self,
         if 'gcm' in router_conf:
             self.routers["gcm"] = GCMRouter(self, router_conf["gcm"])
 
+
```
jrconlin: Bummer. Need to remove this extra linespace and then all tests will pass. (The test is barking because of Flake8.)

crodjer:
@jrconlin Thanks for pointing that out. Fixed.

jrconlin:
Woot! r+

crodjer:
Thanks

PR #114, opened 2015-07-17
Freeze mock to version `1.0.1`

crodjer:
In their recent upgrade, mock `assert_called` has started to fail, which
is being used in [various places](https://gist.github.com/crodjer/7646fb56f49efd363793) by our tests.

For now, this ensures that the test suite passes. Gradually, we can
upgrade to the latest `mock` by updating the `assert_called` calls. We
should upgrade because `assert_called` is something undocumented anyway:
<http://www.voidspace.org.uk/python/mock/mock.html>.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,6 @@
 nose
 coverage
-mock>=1.0.1
+mock==1.0.1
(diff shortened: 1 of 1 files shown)


crodjer:
Closing, a better patch is available: #115

PR #115, opened 2015-07-23
Convert ...assert_called to assert(...called)

jrconlin:
assert_called is not fully supported.

@bbangert || @kitcambridge r?

cc: @crodjer


autopush/tests/test_websocket.py
```diff
@@ -458,19 +458,6 @@ def check_result(result):
         self._wait_for_close(d)
         return d
 
-    def test_hello_router(self):
```
jrconlin: no longer needed (moved bridge creation to register endpoint)

bbangert:
r+

PR #116, opened 2015-07-24
Move all dependencies into requirements.txt and freeze

bbangert:
Freezes all deps with versions in a single location.


jrconlin:
Single source for requirements?
**R+** 
default

PR #120, opened 2015-07-28
Add webpush style message delivery

### LINKED ISSUE #57: Add WebPush style individual message delivery w/data
Modify the architecture to store individual messages with payload

# Implementation Plan

To retain efficient message delivery and storage, a separate router and db table is needed for storing individual messages. Our current router table has some space to store arbitrary additional data, but in this case we need to know in advance whether the channel exists or not, to avoid storing lots of arbitrary data that may be for expired/unregistered channels.

This first version will have a new table, keyed by hash/range key: uaid-channel_id / timestamp

Using a channel_id of "" combined with uaid will return a structure with a 'channels' field that has all registered channels.

After the router lookup indicates to use individual message delivery, the individual message router will verify the channel id is valid, and deliver/store it as needed.

## Websocket Changes

The router type will be stored for the connection, so several methods can act appropriately based on whether the prior simplepush style delivery is used, or webpush style. The following methods will need to be modified to toggle implementation based on router_type:

- register
- unregister
- process_notifications
- ack

## Database Changes & Additions

A new table for individual message storing, and all the registered channels will be added for this user.

## Router

New 'webpush router' for webpush style data retention and routing will be added.s, instead of collapsing by version.

bbangert:
Add WebPush style data delivery with crypto headers to connected clients.

Implementation Notes:
- Added Message table, with hash key of UAID, range key of CHANNEL_ID:MESSAGE_ID.
  This allows querying for messages of a specific channel with beginswith operator, along with
  easy retrieval of all notifications for a single UAID. The blank CHID:MID is a record containing
  all registered channel ID's to avoid delivery for a CHID we don't support.
- Valid channel's are verified by the endpoint before attempting delivery as a preflight check in the router.
- Each message gets a message ID assigned and returned as Location header to AppServer per WebPush spec.
- Websocket client ack's messages with channel_id/version as normal.
- Notifications are delivered as independent websocket messages per Issue #57.
- Added Webpush router as a minimal subclass of the simple router, that saves the message appropriately and overrides the RouterResponse to include the Location header and use a 201.
- Minimal changes to endpoint.py to add WebPush special-case. This was chosen instead of trying to make things more pluggable, since we will only be supporting legacy SimplePush, or WebPush style.
- Minimal changes to websocket.py (unfortunately still not trivial), to add hooks for clients using webpush style individual message notification.

Other Notes:
- No changes were made to most tests, nor were any made to the smoke-test (to ensure all existing clients will continue to work without change).
- Moto required brain surgery to add set operations for channel lookup operations. As such, the version in test-requirements is required for tests to run.

Closes #57.


autopush/endpoint.py
```diff
@@ -360,36 +366,56 @@ def put(self, token):
...
+            req_fields = ["content-encoding", "encryption"]
+            if not all([x in self.request.headers for x in req_fields]):
+                self.set_status(401)
+                log.msg("Missing Crypto headers", **self._client_info())
+                self.write("Missing crypto headers.")
+                return self.finish()
+
+            version = uuid.uuid4().hex
+            data = self.request.body
+            router.message_id = version
```
kitcambridge: Maybe I'm missing something, but it looks like there's one instance of each router per server (`self.ap_settings.routers[router_key]`). To make this work, wouldn't you need to instantiate a new router per request?

autopush/db.py
```diff
@@ -175,6 +207,113 @@ def delete_notification(self, uaid, chid, version=None):
             return False
 
 
+class Message(object):
+    """Create a Message table abstraction on top of a DynamoDB Table object"""
+    def __init__(self, table, metrics):
+        """Create a new Router object
```
kitcambridge: s/Router/Message.

autopush/db.py
```diff
@@ -175,6 +207,113 @@ def delete_notification(self, uaid, chid, version=None):
...
+        results = self.table.query_2(
+            uaid__eq=uaid,
+            chidmessageid__beginswith="%s:" % channel_id,
+            consistent=True,
+            attributes=("chidmessageid",),
+        )
+        chidmessageids = []
+        for item in results:
+            chidmessageids.append(item["chidmessageid"])
+            if len(chidmessageids) == 25:
```
kitcambridge: Oh, I had no idea DynamoDB limits batches to 25 operations. Could you add a quick comment here, please?

autopush/endpoint.py
bbangert: Ah crap, you're right on that. Will fix.

autopush/endpoint.py
```diff
@@ -360,36 +366,56 @@ def put(self, token):
...
         # Save the whole record
         router_key = result.get("router_type", "simplepush")
         router = self.ap_settings.routers[router_key]
+
+        # Only simplepush uses version/data out of body/query, GCM/APNS will
+        # use data out of the request body 'WebPush' style.
+        if router_key == "simplepush":
+            version, data = parse_request_params(self.request)
+        else:
+            # We need crypto headers
```
kitcambridge: This is only for Web Push, right? Or are we planning to send encrypted messages via GCM/APNS, too?

autopush/db.py
bbangert: Oops, leftover from a prior method of fetching. I've removed this entirely.

autopush/endpoint.py
bbangert: I was trying to avoid bloating the router interface any farther for this. Since GCM/APNS both can handle data, and neither is in actual use yet, my thought was to allow them to use data right away rather than have yet another set of possibility combinations.

autopush/tests/test_router.py
```diff
@@ -457,3 +458,73 @@ def verify_deliver(result):
...
+        self.router_mock = settings.router = Mock(spec=Router)
+        self.message_mock = settings.message = Mock(spec=Message)
+        self.agent_mock = Mock(spec=settings.agent)
+        settings.agent = self.agent_mock
+        self.router.metrics = Mock()
+
+    def test_route_to_busy_node_saves_looks_up_and_sends_check_201(self):
+        self.agent_mock.request.return_value = response_mock = Mock()
+        response_mock.addCallback.return_value = response_mock
+        type(response_mock).code = PropertyMock(
```
kitcambridge: Nifty.

autopush/endpoint.py
jrconlin: @bbangert +1.

autopush/websocket.py
```diff
@@ -472,7 +523,11 @@ def _check_other_nodes(self, result):
     def finish_hello(self, *args):
         """callback for successful hello message, that sends hello reply"""
         self._register = None
-        msg = {"messageType": "hello", "uaid": self.uaid, "status": 200}
+        if self.use_webpush:
+            msg = {"messageType": "hello", "uaid": self.uaid, "status": 200,
+                   "use_webpush": True}
+        else:
+            msg = {"messageType": "hello", "uaid": self.uaid, "status": 200}
```
jrconlin: going to ping on the duplicate dictionary here (same as in the smoke tests

autopush/endpoint.py
kitcambridge: Sounds good. Thanks for clarifying!

bbangert: Correct, we are sending encrypted messages via the bridges.

kitcambridge:
LGTM!

jrconlin:
r+

PR #123, opened 2015-07-29
Add WebPush TTL's

### LINKED ISSUE #56: Add a TTL header for incoming messages
The Web Push spec defines a `TTL` header for app servers to specify message retention duration. Setting `TTL: 0` indicates a message is ephemeral, and can be dropped immediately if the client is disconnected. This feature was requested by the Loop team to match the current Loop Push behavior, so let's add it to Simple Push.

DynamoDB doesn't support TTLs, but we can work around this by storing the expiration time and dropping expired messages when the client reconnects. We can also specify a `TTL` header in the response with the actual time-to-live...so, if an app server sends too many updates, we can indicate we're dropping messages via `TTL: 0`.

The Web Push spec also says that an omitted `TTL` is equivalent to `TTL: 0`. This could be a back-compat issue for us, since the current behavior is to store messages if the client is disconnected. OTOH, I like that storing messages requires an explicit opt-in, so it'd be nice to make this change while we have few users.

bbangert:
Add WebPush TTL scheme per spec (as of July 28th 2015).

This is a relatively minor change to store the TTL value, and avoid storage of it if the TTL is set to 0. The spec is slightly ambiguous on whether a Location should still be returned if a TTL is 0 and it couldn't be delivered (such that no such resource exists, and its not clear if a NACK exists). So for now, no Location is returned if the message couldn't be delivered, the spec says POST always returns a 201 though, so, its still a 201.

To break off router execution if the TTL is 0, I threw a RouterException in the webpush router, and added a headers field for the possibility that in the future, a Location should still be returned in this case.

Closes #56.

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -206,7 +206,7 @@ from autopush.utils import (
 
 
 class Notification(namedtuple("Notification",
-                   "version data channel_id headers")):
+                   "version data channel_id headers ttl")):
     """Parsed notification from the request"""
 
 
@@ -335,8 +335,11 @@ class AutoendpointHandler(cyclone.web.RequestHandler):
     def _router_fail_err(self, fail):
         """errBack for router failures"""
         fail.trap(RouterException)
-        log.err(fail, **self._client_info())
         exc = fail.value
+        if not str(exc.status_code).startswith('2'):
+            # Only log non-2XX responses, we sometimes have to throw one to
+            # stop processing even though its a 2XX response
+            log.err(fail, **self._client_info())
         self.set_status(exc.status_code)
         self.write(exc.response_body)
         self.finish()
@@ -405,6 +408,7 @@ class EndpointHandler(AutoendpointHandler):
             version = uuid.uuid4().hex
             data = self.request.body
 
+        ttl = self.request.headers.get("TTL", 0)
         if data and len(data) > self.ap_settings.max_data:
             self.set_status(401)
             log.msg("Data too large", **self._client_info())
@@ -413,7 +417,8 @@ class EndpointHandler(AutoendpointHandler):
 
         notification = Notification(version=version, data=data,
                                     channel_id=self.chid,
-                                    headers=self.request.headers)
+                                    headers=self.request.headers,
+                                    ttl=ttl)
 
         d = Deferred()
         d.addCallback(router.route_notification, result)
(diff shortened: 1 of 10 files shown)


autopush/endpoint.py
```diff
@@ -335,8 +335,11 @@ def _overload_err(self, fail):
     def _router_fail_err(self, fail):
         """errBack for router failures"""
         fail.trap(RouterException)
-        log.err(fail, **self._client_info())
         exc = fail.value
+        if not str(exc.status_code).startswith('2'):
+            # Only log non-2XX responses, we sometimes have to throw one to
```
jrconlin: Heh, not that it matters for this, but it's 3x slower than numeric range checking. (Yeah, I was curious so i ran a test)

autopush/router/webpush.py
```diff
@@ -62,6 +63,7 @@ def _send_notification(self, uaid, node_id, notification):
                               "version": notification.version,
                               "data": notification.data,
                               "headers": self._crypto_headers(notification),
+                              "ttl": notification.ttl+int(time.time()),
```
jrconlin: This might cause a few headaches later if the server clocks ever start to skew by more than a few seconds. Granted, we'll have other problems then. Might want to note that we could pass the ttl as a vector and let the router adjust based off it's local clock in case this problem does creep up.

autopush/tests/test_router.py
```diff
@@ -508,6 +508,29 @@ def verify_deliver(result):
...
+            side_effect=MockAssist([202, 200]))
+        self.message_mock.store_message.return_value = True
+        self.message_mock.all_channels.return_value = [dummy_chid]
+        router_data = dict(node_id="http://somewhere", uaid=dummy_uaid)
+        self.router_mock.get_uaid.return_value = router_data
+        self.router.message_id = uuid.uuid4().hex
+
+        d = self.router.route_notification(notif, router_data)
+
+        def verify_deliver(fail):
```
jrconlin: silly question: Why is the param called fail if we're looking for a status of 201?

autopush/router/webpush.py
bbangert: The spec allows for various differences in clock skew when the TTL hits the server, which was one reason for this. The other reason was that I needed the raw TTL value in _store_notification, so I couldn't do the calculation in endpoint.py.

autopush/tests/test_router.py
bbangert: Because its a twisted failure instance with addErrback.

Issue #125, opened 2015-07-30
RouterException needs a 'should_log' flag


bbangert:
All RouterExceptions that aren't 2XX series are logged at the moment. Node was Invalid returns a 503 because we want an app-server to retry (now that we cleared the node entry from the db), and will happen a lot in production intentionally, so this shouldn't be logged as an error.

To remedy this, and remove the odd 2XX check around the RouterException handling in endpoint.py, an attribute should be added to RouterException indicating whether it should be logged as something we should attend to. The 2XX RouterExceptions from router/webpush.py and the NodeInvalid one should be flagged to not log.

PR #126, opened 2015-08-04
Remove excessive exception logging

bbangert:
Use log_exception on RouterException to determine if it should be logged (and flagged in Sentry).
Don't log common exception use, where the exception is not actually an actionable bug.

Closes Issue #125.


kitcambridge:
LGTM!

PR #129, opened 2015-08-11
Integrate smoke-tests, also fixes bugs found as a result

### LINKED ISSUE #119: Integrate smoke-test into test harness
The simplepush smoke test needs to be integrated as a 'normal' test inside autopush. The code should be moved over, and some of the code from my ssl-ram-testing that controls processes can be run to setup all the processes and run the smoke test.

bbangert:
Bug Fixes
---------

* Fix deferToLater to not call the function if it was cancelled using a
  canceller function.
* Fix finish_webpush_notifications to not immediately call
  process_notifications as that will be called as needed after ack's have been
  completed.
* Fix process_ack to not call process_notifications when using webpush if there
  are still remaining notifications to ack.

Features
--------

* Integrate simplepush_test smoke-test client with the main autopush test-suite
  into the test-runner. Issue #119.

Closes #119.

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -160,6 +160,7 @@ keyid="http://example.org/bob/keys/123;salt="XZwpw6o37R-6qoZjw6KwAw"\
         log.debug("%s body: %s", method, body)
         http.request(method, url.path, body, headers)
         resp = http.getresponse()
+        http.close()
         log.debug("%s Response: %s", method, resp.read())
         eq_(resp.status, status)
         if self.use_webpush and ttl != 0:
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -1,16 +1,31 @@
+import httplib
+import json
+import logging
 import os
+import random
 import signal
 import subprocess
+import time
+import urlparse
+import uuid
 from unittest.case import SkipTest
 
 import boto
 import psutil
+import websocket
+from nose.tools import eq_, ok_
 from twisted.trial import unittest
 from twisted.internet import reactor
+from twisted.internet.defer import inlineCallbacks, returnValue
+from twisted.internet.threads import deferToThread
 
+log = logging.getLogger(__name__)
 here_dir = os.path.abspath(os.path.dirname(__file__))
 root_dir = os.path.dirname(os.path.dirname(here_dir))
 moto_process = None
+
+import twisted.internet.base
+twisted.internet.base.DelayedCall.debug = True
 
 
 def setUp():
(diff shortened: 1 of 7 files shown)


autopush/tests/test_integration.py
```diff
@@ -0,0 +1,626 @@
...
+        log.debug("Send: %s", msg)
+        self.ws.send(msg)
+
+    def disconnect(self):
+        self.ws.send_close()
+        self.ws.close()
+        self.ws = None
+
+
+class IntegrationBase(unittest.TestCase):
```
jrconlin: Kinda get why you need to do this, but it does seem like we're potentially covering future bugs that might be in the handlers. I have no idea how we'd prevent that at the moment.

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,626 @@
...
+        client = yield self.quick_register(use_webpush=True)
+        yield client.disconnect()
+        ok_(client.channels)
+        yield client.send_notification(data=data, status=202)
+        yield client.send_notification(data=data2, status=202)
+        yield client.connect()
+        yield client.hello()
+        result = yield client.get_notification()
+        ok_(result != {})
+        ok_(result["data"] in [data, data2])
```
jrconlin: is the return data order important?

jrconlin:
One general Q, but r+ pending travis.

autopush/tests/test_integration.py
bbangert: Nope, thus why I just check to see if its some of the data I'm expecting.

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,629 @@
...
+                body = "version=%s" % (version or "")
+            if use_header:
+                headers = {"Content-Type": "application/x-www-form-urlencoded"}
+            else:
+                headers = {}
+            method = "PUT"
+
+        log.debug("%s body: %s", method, body)
+        http.request(method, url.path, body, headers)
+        resp = http.getresponse()
```
kitcambridge: That's an interesting API. Does that mean you can only have one outstanding HTTP request at a time? (Not really relevant to the PR, just a curiosity).

bbangert: Not entirely sure, I haven't tried sending multiple at once without checking the response.

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,629 @@
...
+        settings.factory = factory
+        self.websocket = reactor.listenTCP(9010, factory)
+
+        # Websocket HTTP router
+        # Internal HTTP notification router
+        r = RouterHandler
+        r.ap_settings = settings
+        n = NotificationHandler
+        n.ap_settings = settings
+        ws_site = cyclone.web.Application([
```
kitcambridge: It'd be nice if these were unified with the definitions in `main.py`, so we'd only need to change them in one place. OTOH, apart from adding message `DELETE` and receipts, we won't be changing them frequently.

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,629 @@
...
+        )
+        self.website = reactor.listenTCP(9020, site)
+        self._settings = settings
+
+    def tearDown(self):
+        self.websocket.stopListening()
+        self.website.stopListening()
+        self.ws_website.stopListening()
+
+    @inlineCallbacks
```
kitcambridge: Ah, so nice. :smile_cat:

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,629 @@
...
+log = logging.getLogger(__name__)
+here_dir = os.path.abspath(os.path.dirname(__file__))
+root_dir = os.path.dirname(os.path.dirname(here_dir))
+moto_process = None
+
+import twisted.internet.base
+twisted.internet.base.DelayedCall.debug = True
+
+
+def setUp():
```
kitcambridge: Are these called automatically by the test runner?

autopush/tests/test_integration.py
```diff
@@ -0,0 +1,629 @@
...
+                                            version=version)]))
+        log.debug("Send: %s", msg)
+        self.ws.send(msg)
+
+    def disconnect(self):
+        self.ws.close()
+
+
+class IntegrationBase(unittest.TestCase):
+    def setUp(self):
```
kitcambridge: Silly question—are these `setUp` and `tearDown` methods called for each `test_` method, or just once?

autopush/tests/test_integration.py
bbangert: Yup.

autopush/tests/test_integration.py
bbangert: I considered that, but the setup here is rather unique and doesn't fit well with what the main script wants to do without making even more changes and a large refactor to main.py.

autopush/tests/test_integration.py
bbangert: Each one.

autopush/tests/test_integration.py
bbangert: If a bug was in the handler, wouldn't that still error out, but just in more of the tests?

autopush/websocket.py
```diff
@@ -540,12 +542,14 @@ def process_notifications(self):
 
         # Are we paused? Try again later.
         if self.paused:
-            self.deferToLater(1, self.process_notifications)
+            d = self.deferToLater(1, self.process_notifications)
+            d.addErrback(self.trap_cancel)
```
kitcambridge: Just to clarify: the deferred will still be rejected with a `CancelledError`, even if you pass a custom `canceller`?

autopush/websocket.py
```diff
@@ -839,6 +841,11 @@ def check_missed_notifications(self, results, resume=False):
             # Resume consuming ack's
             self.transport.resumeProducing()
 
+        # When using webpush, we don't check again unless we have no
+        # outstanding notifications
```
kitcambridge: Nit: "if we have outstanding notifications".

kitcambridge:
Looks good! I have questions, but none related to the tests themselves. :shipit:

autopush/tests/test_integration.py
bbangert: Yea, the tests came out pretty nice with that.

autopush/websocket.py
bbangert: https://twistedmatrix.com/documents/15.3.0/api/twisted.internet.defer.Deferred.html#__init__ explains it a bit. The custom canceller is used in this case just to record that it was cancelled so the func doesn't run. I removed the cancel trapping from the self.deferToLater/self.deferToThread, and I don't want any errors from this, so I added the trap to capture them and ignore them.

kitcambridge: > If a canceller is not given, or does not invoke its argument's callback or errback method, `Deferred.cancel` will invoke `Deferred.errback` with a `CancelledError`.

That's the part I was missing. Thanks!

PR #130, opened 2015-08-12
Remove adaptive ping, log connection losses

### LINKED ISSUE #103: Send a special WebSocket close code for clients that ping too frequently
For context: https://bugzilla.mozilla.org/show_bug.cgi?id=1152264

In #78, we introduced an adaptive response delay for clients that ping too frequently. Unfortunately, folks are still reporting high battery and data usage, even with the fix in place. This is caused by a bug in the client's adaptive ping logic, and affects all FxOS 2.x releases (1.x is unaffected because it didn't ship with adaptive pings).

Any client can potentially enter this state (especially those on unreliable networks), and there's no recovery apart from manually resetting the prefs. The client patch is in place, but has not yet been uplifted.

On our end, we can detect when clients enter a ping loop, and send a special WebSocket close code (4774). This is normally used for UDP wake-up: if the client detects this code, it won't reconnect, as it expects the server to wake it up for incoming notifications.

The trade-off is that phones on non-TEF networks won't receive any push notifications until their network status changes—either they lose reception and reconnect, or their phone switches between cellular and Wi-Fi. (TEF has their own UDP wake-up platform, so we can actually make this work for them). But it's a small price to pay for battery life and reasonable data usage.

A vague plan:

- [ ] Remove the adaptive response delay, and disconnect clients that ping too frequently.
- [ ] On disconnect, store the connection lifespan in DynamoDB. I think this calls for a weighted moving average, to minimize the impact on well-behaved clients that happen to be on spotty networks.
- [ ] When the client reconnects, look up its average connection lifespan. If it drops below a threshold (15 seconds?), flush any pending messages, then close its connection via `self.sendClose(code=4774)`.

bbangert:
* Remove adaptive ping entirely. Send special close code and drop clients that
  ping more frequently than 55 seconds (approx 1 min). This will result in
  clients that ping too much being turned away for awhile, but will alleviate
  data/battery issues in buggy mobile clients. Issue #103.
* Log all disconnects, whether they were clean, the code, and the reason.

Closes #103.

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -549,6 +549,14 @@ class WebsocketTestCase(unittest.TestCase):
         f.addErrback(lambda x: d.errback(x))
         return d
 
+    def test_auto_ping(self):
+        self.proto.ping_time_out = False
+        self.proto.dropConnection = Mock()
+        self.proto.debugCodePaths = False
+        self.proto.onAutoPingTimeout()
+        ok_(self.proto.ping_time_out, True)
+        ok_(self.proto.dropConnection.called)
+
     def test_deferToLater(self):
         self._connect()
 
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -54,6 +54,7 @@ class WebsocketTestCase(unittest.TestCase):
         self.proto.sendClose = self.close_mock = Mock()
         self.proto.transport = self.transport_mock = Mock()
         self.proto.closeHandshakeTimeout = 0
+        self.proto.autoPingInterval = 300
         self.proto._force_retry = self.proto.force_retry
         settings.metrics = Mock(spec=Metrics)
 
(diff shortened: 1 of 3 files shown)


kitcambridge:
Looks good! Going to try this out with my Flame tomorrow.

autopush/websocket.py
```diff
@@ -657,28 +672,24 @@ def _send_ping(self):
...
 
         """
         now = time.time()
         last_ping_ago = now - self.last_ping
-        if last_ping_ago >= 9:
+        if last_ping_ago >= 55:
             self._send_ping()
         else:
-            return self.deferToLater(9 - last_ping_ago, self._send_ping)
+            self.sendClose(code=4774)
```
kitcambridge: Actually, could you add a counter here, please? I'm wondering how many times we'll hit this...

autopush/tests/test_websocket.py
```diff
@@ -54,6 +54,7 @@ def setUp(self):
         self.proto.sendClose = self.close_mock = Mock()
         self.proto.transport = self.transport_mock = Mock()
         self.proto.closeHandshakeTimeout = 0
+        self.proto.autoPingInterval = 300
```
jrconlin: Do we want to make this a config value now? I'm  being a bit paranoid because the value is a guess and having it a config would let us make a tweak without pushing code.

jrconlin:
Looks good.

autopush/tests/test_websocket.py
bbangert: This value is set by the config already, since we don't fully instantiate the class here with all the settings, this just sets it up as it will be when used outside of the unit test.

jrconlin: sorry, my confusion. I'm ok with merging once the counter Kit asked for is in.

kitcambridge: Ben and I talked about this a bit more yesterday, but I'd love to hear your thoughts, too. A counter would satisfy our curiosity, but it's not really actionable. (This is the same guidance I've received for Sync telemetry in Firefox).

In our case, we already know clients are susceptible to this bug; we just can't do anything except institute the workaround. It doesn't tell us if the workaround is effective, just that clients are buggy. Reduced reconnect rates might be a better measure.

Ben pointed out this applies to other metrics we record, too—for example, what new information do we get from counting the number of client pings?

jrconlin: Well, I'm a fan of collecting metrics that answer questions rather than just bulk collection, but part of this exercise is that we really don't have much information backing up some of the decisions we're making. In this case, knowing how many aggressive pings we're shutting down shows how many clients we're effectively orphaning. These will be devices that will not have ping work correctly, and that's a concern. Hopefully it's a small percentage.

I'll agree that auditing other metrics is definitely worth it.

jrconlin: Removing my concern after discussion. We can derive the information from analyzing the logging metadata. The metric doesn't carry enough data to be useful.

jrconlin:
r+

PR #131, opened 2015-08-14
Message deletion endpoint

kitcambridge:
Here's a sketch of a `DELETE` implementation for messages. Still needs tests, but I thought I'd extract what I have so far.

This only deletes messages from storage, not direct updates. I thought about having it handle those, too, but that'd mean an extra routing table read to look up the home node. That seems fine, since deleting a message sent to a connected client is inherently racy, and it's far more likely the message will be delivered before the connection node receives the "oops, don't send that" signal.

f? @bbangert @jrconlin

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -342,6 +342,13 @@ class AutoendpointHandler(cyclone.web.RequestHandler):
         self.write("Server busy, try later")
         self.finish()
 
+    def _router_response(self, response):
+        self.set_status(response.status_code)
+        for name, val in response.headers.items():
+            self.set_header(name, val)
+        self.write(response.response_body)
+        self.finish()
+
     def _router_fail_err(self, fail):
         """errBack for router failures"""
         fail.trap(RouterException)
@@ -351,9 +358,7 @@ class AutoendpointHandler(cyclone.web.RequestHandler):
         if 200 <= exc.status_code < 300:
             log.msg("Success", status_code=exc.status_code,
                     **self._client_info())
-        self.set_status(exc.status_code)
-        self.write(exc.response_body)
-        self.finish()
+        self._router_response(exc)
 
     def _uaid_not_found_err(self, fail):
         """errBack for uaid lookup not finding the user"""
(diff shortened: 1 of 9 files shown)


autopush/main.py
```diff
@@ -347,6 +351,7 @@ def endpoint_main(sysargs=None):
     # Endpoint HTTP router
     site = cyclone.web.Application([
         (r"/push/([^\/]+)", EndpointHandler, dict(ap_settings=settings)),
+        (r"/m/([^\/]+)", MessageHandler, dict(ap_settings=settings)),
```
bbangert: How is this url communicated to the AppServer?

autopush/router/interface.py
```diff
@@ -33,6 +33,13 @@ def __init__(self, status_code=200, response_body="", router_data=None,
         self.router_data = router_data
         self.headers = headers
 
+    def respond(self, handler):
+        handler.set_status(self.status_code)
+        handler.write(self.response_body)
+        for name, val in self.headers.items():
+            handler.set_header(name, val)
+        handler.finish()
+
```
bbangert: Why isn't it sufficient to have the handler take the response and fill that in like it does now?

autopush/endpoint.py
```diff
@@ -449,32 +507,7 @@ def _router_completed(self, response, uaid_data):
...
-    #                    Error Callbacks
-    #############################################################
-    def _token_err(self, fail):
-        """errBack for token decryption fail"""
-        fail.trap(InvalidToken)
-        self.set_status(401)
-        log.msg("Invalid token", **self._client_info())
-        self.write("Invalid token")
-        self.finish()
+            response.respond(self)
```
bbangert: I'd prefer to avoid extra entangling of the handler, and make the RouterResponse handling matching the RouterException, ie, a separate method on the handler that takes it and fills it in. The RouterException handling method could be updated so that the logic to dump the contents is the same.

bbangert:
Few minor nits on location chosen for dumping the response object attr's to the handler, other than that it looks good so far!

autopush/main.py
kitcambridge: It's returned in the `Location` header when the app server sends a message.

kitcambridge:
Ready for review.

autopush/router/simple.py
```diff
@@ -142,14 +154,20 @@ def route_notification(self, notification, uaid_data):
...
             self.metrics.increment("router.broadcast.miss")
-            returnValue(self.stored_response(notification))
+            result = yield maybeDeferred(self.stored_response, uaid,
+                                         notification)
+            returnValue(result)
 
         if result.code == 200:
             self.metrics.increment("router.broadcast.save_hit")
-            returnValue(self.delivered_response(notification))
+            result = yield maybeDeferred(self.delivered_response, uaid,
```
bbangert: What happens if this throws an exception here?

autopush/router/simple.py
```diff
@@ -142,14 +154,20 @@ def route_notification(self, notification, uaid_data):
...
+                                         notification)
+            returnValue(result)
         else:
             self.metrics.increment("router.broadcast.miss")
-            returnValue(self.stored_response(notification))
+            result = yield maybeDeferred(self.stored_response, uaid,
```
bbangert: If fernet throws an exception, what happens here?

autopush/router/simple.py
kitcambridge: `fernet.encrypt` shouldn't be throwing exceptions, but that's a risky position to take. We can generate the responses before routing or writing to storage, so we're guaranteed to have a valid URL to send back to the app server.

bbangert:
r+, LGTM!

PR #134, opened 2015-08-19
Omit headers and data for empty messages

### LINKED ISSUE #132: Web Push: Empty messages don't need crypto headers
IIRC, it's okay to omit the `Content-Encoding` and `Encryption` headers for push messages without a body. The DOM API can distinguish these from empty encrypted messages: `pushEvent.data === null` for the former; `pushEvent.data.text() === ""` for the latter.

kitcambridge:
This lets folks use Web Push as if it were Simple Push. For that case, I don't think there's much point to sending an encrypted empty string.

Closes #132.

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -127,7 +127,8 @@ keyid="http://example.org/bob/keys/123;salt="XZwpw6o37R-6qoZjw6KwAw"\
         return result
 
     def send_notification(self, channel=None, version=None, data=None,
-                          use_header=True, status=200, ttl=200):
+                          use_header=True, status=None, ttl=200,
+                          timeout=0.2):
         if not channel:
             channel = random.choice(self.channels.keys())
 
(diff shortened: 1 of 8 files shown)


autopush/tests/test_integration.py
```diff
@@ -183,6 +188,9 @@ def get_notification(self, timeout=0.2):
         except:
             return None
 
+    def get_notification(self, timeout=0.2):
```
bbangert: What's the rationale for refactoring this into two methods?

bbangert:
No changelog entry re: this change?

autopush/tests/test_integration.py
kitcambridge: I wanted `send_notification` to accept a timeout, but I couldn't call `self.get_notification` directly—that caused  `__getattribute__` to wrap it. I also made a change on line 74 to ignore all `_`-prefixed methods. Is there a better way to bypass `__getattribute__`?

bbangert: Yup, the same way __getattribute__ bypasses it, grab the actual one with:
 ``get = object.__getattribute__(self, "get_notification")```

kitcambridge: D'oh, I forgot `__getattribute__` is inherited and you can actually do that. :stuck_out_tongue_winking_eye:

bbangert:
r+

PR #135, opened 2015-08-19
Base64-encode Web Push message blobs

kitcambridge:
...And store those blobs in DynamoDB binary fields.

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -1,6 +1,7 @@
 """Database Interaction"""
 import uuid
 from functools import wraps
+from base64 import urlsafe_b64decode, urlsafe_b64encode
 
 from boto.exception import JSONResponseError
 from boto.dynamodb2.exceptions import (
@@ -12,6 +13,7 @@ from boto.dynamodb2.fields import HashKey, RangeKey, GlobalKeysOnlyIndex
 from boto.dynamodb2.layer1 import DynamoDBConnection
 from boto.dynamodb2.table import Table
 from boto.dynamodb2.types import NUMBER
+from boto.dynamodb.types import Binary
 
 
 def create_router_table(tablename="router", read_throughput=5,
@@ -266,6 +268,12 @@ class Message(object):
     def store_message(self, uaid, channel_id, data, headers, message_id, ttl):
         """Stores a message in the message table for the given uaid/channel with
         the message id"""
+        if data:
+            # Store Base64-encoded payloads as binary fields.
+            if not isinstance(data, bytes):
+                data = data.encode('utf-8')
+            data = Binary(urlsafe_b64decode(data))
+
         self.table.put_item(data=dict(
             uaid=uaid,
             chidmessageid="%s:%s" % (channel_id, message_id),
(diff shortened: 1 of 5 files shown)


kitcambridge:
I'm not sure why Travis is unhappy—I see three `POST` requests and responses, followed by three acks. Where's the `ConnectionRefusedError` coming from?

jrconlin:
Looks like the problem is that travis isn't resolving the hostname in test_webpush_data_delivery_to_disconnected_client, which is odd, since it's localhost. Perhaps the port is being blocked?

autopush/db.py
```diff
@@ -1,6 +1,7 @@
 """Database Interaction"""
 import uuid
 from functools import wraps
+from base64 import urlsafe_b64decode, urlsafe_b64encode
```
bbangert: Why the urlsafe version?

autopush/db.py
```diff
@@ -266,6 +268,12 @@ def all_channels(self, uaid):
     def store_message(self, uaid, channel_id, data, headers, message_id, ttl):
         """Stores a message in the message table for the given uaid/channel with
         the message id"""
+        if data:
+            # Store Base64-encoded payloads as binary fields.
+            if not isinstance(data, bytes):
+                data = data.encode('utf-8')
+            data = Binary(urlsafe_b64decode(data))
```
bbangert: B64 is entirely safe to store as a string, why use a Binary field?

autopush/db.py
```diff
@@ -266,6 +268,12 @@ def all_channels(self, uaid):
     def store_message(self, uaid, channel_id, data, headers, message_id, ttl):
         """Stores a message in the message table for the given uaid/channel with
         the message id"""
+        if data:
+            # Store Base64-encoded payloads as binary fields.
+            if not isinstance(data, bytes):
```
bbangert: Under what conditions does this happen? Is it intentional that we'd pass it possibly bytes, but not?

autopush/db.py
```diff
@@ -306,8 +314,13 @@ def delete_messages_for_channel(self, uaid, channel_id):
...
     def fetch_messages(self, uaid, limit=10):
         """Fetches messages for a uaid"""
-        return self.table.query_2(uaid__eq=uaid, chidmessageid__gt=" ",
-                                  consistent=True, limit=limit)
+        # Eagerly fetches all results in the result set.
+        messages = list(self.table.query_2(uaid__eq=uaid,
+                        chidmessageid__gt=" ", consistent=True, limit=limit))
+        for message in messages:
+            message["data"] = urlsafe_b64encode(message["data"].value)
+            message.mark_clean()
```
bbangert: Why does it need to be marked clean? Do we call save on it somewhere? (It shouldn't auto-save these.... does it?)

autopush/db.py
kitcambridge: Because that's how the headers are encoded, and I wanted the client to use the same decoder for both. :wink: (It's easy to support URL-safe and regular; just thought I'd keep it consistent).

autopush/db.py
bbangert: Why not leave it as b64 when its stored, and keep it a string field? Then we won't need to b64 encode it on the way out.

autopush/db.py
bbangert: Ah, ok.

autopush/db.py
kitcambridge: Sounds good; I'll revert that commit.

bbangert:
r+

bbangert:
Er, no changelog entry needed?

kitcambridge:
> Er, no changelog entry needed?

Sorry, I keep forgetting to add those. Thanks!

PR #137, opened 2015-08-21
Add `/status` to the connection node

### LINKED ISSUE #136: Expose a health check endpoint on the connection node
The Loop folks have a `/__heartbeat__` endpoint that pings our server to verify it's up. Unfortunately, we only expose [`/status`](https://updates.push.services.mozilla.com/status) on the endpoint node, not the connection node. Since Loop already stores the connection node URL in a config value, let's just add `/status` for their use.

kitcambridge:
Closes #136.

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -1,8 +1,10 @@
 """autopush/autoendpoint daemon scripts"""
 import configargparse
 import cyclone.web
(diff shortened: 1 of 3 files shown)


jrconlin:
r+

PR #139, opened 2015-08-26
Fix routing errors for blank updates with headers

kitcambridge:
Fixes a regression from #134.

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -897,18 +897,21 @@ class SimplePushServerProtocol(WebSocketServerProtocol):
             return
 
         if self.use_webpush:
-            self.direct_updates[chid].append(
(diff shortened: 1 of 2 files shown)


autopush/websocket.py
```diff
@@ -897,18 +897,21 @@ def send_notifications(self, update):
...
         if self.use_webpush:
+            response = dict(
+                messageType="notification",
+                channelID=chid,
+                version=version,
+            )
+            data = update.get("data")
+            if data:
+                response["headers"] = update["headers"]
+                response["data"] = update["data"]
```
bbangert: Use the data just extracted rather than fetching the key again.

bbangert:
r+

PR #148, opened 2015-08-27
Expose Web Push headers for CORS requests

kitcambridge:
This lets our tests make cross-origin requests directly.

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -246,6 +246,8 @@ def parse_request_params(request):
 class AutoendpointHandler(cyclone.web.RequestHandler):
     """Common overrides for Autoendpoint handlers"""
     cors_methods = ""
+    cors_request_headers = []
+    cors_response_headers = []
 
     #############################################################
     #                    Cyclone API Methods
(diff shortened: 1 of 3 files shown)


autopush/endpoint.py
```diff
@@ -261,7 +263,15 @@ def prepare(self):
         """Common request preparation"""
         if self.ap_settings.cors:
             self.set_header("Access-Control-Allow-Origin", "*")
-            self.set_header("Access-Control-Allow-Methods", self.cors_methods)
+            if self.cors_methods:
```
jrconlin: Would we ever have occasion to not have values for any of these? If I recall, passing an empty header is equivalent to the header not being present.

autopush/tests/test_endpoint.py
```diff
@@ -435,37 +435,55 @@ def handle_finish(result):
...
         endpoint = self.endpoint
         endpoint.ap_settings.cors = True
         endpoint.prepare()
         endpoint.options(None)
         eq_(endpoint._headers[ch1], "*")
-        eq_(endpoint._headers[ch2], "PUT")
+        eq_(endpoint._headers[ch2], "POST,PUT")
+        eq_(endpoint._headers[ch3], "content-encoding,encryption,"
+            "encryption-key,content-type")
+        eq_(endpoint._headers[ch4], "location")
```
jrconlin: Should we check that passing an empty arg does not set a header (if we're going to the effort to check?)

autopush/endpoint.py
kitcambridge: Good call.

jrconlin:
r+

Issue #151, opened 2015-08-31
Sending 100 ChannelIDs during a register message causes the server to abruptly close connection


jrconlin:
This may be the root of https://bugzilla.mozilla.org/show_bug.cgi?id=1189729, or it may just be another bug.

Tweaking the smoketest to send 10 UUIDs in the ChannelIDs succeeds, sending 1000 causes the connection to be suddenly terminated.

This may be the websocket library, or JSON parser blowing up because of a buffer allocation problem. Need to investigate work arounds, since clients will be sending LOTS of channelIDs. (Clients can discard and fetch new ChannelIDs if a new UAID is sent to them.)

jrconlin:
An interim fix is to increase the max_message_size for cyclone. 

A proper fix is to prevent the Client from including ChannelIDs as part of the incoming Hello message, since the server ignores them anyway. This will reduce data requirements and speed up push connections as a nice benefit.

kitcambridge:
Fantastic find, JR! :tada:

bbangert:
Closed by the PR.

PR #152, opened 2015-08-31
Fix for #151 allow for larger cyclone messages

jrconlin:
Allow for servers to specify larger default message buffer sizes
(Current recommendation is to raise them to 3072 or larger until the
client lands which turns off client send of ChannelIDs as part of the
"hello")

r? @bbangert @kitcambridge


jrconlin:
I would like to promote this for 1.4.1 since we have folks blocked on it. Even if the client push doesn't go out, this would allow the server to keep working correctly.

kitcambridge:
r+

PR #153, opened 2015-09-01
Feature/avoid dict reallocation

### LINKED ISSUE #149: Avoid dict re-allocation increase
Due to the amount of attributes we tack-on to the websocket instance, the underlying dict is re-allocated and grows substantially in size.

To avoid this, we should make a separate class (maybe 'PushState' or something), with a ``__slots__`` attribute listing all our needed things, then attach an instance of that to the websocket instance.

bbangert:
* Refactor of attribute assignment to the Websocket instance to avoid memory
  increases due to Python reallocating the underlying dict datastructure. Issue
  #149.

Closes #149.

@kitcambridge, @jrconlin r?

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -35,6 +35,7 @@ twisted.internet.base.DelayedCall.debug = True
 
 
 def setUp():
+    logging.getLogger('boto').setLevel(logging.CRITICAL)
     boto_path = os.path.join(root_dir, "automock", "boto.cfg")
     boto.config.load_from_path(boto_path)
     if "SKIP_INTEGRATION" in os.environ:  # pragma: nocover
@@ -239,7 +240,7 @@ class IntegrationBase(unittest.TestCase):
 
         # Websocket server
         self._ws_url = "ws://localhost:9010/"
-        factory = WebSocketServerFactory(self._ws_url)
+        factory = WebSocketServerFactory(self._ws_url, debug=False)
         factory.protocol = SimplePushServerProtocol
         factory.protocol.ap_settings = settings
         factory.setProtocolOptions(
@@ -264,7 +265,8 @@ class IntegrationBase(unittest.TestCase):
             (r"/notif/([^\/]+)(/([^\/]+))?", n),
         ],
             default_host=settings.router_hostname,
-            log_function=skip_request_logging
+            log_function=skip_request_logging,
+            debug=False,
         )
         self.ws_website = reactor.listenTCP(9030, ws_site)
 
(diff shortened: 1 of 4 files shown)


autopush/tests/test_integration.py
```diff
@@ -35,6 +35,7 @@
 
 
 def setUp():
+    logging.getLogger('boto').setLevel(logging.CRITICAL)
```
jrconlin: Was this generating too much noise? Curious why we're setting to critical.

bbangert: It was generating a large amount of noise, yes. Under individual nosetests run it was in debug mode, so this is to drop it back to only show critical errors.

jrconlin: Ah, that explains it. Thanks!

jrconlin:
r+

autopush/websocket.py
```diff
@@ -163,42 +238,29 @@ def wrapper(result, *w_args, **w_kwargs):
...
         """Property that uses None if there's no tags due to a DataDog library
         bug"""
-        return self._base_tags if self._base_tags else None
+        return self.ps._base_tags if self.ps._base_tags else None
 
     def log_err(self, failure, **kwargs):
         """Log a twisted failure out through twisted's log.err"""
         log.err(failure, **kwargs)
 
-    def pauseProducing(self):
```
kitcambridge: Are `pauseProducing` and friends are only called by us, or can Twisted call them, too? (Say, if we're sending a batch of updates to a slow client).

autopush/websocket.py
```diff
@@ -233,50 +295,19 @@ def verifyNuke(self):
...
+                                  tags=self.base_tags)
 
     @log_exception
     def onConnect(self, request):
         """autobahn onConnect handler for when a connection has started"""
         track_object(self, msg="onConnect Start")
+        self.ps = PushState(settings=self.ap_settings, request=request)
 
         # Setup ourself to handle producing the data
         self.transport.bufferSize = 2 * 1024
```
kitcambridge: Not related to this PR, but how does this differ from `maxFramePayloadSize`?

kitcambridge:
r+

autopush/websocket.py
bbangert: In onConnect, the PushState object created is registered with the protocol as a IProducer. So it will be calling these.

autopush/websocket.py
bbangert: bufferSize for the transport is the point at which it will call the producers pauseProducing method to halt output. So if the kernel buffer is full, and there's 2kb backed up, it will cause pauseProducing until it can empty out the buffers.

autopush/websocket.py
jrconlin: As I understand, they're a part of the IProducer interface, and thus are called by twisted after the Producer is registered. see https://github.com/mozilla-services/autopush/pull/153/files#diff-eff17175be11d1f3b2faddd0c7c93cd9L246 I believe that PushState simply records the state transitions. 

@bbangert, obviously, knows better.

PR #156, opened 2015-09-02
Add option for closing handshake timeout

kitcambridge:
"0" disables Autobahn's closing handshake timeout. This still nukes the connection if it hasn't been closed within 5 seconds, which seems reasonable.

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -172,6 +172,10 @@ def _parse_connection(sysargs):
                         help="The maximum size that messages from client " +
                         "can be (e.g. header, data, json formatting, etc.)",
                         default=2048, type=int, env_var="MAX_MESSAGE_SIZE")
+    parser.add_argument('--close_handshake_timeout',
+                        help="The WebSocket closing handshake timeout. Set to "
+                        "0 to disable.", default=0, type=int,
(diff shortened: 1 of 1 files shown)


jrconlin:
r+

autopush/main.py
```diff
@@ -172,6 +172,10 @@ def _parse_connection(sysargs):
                         help="The maximum size that messages from client " +
                         "can be (e.g. header, data, json formatting, etc.)",
                         default=2048, type=int, env_var="MAX_MESSAGE_SIZE")
+    parser.add_argument('--close_handshake_timeout',
+                        help="The WebSocket closing handshake timeout. Set to "
+                        "0 to disable.", default=0, type=int,
+                        env_var="CLOSE_HANDSHAKE_TIMEOUT")
```
jrconlin: According to the docs: http://autobahn.ws/python/reference/autobahn.websocket.html#autobahn.websocket.protocol.WebSocketClientProtocol 1 is the default. I presume that 0 means that no timeout is set and that thing would hang forever?

bbangert: It would if I hadn't tacked on a double-check that runs 5 seconds later to verify its closed, yea. (The timer I tack on is in the sendClose).

bbangert:
r+

Issue #155, opened 2015-09-01
Restore missing metrics and log messages


bbangert:
Between 1.2.3 and 1.4.1, several important metrics and log messages have disappeared. This is most likely a result of the router refactoring and other various changes. These metrics/log messages should be restored.

#### Known missing data

Metrics:
- updates.handled timer tracking time to process an update

Log messages:
- Successful delivery
- Router miss, message stored

bbangert:
Closed by #157.

PR #157, opened 2015-09-02
Bug/metrics restore

### LINKED ISSUE #154: Connection close logging should include user-agent
Connection closing is logged, but fails to include the user-agent, it should.

bbangert:
* Don't cancel a deferred that was already called.
* Restore logging of simplepush successfull/stored delivery based on status.
* Restore updates.handled endpoint timer to track time to deliver.
* Remove logging of connection closing.
* Add changelog for close connection timer.
* Remove metric logging of sendClose timer since its now the default.

Closes #154, #155.

@kitcambridge, @jrconlin r?

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -279,11 +279,6 @@ class SimplePushServerProtocol(WebSocketServerProtocol):
         if hasattr(self, "_shutdown_ran"):
             return
 
-        # Uh-oh, we have not been shut-down properly, report detailed data
-        self.ps.metrics.increment("client.error.sendClose_failed",
-                                  tags=self.base_tags)
-        log.msg("sendClose failed to result in onClose", state=str(self.state))
-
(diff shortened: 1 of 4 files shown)


autopush/websocket.py
```diff
@@ -279,11 +279,6 @@ def nukeConnection(self):
         if hasattr(self, "_shutdown_ran"):
             return
 
-        # Uh-oh, we have not been shut-down properly, report detailed data
-        self.ps.metrics.increment("client.error.sendClose_failed",
```
kitcambridge: This might still be helpful, actually...I think we'll only get here if the client didn't respond to the close frame, right?

bbangert: Good point, will re-add it.

jrconlin:
r+

kitcambridge:
r+

PR #158, opened 2015-09-02
Up close handshake and add chid logging to endpoint

bbangert:
* Up default close handshake timer to 10 seconds for slower clients.
* Add channel id logging to endpoint.

@kitcambridge, @jrconlin r?


kitcambridge:
r+

jrconlin:
r+ (mumbles something about config values)
Not merging pending review of the travis failure.

PR #164, opened 2015-09-04
Use the message token as the message ID

kitcambridge:
I realized we don't need a separate message ID; we can just use the encrypted token as the DynamoDB hash key. Since each token uses a 128-bit IV, it's as unique as a UUID. We still decrypt and verify the token to extract the UAID and channel ID; we just use the original, encrypted form as the message ID.

This is particularly helpful for receipts, where we need to send push promises containing the message URL. By reusing the token as the message ID (and version), we avoid having to store and pass around the URL. It also makes for shorter message URLs, and drops `prepare_notification`—which I didn't really like.

@bbangert, @jrconlin r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -398,8 +398,9 @@ class MessageHandler(AutoendpointHandler):
         successfully routed to a client as direct updates, but not delivered
         yet, will not be dropped.
         """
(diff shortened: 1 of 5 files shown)


bbangert:
r+

PR #165, opened 2015-09-10
Check for stale connection nodes when routing

### LINKED ISSUE #163: Check for stale connection nodes when routing
We [maintain an LRU cache](https://github.com/mozilla-services/autopush/blob/master/autopush/router/simple.py#L37) of stale nodes, but don't actually check it before routing.

kitcambridge:
Closes #163.

@bbangert r?


bbangert:
r+

PR #167, opened 2015-09-11
Reset UAIDs for clients that change their router type

kitcambridge:
If an existing client opts in to data delivery via the `use_webpush` flag, we should reset its ID so that its existing subscriptions are invalidated. (It's possible for the client to keep track of this, too; I just thought it would be cleaner for the server to handle it, since we already store the router type).

@bbangert, @jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,7 +3,7 @@ Autopush Changelog
 ==================
 
 1.7.0 (**dev**)
-==================
+===============
 
 Features
 --------
@@ -12,6 +12,10 @@ Features
   Some devices which use SimplePush routing offer a feature to wake on
   a carrier provided UDP ping. See issue #106 for details.
 
+Bug Fixes
+---------
+
+* Reset UAIDs for clients that change their router type. PR #167.
 
 1.6.0 (2015-09-14)
 ==================
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -118,7 +118,8 @@ def preflight_check(storage, router):
 
     # Store a router entry, fetch it, delete it
     router.register_user(dict(uaid=uaid, node_id=node_id,
-                              connected_at=connected_at))
+                              connected_at=connected_at,
+                              router_type="simplepush"))
     item = router.get_uaid(uaid)
     assert item.get("node_id") == node_id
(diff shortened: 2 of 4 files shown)


jrconlin:
r+

kitcambridge:
A potential side effect is that we'll now issue new IDs to flapping clients (where `connected_at >= :connected_at`), instead of failing the handshake. Not sure if that's a concern.

Since we return `ALL_OLD`, another approach is to check if the `router_type` matches, and issue a new ID if it doesn't. But that incurs two writes, and I expect many clients to hit this as they update.

jrconlin:
The registration logic appears to be 
If either the router_type is unknown or matches and either the node id is unknown or it connected earlier, which pretty much leads to the following states being true:
* Router_type and node_id are unknown
* Router_type is simplepush and node_id unknown
* Router_type is unknown and it connected earlier (not sure how this state could happen)
* Router_type is simplepush and it connected earlier. 

That third case bothers me, but I don't believe it's a state we can get into. Requesting a client use a new UAID inserts an effective break because it forces a channel refresh and transmit. In essence, the server is kinda forcing the client to do the right action in any case, and should limit flapping better than just killing the handshake.

CHANGELOG.rst
```diff
@@ -19,6 +19,7 @@ Bug Fixes
   months it was in, indicating that abortConnection is 100% effective.
   Issue #161.
 * Rename `SimplePushServerProtocol` to `PushServerProtocol`. Issue #117.
+* Reset UAIDs for clients that change their router type. PR #167.
```
bbangert: Needs to be moved to 1.7.0 now. Sorry bout that.

autopush/db.py
```diff
@@ -375,10 +376,16 @@ def register_user(self, data):
...
-            ])
+            cond = (
+                # noqa
+                "("
+                    "attribute_not_exists(router_type) or"
+                    "(router_type = :router_type)"
+                ") and ("
+                    "attribute_not_exists(node_id) or "
+                    "(connected_at < :connected_at)"
+                ")"
```
bbangert: Triple-quote it?

bbangert:
r+

PR #168, opened 2015-09-11
Use Codecov for code coverage

kitcambridge:
Coveralls hasn't been happy for a while. The Esprima folks seem to like Codecov; let's give it a try.


jrconlin:
r+

PR #172, opened 2015-09-18
Ensure `all_channels` always returns an empty set

kitcambridge:
Fixes an uncaught `TypeError` in production.

@jrconlin r?


jrconlin:
r+

PR #176, opened 2015-09-28
Hotfix/1.6.0 issues

bbangert:
* Use proper client connected_at in delete handler.
* Trap provisioned exception error in websocket handler.
* Clear the node entry for a client even if we know the node is dead.
* Ignore Cancelled exceptions.


kitcambridge:
r+

PR #181, opened 2015-10-12
Code fixes for env and hash

### LINKED ISSUE #50: Services should self identify

jrconlin:
* Fix #50 patch to set env in Hello handler, not send_json
* Fix bad hash encoding in endpoint

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -688,7 +688,7 @@ class RegistrationHandler(AutoendpointHandler):
         """Called after the endpoint was made and should be returned to the
         requestor"""
         if new_uaid:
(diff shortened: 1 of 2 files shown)


bbangert:
r+

PR #187, opened 2015-10-16
set autoendpoint to read proper config values

### LINKED ISSUE #186: Fix autoendpoint to use proper config value rather than hostname
autoendpoint defaults to using HOSTNAME over the configuration value. This can lead to problems on EC2 instances where HOSTNAME is already defined by the environment.

Do the right thing and use the endpoint_hostname first, then fail over to hostname.

jrconlin:
* endpoint_hostname and endpoint_scheme now shared values
* default back to previous behaviors

fixes #186
@bbangert r?

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -380,10 +380,11 @@ def endpoint_main(sysargs=None):
     """Main entry point to setup an endpoint node, aka the autoendpoint
     script"""
     args, parser = _parse_endpoint(sysargs)
+    scheme = args.endpoint_scheme or \
+        "https" if args.ssl_key else "http"
     settings = make_settings(
         args,
-        endpoint_scheme=args.endpoint_scheme or "https"
-        if args.ssl_key else "http",
+        endpoint_scheme=scheme,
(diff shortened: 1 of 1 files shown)


autopush/main.py
```diff
@@ -382,8 +382,9 @@ def endpoint_main(sysargs=None):
     args, parser = _parse_endpoint(sysargs)
     settings = make_settings(
         args,
-        endpoint_scheme="https" if args.ssl_key else "http",
-        endpoint_hostname=args.hostname,
+        endpoint_scheme=args.endpoint_scheme or "https"
+        if args.ssl_key else "http",
```
bbangert: Nit: the wrapping on this is confusing. Indent?

bbangert:
r+ with nit

autopush/main.py
jrconlin: flake8 was complaining when I had indentation, but I agree that it's confusing. Let me try a few things to see if I can make it better.

bbangert: Did you want to try a different way, or we could let this in?

bbangert:
LGTM!

PR #192, opened 2015-10-20
WebPush Message Collapsing

### LINKED ISSUE #141: Message collapsing
See webpush-wg/webpush-protocol#12 for the strawman. Current Simple Push usage patterns indicate this might be helpful.

bbangert:
* Add Push message update mechanism. Issue #141.
* Update the TTL mechanism so that the original posted time is stored along with the TTL for proper message updating.
* Require an update_id on the message to ensure the client doesn't delete an updated message.

Closes #141.

@jrconlin r?


autopush/router/webpush.py
```diff
@@ -60,7 +60,8 @@ def _send_notification(self, uaid, node_id, notification):
         """
         payload = {"channelID": notification.channel_id,
                    "version": notification.version,
-                   "ttl": notification.ttl+int(time.time()),
+                   "ttl": notification.ttl,
+                   "timestamp": int(time.time()),
```
jrconlin: If we're not doing vector clocks or clock sync, I wonder if we really need to have a timestamp here? TTL could be set as the storage time+TTL on the machine that will reap.

jrconlin: Ah, nevermind. I see that we're using timestamp in testing.

jrconlin:
r+

PR #193, opened 2015-10-21
Make errors in the HTTP endpoint usable

### LINKED ISSUE #178: Provide actionable 4** error messages
We should provide more actionable error messages in autoendpoint about why in particular a 4** error response was returned, rather than just "400 - Invalid Arguments" with an empty body failing to indicate what was missing.

This should also cover the case where a router type we don't recognize is specified.

bbangert:
* Add actionable JSON errors to the Endpoint responses. Issue #178.

Closes #178.

@jrconlin r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -57,7 +57,7 @@ The current application-level error numbers are:
 * status code 404, errno 107: message not found
 * status code 400, errno 108: router type is invalid
 * status code 401, errno 109: invalid authentication
-* status code 503, errno 201: service temporailly unavailable due to high
+* status code 503, errno 201: service temporarily unavailable due to high
   load (use exponential back-off for retries)
 * status code 503, errno 202: temporary failure, immediate retry ok
 * status code 500, errno 999: unknown error
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -662,6 +662,7 @@ class EndpointHandler(AutoendpointHandler):
         d = deferToThread(self.ap_settings.fernet.encrypt, ':'.join([
             'm', self.uaid, self.chid]).encode('utf8'))
         d.addCallback(self._route_notification, result, data, ttl)
+        return d
 
     def _route_notification(self, version, result, data, ttl=None):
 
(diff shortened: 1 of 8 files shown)


autopush/endpoint.py
```diff
@@ -25,6 +28,41 @@
...
+
+* status code 400, errno 101: missing necessary crypto keys
+* status code 404, errno 102: invalid URL endpoint
+* status code 404, errno 103: expired URL endpoint
+* status code 413, errno 104: data payload is too large
+* status code 404, errno 105: endpoint became unavailable during request
+* status code 404, errno 106: invalid subscription
+* status code 404, errno 107: message not found
+* status code 400, errno 108: router type is invalid
+* status code 401, errno 109: invalid authentication
```
jrconlin: seems redundant to have a discrete errno for an existing well defined status response, but whatever.

autopush/endpoint.py
```diff
@@ -25,6 +28,41 @@
...
+* status code 503, errno 201: service temporailly unavailable due to high
```
jrconlin: typpo temporarily

autopush/tests/test_endpoint.py
```diff
@@ -173,7 +170,7 @@ def test_update_too_much_data(self):
 
         def handle_finish(result):
             self.assertTrue(result)
-            self.status_mock.assert_called_with(401)
+            self.status_mock.assert_called_with(413)
```
jrconlin: should we have the tests check for the errnos as well if we're publishing them and clients are relying on them?

autopush/endpoint.py
bbangert: Firefox Accounts has it, they are well defined, but they do mean different things. Someone with an errno can get more detail and provide us with more insight into helping them if we know more precisely where in our code the error came from (most of these are unique to one bit of code).

autopush/tests/test_endpoint.py
bbangert: I considered that, but it seems like it'd create a huge amount of work to do any tweak later. I did add a test to make sure the error has JSON. I can have it verify the rest of the attributes too, as long as it works once, since all the others call the same function I don't see why it wouldn't be fine for them.

bbangert: I think there could be some better checks though, I'll add a few more to ensure the error is formatted right, all the fields are present, etc.

bbangert:
Ok, in theory this should be golden now.

jrconlin:
r+ (pending travis)

PR #196, opened 2015-10-21
Add server generated SenderIDs and ChannelIDs for GCM clients

jrconlin:
Randomly select a GCM senderid from a list stored in an S3 bucket and include it with the initial registration request. This was proposed by the client team as a way to allow for GCM SenderIDs to be rotated in case an ID is disqualified for any reason. 

In addition, if the client doesn't include a ChannelID in the incoming registration request, there's no reason why we can't generate and return one on the server. This simplifies things on the client.

Since the GCM router requires modifying the registration response, it was decided that all routers should be given that option.

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -229,7 +229,8 @@ class Message(object):
         db_key = self.encode({"uaid": uaid, "chidmessageid": " "})
         # Generate our update expression
         expr = "ADD chids :channel_id"
-        expr_values = self.encode({":channel_id": set([channel_id])})
+        expr_values = self.encode({":channel_id": set([channel_id]),
+            })
         conn.update_item(
             self.table.table_name,
             db_key,
@@ -370,6 +371,7 @@ class Router(object):
             exceeds throughput.
 
         """
+        ## Fetch a senderid for this user
         conn = self.table.connection
         db_key = self.encode({"uaid": data.pop("uaid")})
         # Generate our update expression
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -236,8 +236,7 @@ class Message(object):
         db_key = self.encode({"uaid": uaid, "chidmessageid": " "})
         # Generate our update expression
         expr = "ADD chids :channel_id"
-        expr_values = self.encode({":channel_id": set([channel_id]),
-            })
+        expr_values = self.encode({":channel_id": set([channel_id])})
         conn.update_item(
             self.table.table_name,
             db_key,
@@ -432,7 +431,7 @@ class Router(object):
             exceeds throughput.
 
         """
-        ## Fetch a senderid for this user
+        # Fetch a senderid for this user
         conn = self.table.connection
         db_key = self.encode({"uaid": data.pop("uaid")})
         # Generate our update expression
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -111,8 +111,8 @@ def preflight_check(storage, router):
     Failure to run correctly will raise an exception.
 
     """
-    uaid = str(uuid.uuid4())
-    chid = str(uuid.uuid4())
+    uaid = uuid.uuid4().hex
+    chid = uuid.uuid4().hex
     node_id = "mynode:2020"
     connected_at = 0
     version = 12
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -620,7 +620,7 @@ class RegistrationHandler(AutoendpointHandler):
             d.addCallback(router.register, params.get("data", {}))
             d.addCallback(self._save_router_data, router_type)
             d.addCallback(self._make_endpoint)
-            d.addCallback(self._return_endpoint, new_uaid)
+            d.addCallback(self._return_endpoint, new_uaid, router)
             d.addErrback(self._router_fail_err)
             d.addErrback(self._response_err)
             d.callback(uaid)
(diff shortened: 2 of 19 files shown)


autopush/db.py
```diff
@@ -431,6 +431,7 @@ def register_user(self, data):
             exceeds throughput.
 
         """
+        # Fetch a senderid for this user
```
bbangert: Where does this fetch occur?

autopush/endpoint.py
```diff
@@ -744,8 +744,9 @@ def post(self, uaid=""):
         self.start_time = time.time()
         self.add_header("Content-Type", "application/json")
         params = self._load_params()
+        # If the client didn't provide a CHID, make one up.
         if "channelID" not in params:
-            return self._error(400, "Invalid arguments")
+            params["channelID"] = uuid.uuid4()
```
bbangert: Shouldn't this be a uuid.uuid4().hex so that its a uuid string rather than an object?

autopush/main.py
```diff
@@ -219,6 +220,14 @@ def _parse_endpoint(sysargs):
...
     parser.add_argument('--cors', help='Allow CORS PUTs for update.',
                         type=bool, default=False, env_var='ALLOW_CORS')
+    parser.add_argument('--s3_bucket', help='S3 Bucket for SenderIDs',
+                        type=str, default=DEFAULT_BUCKET,
+                        env_var='S3-BUCKET')
+    parser.add_argument('--senderid_expry', help='Cache expry for senderIDs',
+                        type=int, default=SENDERID_EXPRY,
+                        env_var='SENDERID_EXPRY')
+    parser.add_argument('--senderid_list', help='SenderIDs to load to S3',
+                        type=int, default=[], action='append')
```
bbangert: type=int, but default is [] ?

autopush/router/interface.py
```diff
@@ -56,6 +56,15 @@ def register(self, uaid, routing_data):
         """
         raise NotImplementedError("register must be implemented")
 
+    def amend_msg(self, msg):
+        """Modify an outbound response message to include router info
+
+        Some routers may require additional info to be returned to clients.
+
+        :param msg: A dict of the response data to be sent to the client
```
bbangert: Nit: needs one blank line before trailing """

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
bbangert: Not: The project style is to not have the header per file.

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
...
+
+from boto.exception import S3ResponseError
```
bbangert: No module docs describing this?

autopush/endpoint.py
jrconlin: it would eventually get serialized to a dashed string, but I'm fine explicitly using hex. We should probably also fix the uaid fetch later to do the same so we're consistent.

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
...
+import time
+import json
+import random
+
+# re-read from source every 15 minutes or so.
+SENDERID_EXPRY = 15*60
+DEFAULT_BUCKET = "org.mozilla.services.autopush"
+
+
+class SenderIDs(object):
```
bbangert: No class/method docs for any of this?

autopush/main.py
jrconlin: It's a list of integers, where an empty list is valid. so yeah. If there's a different type declaration for that, I'm fine switching.

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from boto.exception import S3ResponseError
+from boto.s3.connection import S3Connection
+from boto.s3.key import Key
+
+import time
+import json
+import random
```
bbangert: Nit: PEP8 ordering is std lib imports, then 3rd party, then package.

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
...
+        """Return a list of senderIDs, refreshing if required """
+        if time.time() > self._updated + self._expry:
+            self._refresh()
+        return self._senderIDs
+
+    def getID(self):
+        self._refresh()
+        return random.choice(self._senderIDs)
+
+    def __init__(self, args):
```
bbangert: Nit: init is usually the first method.

autopush/db.py
jrconlin: The senderID fetch occurs at REST registration for a GCM client. In the code, the fetch occurs in autopush.router.GCMRouter.register (where it's added to both the instance and the router_data for storage), and then reflected back to the client via the new amend_msg() call.

autopush/endpoint.py
bbangert: Agreed.

autopush/main.py
bbangert: Not that I'm aware of, I figured that might be what it is.

autopush/senderids.py
```diff
@@ -0,0 +1,74 @@
...
+        self.conn = S3Connection()
+        self.ID = args.get("s3_bucket", DEFAULT_BUCKET)
+        self._expry = args.get("senderid_expry", SENDERID_EXPRY)
+        senderIDs = args.get("senderid_list", [])
+        if len(senderIDs):
```
bbangert: len isn't needed, if its an empty list it'll eval to false.

autopush/tests/test_senderids.py
```diff
@@ -0,0 +1,66 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
bbangert: Nit: no per-file header

autopush/tests/test_senderids.py
```diff
@@ -0,0 +1,66 @@
...
+
+from autopush.senderids import SenderIDs
+from mock import Mock
+from boto.exception import S3ResponseError
+from moto import mock_s3, mock_dynamodb2
+from nose.tools import eq_, ok_
+from twisted.trial import unittest
+
+import json
```
bbangert: Nit: PEP8 imports

bbangert:
r+ pending nits, nicely done!

PR #206, opened 2015-10-23
Convert arg handlers to ConfigArgParse std.

### LINKED ISSUE #205: Remove special boolean processing for arguments
ConfigArgParse has a significant problem with foo=True. 
So, don't do that. 
The proper way to specify arguments using ConfigArgParse is to note if having the argument appear is a positive or negative and deal with it that way.

NOTE: THIS WILL ALTER HOW PUPPET CONFIGS NEED TO BE WRITTEN!

jrconlin:
ConfigArgParse has a serious issue with
foo=True
it wants a flag to simply exist or not to indicate value.

* Remove str2bool processing
* Changed default behavior to always enable CORS for endpoints

_**NOTE: THIS WILL REQUIRE MODIFICATION OF THE PUPPET DEPLOYMENT SCRIPT**_

resolves #205 

@bbangert r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,12 @@ Bug Fixes
 WebPush
 -------
 
+Backwards Incompatibilities
+---------------------------
+* Do not specify values for boolean flags.
+* 'cors' is now enabled by default. In it's place use --nocors if you wish
+  to disable CORS. Please remove "cors" flag from configuration files.
+
 1.7.1 (2015-10-23)
 ==================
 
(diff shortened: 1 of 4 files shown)


jrconlin:
And since I can't do a PR to the puppet repo: 

``` diff
diff --git a/autopush/modules/autopush/manifests/settings/autoendpoint.pp b/auto
index ab849b7..e88f3ef 100644
--- a/autopush/modules/autopush/manifests/settings/autoendpoint.pp
+++ b/autopush/modules/autopush/manifests/settings/autoendpoint.pp
@@ -1,6 +1,5 @@
 # autopush autoendpoint settings
 class autopush::settings::autoendpoint(
-  $cors = '1',
   $hostname = 'localhost',
   $port = '8082',
 ) {
```

bbangert:
There's no change-log. The change-log is where we'd record a 'Backwards Incompatibility' blurb about how all users need to change the config setting.

bbangert:
r+

PR #213, opened 2015-10-27
Bug/issue 194

### LINKED ISSUE #194: TTL should always be included in WebPush endpoint response
It can be somewhat non-obvious that not including a TTL header in the request will result in a TTL of 0. We also need to cap the TTL to a max.

The WebPush spec says this should be handled by returning the TTL in the response, with the capped value. If no TTL is included, we should respond with TTL: 0 as a header in the response to make it more obvious this happened.

bbangert:
* Always include TTL header in response to a WebPush notification. Issue #194.

Closes #194.

@jrconlin r?


jrconlin:
r+

PR #221, opened 2015-11-04
Add Auth token handling for Sender IDs

### LINKED ISSUE #220: GCM: Fix SenderID list to include associated Auth token
SenderIDs and Auth tokens are associated and must be used as a pair.

jrconlin:
Resolves #220 

In addition, the following were added to help GCM debugging:

Added the flags:
--gcm_enabled - Enable GCM bridge
--s3_bucket=None - will skip using remote S3 (to resolve issues with moto)
--human_logs - Render stdout logs in a more human readable format

Removed the flags:
--gcm_apikey

* Changed the format of the SenderID list
* Updated the docs to fix errors and reflect new state.

@bbangert r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,16 +11,16 @@ Features
 * Server provided SenderID values for GCM router using clients
   The GCM router will randomly select one of a list of SenderIDs stored in
   S3 under the "org.mozilla.services.autopush"/"senderids" key. The values can
-  be loaded into S3 either via the S3 console, or by runnig an instance of
+  be loaded into S3 either via the S3 console, or by running an instance of
   autopush and passing the values as the "senderid_list" argument. Issue #185.
 * REST Registration will now return a valid ChannelID if one is not specified.
-  Issue #182
+  Issue #182.
 * Add hello timeout. Issue #169.
-* Add .editorconfig for consistent styling in editors. Issue #218.
+* Add .editorconfig for consistent styling in editors. Issue #2.
 * Added --human_logs to display more human friendly logging
 * If you specify the --s3_bucket=None, the app will only use local memory
   and will not call out to the S3 repository. It is STRONGLY suggested that
-  you specify the full --senderid_list data set.
+  you specify the full --senderid_list data set
 
 =======
 
@@ -33,7 +33,7 @@ Bug Fixes
   exceptions ignored. Issue #208.
 * Fix improper attribute reference in delete call. Issue #211.
 * Always include TTL header in response to a WebPush notification. Issue #194.
-* Fixed issue with local senderid data cache. (discovered while debugging)
+* Fixed issue with local senderid data cache. (discovered while debugging.)
 
 WebPush
 -------
@@ -46,7 +46,7 @@ Backwards Incompatibilities
 * Do not specify --gcm_apikey. Instead, store the API key and senderid as
   values in S3. The data may still be written as a JSON string such as:
   ' "`_senderID_`": {"auth": "`_api_key`"}}'
-  activate the GCM bridge by specifying --gcm_enabled
+  activate the GCM bridge by specifying --gcm_enabled.
 
 1.7.2 (2015-10-24)
 ==================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,11 +16,11 @@ Features
 * REST Registration will now return a valid ChannelID if one is not specified.
   Issue #182.
 * Add hello timeout. Issue #169.
-* Add .editorconfig for consistent styling in editors. Issue #2.
-* Added --human_logs to display more human friendly logging
+* Add .editorconfig for consistent styling in editors. Issue #218.
+* Added --human_logs to display more human friendly logging.
 * If you specify the --s3_bucket=None, the app will only use local memory
   and will not call out to the S3 repository. It is STRONGLY suggested that
-  you specify the full --senderid_list data set
+  you specify the full --senderid_list data set.
 
 =======
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,11 @@ Features
 * REST Registration will now return a valid ChannelID if one is not specified.
   Issue #182
 * Add hello timeout. Issue #169.
+* Added --human_logs to display more human friendly logging
+* If you specify the --s3_bucket=None, the app will only use local memory
+  and will not call out to the S3 repository. It is STRONGLY suggested that
+  you specify the full --senderid_list data set.
+
 
 Bug Fixes
 ---------
@@ -26,6 +31,7 @@ Bug Fixes
   exceptions ignored. Issue #208.
 * Fix improper attribute reference in delete call. Issue #211.
 * Always include TTL header in response to a WebPush notification. Issue #194.
+* Fixed issue with local senderid data cache. (discovered while debugging)
 
 WebPush
 -------
@@ -35,6 +41,10 @@ Backwards Incompatibilities
 * Do not specify values for boolean flags.
 * 'cors' is now enabled by default. In it's place use --nocors if you wish
   to disable CORS. Please remove "cors" flag from configuration files.
+* Do not specify --gcm_apikey. Instead, store the API key and senderid as
+  values in S3. The data may still be written as a JSON string such as:
+  ' "`_senderID_`": {"auth": "`_api_key`"}}'
+  activate the GCM bridge by specifying --gcm_enabled
 
 1.7.2 (2015-10-24)
 ==================
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -760,7 +760,8 @@ class RegistrationHandler(AutoendpointHandler):
         new_uaid = False
         if uaid:
             if not self._validate_auth(uaid):
-                return self._error(401, "Invalid Authentication")
+                return self._write_response(
+                    401, 109, message="Unauthorized")
         else:
             # No UAID supplied, make our own
             uaid = uuid.uuid4().hex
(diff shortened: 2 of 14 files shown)


jrconlin:
Will add add'l human_logs coverage point.

CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Features
   Issue #182
 * Add hello timeout. Issue #169.
 * Add .editorconfig for consistent styling in editors. Issue #218.
+* Added --human_logs to display more human friendly logging
```
bbangert: Woooooo!

bbangert: nit: period at the end

CHANGELOG.rst
```diff
@@ -27,6 +33,7 @@ Bug Fixes
   exceptions ignored. Issue #208.
 * Fix improper attribute reference in delete call. Issue #211.
 * Always include TTL header in response to a WebPush notification. Issue #194.
+* Fixed issue with local senderid data cache. (discovered while debugging)
```
bbangert: nit: period

CHANGELOG.rst
```diff
@@ -36,6 +43,10 @@ Backwards Incompatibilities
 * Do not specify values for boolean flags.
 * 'cors' is now enabled by default. In it's place use --nocors if you wish
   to disable CORS. Please remove "cors" flag from configuration files.
+* Do not specify --gcm_apikey. Instead, store the API key and senderid as
+  values in S3. The data may still be written as a JSON string such as:
+  ' "`_senderID_`": {"auth": "`_api_key`"}}'
+  activate the GCM bridge by specifying --gcm_enabled
```
bbangert: nit: period

autopush/endpoint.py
```diff
@@ -760,7 +760,8 @@ def post(self, uaid=""):
         new_uaid = False
         if uaid:
             if not self._validate_auth(uaid):
-                return self._error(401, "Invalid Authentication")
+                return self._write_response(
+                    401, 109, message="Unauthorized")
```
bbangert: Nice catch on fixing these.

autopush/main.py
```diff
@@ -398,6 +405,12 @@ def endpoint_main(sysargs=None):
     args, parser = _parse_endpoint(sysargs)
     scheme = args.endpoint_scheme or \
         "https" if args.ssl_key else "http"
+    senderid_list = None
+    if args.senderid_list:
+        try:
+            senderid_list = json.loads(args.senderid_list)
+        except (ValueError, TypeError), x:
+            log.err("Invalid JSON specified for senderid_list. Ignoring.", x)
```
bbangert: Does this error out of the startup? Should it?

autopush/router/gcm.py
```diff
@@ -47,6 +47,11 @@ def route_notification(self, notification, uaid_data):
 
     def _route(self, notification, router_data):
         """Blocking GCM call to route the notification"""
+        if not self.gcm:
```
bbangert: This is a dangerous operation, as its running in a thread-pool with no mutex lock. Means that we could end up creating this object multiple times. How come it couldn't be created at start-up?

autopush/router/gcm.py
```diff
@@ -47,6 +47,11 @@ def route_notification(self, notification, uaid_data):
...
+            self.gcm = gcmclient.GCM(
+                router_data.get("creds",
+                                {}).get("auth",
+                                        self._default_auth))
```
bbangert: If this really needs to be done in the thread-pool, it should be wrapped with a:

```python
with self._gcm_lock:
    if not self.gcm:
....
```

autopush/tests/test_endpoint.py
```diff
@@ -721,6 +723,12 @@ def setUp(self):
         d = self.finish_deferred = Deferred()
         self.reg.finish = lambda: d.callback(True)
 
+    def _check_error(self, code, errno, error):
+        d = json.loads(self.write_mock.call_args[0][0])
+        eq_(d.get("code"), code)
+        eq_(d.get("errno"), errno)
+        eq_(d.get("error"), error)
+
```
bbangert: Oh, I should've done this when I added these errors. Nice.

autopush/senderids.py
```diff
@@ -78,17 +106,31 @@ def senderIDs(self):
...
+    def get(self, id):
+        """Return the associated record for a given SenderID"""
+        record = self._senderIDs.get(id)
+        if record:
+            record["senderID"] = id
+        return record
+
     def getID(self):
         """Return a randomly selected SenderID, refreshing if required"""
         self._refresh()
```
bbangert: Since getID is called in the GCM register function, which is called in a thread-pool, the refresh operation itself should be mutex locked and/or cached so that only a single refresh executes at a time.

bbangert:
So, for the two bits that are blocking operations updating the same object in a thread-pool... my original thought was to replace it with an object that can return the last 'good' value while updating the cached thing.

However, given the load's we anticipate, so many checks on the current time actually kinda worry me. Then having to do a mutex lock check, etc, ugh.

I think the best solution instead, is to make a little abstraction that occasionally updates the object using a function, every N seconds. Here's what that might look like:

```python
from twisted.internet import task
from twisted.internet import reactor


class UpdatingThreadFunction(object):
    def __init__(self, update_func, func, *args, **kwargs):
        self._update_func = update_func
        self._func = func
        self._args = args
        self._kwargs = kwargs
        self._update_func(func(*args, **kwargs))
        self._d = None

    def run(self):
        if self._d:
            # Still running, return
            return

        self._d = deferToThread(func, *args, **kwargs)
        self._d.addCallback(self._success)
        self._d.addErrback(self._error)

    def _success(self, result):
        try:
            self._update_func(result)
        except:
            log.err()
        self._d = None

    def _error(self, failure):
        log.err(failure)
        self._d = None

# Using it
def my_update_func(result):
    self._gcm = result

updater = UpdatingThreadFunction(my_update_func, other_func, ...)
l = task.LoopingCall(updater.run)
l.start(1.0) # call every second
```

This is just a quick hack version, but should be about right I think.

autopush/main.py
jrconlin: Hrm, you're right. If there's a bad command line config option, the app should bail and not forgive.

jrconlin:
@bbangert This should be ready to go.

CHANGELOG.rst
```diff
@@ -11,12 +11,18 @@ Features
...
   S3 under the "org.mozilla.services.autopush"/"senderids" key. The values can
-  be loaded into S3 either via the S3 console, or by runnig an instance of
+  be loaded into S3 either via the S3 console, or by running an instance of
   autopush and passing the values as the "senderid_list" argument. Issue #185.
 * REST Registration will now return a valid ChannelID if one is not specified.
-  Issue #182
+  Issue #182.
 * Add hello timeout. Issue #169.
-* Add .editorconfig for consistent styling in editors. Issue #218.
+* Add .editorconfig for consistent styling in editors. Issue #2.
```
bbangert: Merge fail? It should be 218.

autopush/logging.py
```diff
@@ -69,14 +70,12 @@ def start(self):
...
-    # uncomment to get concise human readable logging messages.
-    """
-    if message['error']:
-        sys.stdout.write("ERROR: %s\n" % message['message'])
-    else:
-        sys.stdout.write("       %s\n" % message['message'])
-    return
-    # """
+    if HUMAN:
+        if message['error']:
```
bbangert: Is this key guaranteed to exist? Maybe a .get() would be safer.

jrconlin: It's pretty much guaranteed. Plus we're doing the same thing later in the code.
https://github.com/mozilla-services/autopush/blob/master/autopush/logging.py#L84

autopush/main.py
```diff
@@ -255,11 +261,23 @@ def make_settings(args, **kwargs):
...
-        if args.gcm_apikey is not None:
+        if args.gcm_enabled:
+            # Create a common gcmclient
+            list = json.loads(args.senderid_list)
+            senderIDs = SenderIDs(dict(
+                s3_bucket=args.s3_bucket,
+                senderid_expry=args.senderid_expry,
+                use_s3=args.s3_bucket.lower() != "none",
+                senderid_list=list))
+            senderID = senderIDs.getID()
```
bbangert: nit: comment that this is run here just to ensure it works, as the value is tossed.

autopush/logging.py
bbangert: Ah, right.

autopush/senderids.py
```diff
@@ -78,17 +108,33 @@ def senderIDs(self):
...
-        return random.choice(self._senderIDs)
+        if not len(self._senderIDs):
+            return None
+        choice = random.choice(self._senderIDs.keys())
+        record = self._senderIDs.get(choice)
+        record["senderID"] = choice
+        return record
 
     def __init__(self, args):
```
bbangert: nit: __init__ method is always the first method in a class (unless its wacky enough to have a __new__)

autopush/senderids.py
```diff
@@ -78,17 +108,33 @@ def senderIDs(self):
...
         return self._senderIDs
 
+    def get(self, id):
+        """Return the associated record for a given SenderID"""
+        record = self._senderIDs.get(id)
+        if record:
+            record["senderID"] = id
+        return record
+
     def getID(self):
```
bbangert: nit: this isn't Python style, though unfortunately some libs we use have similar failings (ie, twisted). get_id would be pythonic. We might need to file a separate ticket to go through and thoroughly ensure we have a consistent style.

bbangert:
Besides for the few nits, everything else looks good now!

bbangert:
r+, LGTM

PR #231, opened 2015-11-10
Convert SenderID to use TimerService

### LINKED ISSUE #230: Convert SenderID refresh to not use time comparisons
The current SenderID refresh uses a time check which can cause unexpected race conditions. 

It's better (and more efficient) to place the SenderID refresh on it's own timed loop.

jrconlin:
Convert SenderID refresh to TimerService. 

Resolves #230. 

@bbangert r?

### DIFF
--- a/autopush/senderids.py
+++ b/autopush/senderids.py
@@ -31,7 +31,7 @@ from boto.exception import S3ResponseError
 from boto.s3.connection import S3Connection
 from boto.s3.key import Key
 from twisted.python import log
-from twisted.internet import reactor
+from twisted.application.internet import TimerService
 
 # re-read from source every 15 minutes or so.
 SENDERID_EXPRY = 15*60
@@ -43,14 +43,13 @@ class SenderIDs(object):
     _expry = SENDERID_EXPRY
     _senderIDs = {}
     _use_s3 = True
-    _cancelled = False
     KEYNAME = "senderids"
-    _refresher = None
+    service = None
 
     def __init__(self, args):
         """Optionally load or fetch the set of SenderIDs from S3"""
         self.conn = S3Connection()
-        self.ID = args.get("s3_bucket", DEFAULT_BUCKET)
+        self.ID = args.get("s3_bucket", DEFAULT_BUCKET).lower()
         self._expry = args.get("senderid_expry", SENDERID_EXPRY)
         self._use_s3 = args.get("use_s3", True)
         senderIDs = args.get("senderid_list", {})
@@ -61,12 +60,8 @@ class SenderIDs(object):
                 self.update(senderIDs)
         # set up auto-refresh for senderIDs
         if self._use_s3:
-            self._refresh()
-
-    def __del__(self):
-        if self._refresher:
-            import pdb;pdb.set_trace()
-            self._refresher.cancel()
+            self.service = TimerService(self._expry, self._refresh)
+            self.service.startService()
 
     def _write(self, bucket, senderIDs):
         """Write a list of SenderIDs to S3"""
@@ -82,8 +77,6 @@ class SenderIDs(object):
 
     def _refresh(self):
         """Refresh the senderIDs from the S3 bucket"""
-        if self._cancelled:
-            return
         if not self._use_s3:
             return
         try:
@@ -99,8 +92,6 @@ class SenderIDs(object):
                 self._senderIDs = candidates
         except S3ResponseError:
             self._create(self._senderIDs)
-        finally:
-            self.refresher = reactor.callLater(self._expry, self._refresh)
 
     def update(self, senderIDs):
         """Update the S3 bucket containing the SenderIDs"""
@@ -140,6 +131,6 @@ class SenderIDs(object):
         record["senderID"] = choice
         return record
 
-    def cancel(self):
-        if self._refresher:
-            self._refresher.cancel()
+    def stop(self):
+        if self.service:
+            self.service.stopService()
--- a/autopush/senderids.py
+++ b/autopush/senderids.py
@@ -32,6 +32,7 @@ from boto.s3.connection import S3Connection
 from boto.s3.key import Key
 from twisted.python import log
 from twisted.application.internet import TimerService
+from twisted.internet.defer import inlineCallbacks
 
 # re-read from source every 15 minutes or so.
 SENDERID_EXPRY = 15*60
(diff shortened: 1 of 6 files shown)


autopush/senderids.py
```diff
@@ -58,7 +58,10 @@ def __init__(self, args):
                 log.err("senderid_list is not a dict. Ignoring")
             else:
                 self.update(senderIDs)
-        self._refresh()
+        # set up auto-refresh for senderIDs
+        if self._use_s3:
+            self.service = TimerService(self._expry, self._refresh)
```
bbangert: Is _refresh a blocking operation?

jrconlin: It's hard for me to determine from (the docs)[https://twistedmatrix.com/documents/current/api/twisted.application.internet.TimerService.html], the source indicates that (it may be using the global reactor)[http://twistedmatrix.com/trac/browser/tags/releases/twisted-15.4.0/twisted/application/internet.py#L223], so yes, this should be added as a blocking call. My apologies, I thought it was using it's own thread loop.

autopush/senderids.py
```diff
@@ -58,30 +59,32 @@ def __init__(self, args):
...
+            self.service.startService()
 
+    @inlineCallbacks
     def _write(self, bucket, senderIDs):
         """Write a list of SenderIDs to S3"""
         key = Key(bucket)
         key.key = self.KEYNAME
-        key.set_contents_from_string(json.dumps(senderIDs))
+        yield key.set_contents_from_string(json.dumps(senderIDs))
```
bbangert: This method returns a deferred? inlineCallbacks using yield should be yielding deferreds.

autopush/senderids.py
```diff
@@ -58,42 +58,44 @@ def __init__(self, args):
...
-                if type(candidates) is not dict:
-                    log.err("Wrong data type stored for senderIDs. "
-                            "Should be dict. Ignoring.")
-                    return
-                self._senderIDs = candidates
-                self._updated = time.time()
-        except S3ResponseError:
-            self._create(self._senderIDs)
+        d = Deferred()
+        d.addCallback(self._update_senderIDs)
```
bbangert: The function still blocks in the loop, you want to use deferToThread which returns a deferred, but kicks the function to run in a thread. You also want _create to run in the thread with it.

autopush/senderids.py
```diff
@@ -106,18 +108,15 @@ def update(self, senderIDs):
...
-        try:
-            bucket = self.conn.get_bucket(self.ID)
-            self._write(bucket, senderIDs)
-        except S3ResponseError:
-            self._create(senderIDs)
+        self._senderIDs = senderIDs
+        d = Deferred()
+        d.addCallback(self._write)
+        d.addErrback(self._create)
+        d.callback(senderIDs)
```
bbangert: This isn't any different than just running the functions in sequential order, since calling it with callback starts the chain running immediately.

autopush/senderids.py
```diff
@@ -40,63 +41,73 @@
...
+        bucket = self.conn.get_bucket(self.ID)
+        key = Key(bucket)
+        key.key = self.KEYNAME
+        candidates = json.loads(key.get_contents_as_string())
+        if candidates:
+            if type(candidates) is not dict:
+                log.err("Wrong data type stored for senderIDs. "
+                        "Should be dict. Ignoring.")
+                return
+            self._senderIDs = candidates
```
kitcambridge: This isn't thread-safe. I think you can just return `candidates` here. Then, in `_refresh`, you can write `d.addCallback(self._set_senderIDs)`.

autopush/senderids.py
```diff
@@ -40,63 +41,73 @@
...
         key.key = self.KEYNAME
         key.set_contents_from_string(json.dumps(senderIDs))
         self._senderIDs = senderIDs
-        self._updated = time.time()
 
-    def _create(self, senderIDs):
+    def _err(self, state):
+        if (isinstance(state.value, S3ResponseError) and
+                state.value.reason == 'Not Found'):
+            self._create()
```
kitcambridge: Should this be `deferToThread(self._create)`, or is this guaranteed to only happen at startup?

autopush/senderids.py
```diff
@@ -40,63 +41,73 @@
...
             else:
+                # We're initializing, so it's ok to block.
                 self.update(senderIDs)
-        self._refresh()
 
-    def _write(self, bucket, senderIDs):
+    def start(self):
+        if self._use_s3:
+            log.msg("Starting SenderID service...")
+            self.service = TimerService(self._expry, self._refresh)
```
kitcambridge: [`LoopingCall`](http://twistedmatrix.com/documents/current/api/twisted.internet.task.LoopingCall.html) might be a better fit. Looks like there's lots of Twisted Application machinery in `TimerService`.

autopush/senderids.py
```diff
@@ -40,63 +41,73 @@
...
         self._expry = args.get("senderid_expry", SENDERID_EXPRY)
         self._use_s3 = args.get("use_s3", True)
         senderIDs = args.get("senderid_list", {})
+        self.service = None
         if senderIDs:
             if type(senderIDs) is not dict:
                 log.err("senderid_list is not a dict. Ignoring")
             else:
+                # We're initializing, so it's ok to block.
                 self.update(senderIDs)
```
kitcambridge: What do you think of moving this into a factory function, like `fetch_senderIDs`? I'm not a fan of constructors with (blocking) side effects, but it's totally your call.

kitcambridge:
@jrconlin Sorry for the drive-by review. This is shaping up nicely!

autopush/senderids.py
jrconlin: All write operations are only at startup.

autopush/senderids.py
jrconlin: Ah! start() sets the period. cool, I'll see about swapping it in.

autopush/senderids.py
jrconlin: Normally, I'm not either, however in this case, it's kinda important that we have data that needs to be written and successfully before processing starts. Ben noted early on that if there's a failure, it needs to happen very early on so that it's caught by ops. Otherwise, there's the risk that it doesn't get detected.

autopush/senderids.py
bbangert: That's actually ok, the GIL means that this line will run atomically.

bbangert:
r+, this looks good!

PR #234, opened 2015-11-11
Allow for key rotation

### LINKED ISSUE #204: Allow for secondary endpoint encryption key for key rotation
We currently have a single key for encrypting endpoints. We need to make sure we can rotate to a new key reasonably easily.

*Expected Goals from deployment*

1.  Key rotation should not require server restarts, but may require a keys to be cycled in on a clock (e.g. keys are refreshed every 15 minutes from a central repository like S3) S3 has a very high availability rate, particularly for read processes and can be tightly controlled for access.
2.  A limited number of prior keys would be held. Removal of old keys would be a manual process since it's difficult to determine how to do so programatically. (e.g. low usage keys for things like Find My Device).
3.  To reduce CPU overhead, tokens can self-identify the appropriate key to use. Since there is an existing key which does not preform this self-identification, the old key could be used as a last-ditch token decryption attempt.

jrconlin:
Use MultiFernet to provide key rotation facility.

Resolves #204 

r? @bbangert


jrconlin:
Went with the simpler approach rather than an indexed key method. I may revisit that if we have dozens of keys and little idea which are still in the wild.

autopush/settings.py
```diff
@@ -168,7 +170,12 @@ def update(self, **kwargs):
         ``self.fernet`` attribute will be initialized"""
         for key, val in kwargs.items():
             if key == "crypto_key":
-                self.fernet = Fernet(val)
+                fkeys = []
+                if not isinstance(val, type([])):
```
bbangert: I believe you wanted:
```python
isinstance(val, list)
```

autopush/settings.py
```diff
@@ -1,7 +1,7 @@
 """Autopush Settings Object and Setup"""
 import socket
 
-from cryptography.fernet import Fernet
+from cryptography.fernet import Fernet, MultiFernet
```
bbangert: Ok, that's rad that they already got an object for key rotation.

bbangert:
Missing changelog.

autopush/settings.py
jrconlin: Less than fun fact: pdb chokes on isinstance(val, list) saying that local variable "list" is not defined, thus the "type([])" nonsense. I can add a comment if you like.

autopush/settings.py
jrconlin: Well, it's not as rad as I'd prefer. All it does is create a list of Fernets that encrypt() or decrypt() runs through. I might monkey patch it in the future to log which key successfully decrypted so that we know when we can retire an older key. As it stands, we have to do math repeatedly on older keys.

I was creating a system that used a key index prefixed to the encrypted block to designate which key was associated. That way we can track key usage. I had a few sticking points and got clever than I should have, so I reverted to just the simple approach.

bbangert:
r+, looks good

PR #237, opened 2015-11-13
Move to looping call & r's

jrconlin:
Follow up to prior patch. 

Move to Looping Call (which only starts the clock when the last call finishes)


bbangert:
r+, nice and clean.

PR #239, opened 2015-11-16
Convert from Hawk to Bearer tokens

### LINKED ISSUE #238: Convert endpoint Auth from Hawk to bearer token
Per the agreed API, switch auth from Hawk to simple bearer token.

jrconlin:
Resolves #238 

Adds argument: 
--auth_key=[key list]

NOTE: this will fail back to using --crypto_key while we wait for config updates. This behavior will be removed in the future.

@bbangert r?


jrconlin:
Closing PR to retarget merge

PR #240, opened 2015-11-16
Feature/bearer

### LINKED ISSUE #238: Convert endpoint Auth from Hawk to bearer token

jrconlin:
Convert from Hawk auth to Bearer token.

**NOTE: this fix introduces a new config option:**
`--auth_key=[]`

Like crypto_key, this key can be generated using the same `autokey` tool, and allows for fallback from [newest, oldest]. 
This key is used to derive the bearer token used. Currently it will default to using the crypto_key. This function is only a stop-gap until the configs can be updated. 

resolves #238 

r? @bbangert


CHANGELOG.rst
```diff
@@ -50,6 +76,9 @@ Backwards Incompatibilities
   values in S3. The data may still be written as a JSON string such as:
   ' "`_senderID_`": {"auth": "`_api_key`"}}'
   activate the GCM bridge by specifying --gcm_enabled.
+* Please include the new --auth_key which is the base token set for
+  generating bearer tokens. This uses the same format as the --crypto_key, but
+  should be a different value to prevent possible key detection.
```
bbangert: Why is the 1.8.0 changelog being updated after its release?

CHANGELOG.rst
```diff
@@ -2,6 +2,40 @@
...
+--------
+
+Bug Fixes
+---------
+
+WebPush
+-------
+
+1.8.1 (2015-11-16)
+===============
```
bbangert: rst requires the header indicator here to match the line length of the header above it

CHANGELOG.rst
```diff
@@ -2,6 +2,40 @@
...
+  Issue #238.
+
+Bug Fixes
+---------
+
+WebPush
+-------
+
+Configuration Changes
+---------------------
```
bbangert: nit: line breaks after header lines is the standard style throughout

bbangert:
r+, looks good

PR #241, opened 2015-11-16
autokey: run main on exec

oremj:
@kitcambridge r?


jrconlin:
why not just use bin/autokey ?

oremj:
Ah, I didn't notice it was in setup.py. I'm fine with not doing this.

kitcambridge:
Oops, sorry I missed this.

PR #243, opened 2015-11-17
Merge 1.8.1 to master

jrconlin:
* Fix issue with auth_key args.
* Removed bogus default cryptokey

@bbangert r?

### DIFF
--- a/configs/autopush_endpoint.ini
+++ b/configs/autopush_endpoint.ini
@@ -24,3 +24,7 @@ port = 8082
 ;
 ; GCM requires a API Auth key. It is STRONGLY recommended you use one specific to this app.
 ; gcm_apikey = <API_KEY>
+;
+; AuthKey are the keys to use for Bearer Auth tokens. It uses the same
+; autokey generator as the crypto_key argument, and sorted [newest, oldest]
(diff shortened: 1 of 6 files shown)


configs/autopush_endpoint.ini
```diff
@@ -24,3 +24,7 @@ port = 8082
 ;
 ; GCM requires a API Auth key. It is STRONGLY recommended you use one specific to this app.
 ; gcm_apikey = <API_KEY>
+;
+; AuthKey are the keys to use for Bearer Auth tokens. It uses the same
+; autokey generator as the crypto_key argument, and sorted [newest, oldest]
+; auth_key = [HJVPy4ZwF4Yz_JdvXTL8hRcwIhv742vC60Tg5Ycrvw8=]
```
jrconlin: I wonder if this might break puppet? In essence, we're specifying invalid JSON here, since a list is created as [key, key]

bbangert: But this is an INI?

jrconlin: True, but specifying this as ["key"] breaks things.

jrconlin: Also, will add coverage test tomorrow.

bbangert:
Looks fine. r+

PR #249, opened 2015-11-18
Fix endpoint to not require local certs for TLS

### LINKED ISSUE #248: Allow endpoint TLS without certs
Currently endpoint only allows TLS if certs are specified. This prevents us using ELBs.

jrconlin:
@bbangert r?

Resolves #248


bbangert:
r+

PR #250, opened 2015-11-18
Bug 1225253 - Implement DELETE.

ncalexan:
@jrconlin This needs review and a Part 3 -- tests.  I am not at all familiar with your testing regimen, so can you both run the existing test suite, and add tests exercising this functionality?

It's worth noting that `DELETE uaid/chid` gives a 401 when `uaid` is unknown, since the auth validation fails first.  I think this is wrong -- it should return 404 -- but I'm not going to address it right now.

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -455,6 +455,12 @@ class AutoendpointHandler(cyclone.web.RequestHandler):
         log.msg("UAID not found in AWS.", **self._client_info())
         self._write_response(404, 103)
 
+    def _chid_not_found_err(self, fail):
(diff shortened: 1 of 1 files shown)


jrconlin:
I'm fine reviewing this, but I don't think that it's really needed. As per our previous discussion, I'm reworking the HTTP API to use the new

/v1/gcm/{senderID}/registration[/{uaid}/subscription[/{chid}]] 

schema.  I'm happy to incorporate the changes (and modify the tests) as part of that.

jrconlin:
Fix rolled into PR #256 
closing

PR #251, opened 2015-11-18
Bug 1225938 - Make autopush tunnel whitelisted encryption headers to …

ncalexan:
…GCM router. r=jrconlin

@jrconlin Same deal -- over to you for sanity testing and unit tests.  Presumably I should be updating documentation somewhere, as well.


jrconlin:
This patch has been rolled up under #256.
closing

PR #252, opened 2015-11-19
Add message table rotation for webpush users

### LINKED ISSUE #191: Message table rotation
To avoid heavy table scanning to clear out expired messages, message table rotation should be used.

The way it would work is like this, the table name will be based on the current year/month, ie:
messages_2015_02

- [ ] Notifications will be stored in the appropriate month table for that month
- [ ] On websocket register, the old connected_at will be evaluated to see if a scan of last months notifications is needed.
- [ ] During ack, notification will be deleted from appropriate table
- [ ] All new messages to the endpoint will have TTL adjusted to a max of how many months we retain for.
- [ ] Lambda function that lowers read/write unit's on old tables
- [ ] Lambda function that create new table, copies over existing tables read/write units

bbangert:
This PR add's message table rotation for webpush users, and retains a current_month attribute in the router table for the user indicating which message table to use for notifications.

Table rotation occurs when a client connects, after it has ack'd all remaining notifications in the table it was last on. Then the channels entry is copied over, and the current_month updated.

Webpush users with an account prior to table rotation will have their UAID reset so they can re-register channels properly with the table rotation scheme.

Closes #191.

@jrconlin r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -312,9 +312,9 @@ class Message(object):
         try:
             result = self.table.get_item(consistent=True, uaid=uaid,
                                          chidmessageid=" ")
-            return result["chids"] or set([])
+            return (True, result["chids"] or set([]))
         except ItemNotFound:
-            return set([])
+            return False, set([])
 
     @track_provisioned
     def save_channels(self, uaid, channels):
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -22,11 +22,11 @@ from boto.dynamodb2.types import NUMBER
 log = logging.getLogger(__file__)
 
 
-def next_month(delta=1):
+def get_month(delta=0):
     """Basic helper function to get a datetime.date object iterations months
     ahead/behind of now."""
     new = last = datetime.date.today()
-    # Move  until we hit a new month, this avoids having to manually
+    # Move until we hit a new month, this avoids having to manually
     # check year changes as we push forward or backward since the Python
     # timedelta math handles it for us
     for _ in range(abs(delta)):
@@ -42,7 +42,7 @@ def next_month(delta=1):
 def make_rotating_tablename(prefix, delta=0):
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
-    date = next_month(delta=delta)
+    date = get_month(delta=delta)
     return "{}_{}_{}".format(prefix, date.year, date.month)
 
 
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -1,6 +1,7 @@
 """Database Interaction"""
 from __future__ import absolute_import
 
+import datetime
 import logging
 import time
 import uuid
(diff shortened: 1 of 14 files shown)


autopush/db.py
```diff
@@ -21,6 +22,51 @@
 log = logging.getLogger(__file__)
 
 
+def next_month(delta=1):
+    """Basic helper function to get a datetime.date object iterations months
+    ahead/behind of now."""
+    new = last = datetime.date.today()
+    # Move  until we hit a new month, this avoids having to manually
```
jrconlin: nit: double space

autopush/db.py
```diff
@@ -21,6 +22,51 @@
 log = logging.getLogger(__file__)
 
 
+def next_month(delta=1):
```
jrconlin: It looks like you never call next_month(); there's always a parameter. Calling functions default delta to 0. Should we be consistent here?

Also minor nit: might want to rename this get_month since it tends to either return the current, or previous month.

autopush/db.py
```diff
@@ -21,6 +22,51 @@
...
+
+
+def make_rotating_tablename(prefix, delta=0):
+    """Creates a tablename for table rotation based on a prefix with a given
+    month delta."""
+    date = next_month(delta=delta)
+    return "{}_{}_{}".format(prefix, date.year, date.month)
+
+
+def create_rotating_message_table(prefix="message", read_throughput=5,
```
jrconlin: What's calling this function?

autopush/settings.py
```diff
@@ -136,15 +144,21 @@ def __init__(self,
...
-                                               storage_read_throughput,
-                                               storage_write_throughput)
-        self.message_table = get_message_table(message_tablename,
-                                               message_read_throughput,
-                                               message_write_throughput)
+        self.storage_table = get_storage_table(
+            storage_tablename,
+            storage_read_throughput,
+            storage_write_throughput)
+        self.message_table = get_rotating_message_table(
```
jrconlin: Nit: Might want to note that the message table needs to be created before running, otherwise badness. I'm presuming that there's a creator eventually.

autopush/tests/test_db.py
```diff
@@ -182,19 +188,34 @@ def test_all_channels(self):
...
 
+    def test_save_channels(self):
+        chid = str(uuid.uuid4())
+        chid2 = str(uuid.uuid4())
+        m = get_message_table()
+        message = Message(m, SinkMetrics())
+        message.register_channel(self.uaid, chid)
+        message.register_channel(self.uaid, chid2)
+
+        exists, chans = message.all_channels(self.uaid)
```
jrconlin: is `exists` checked? It's used as the primary flag in webpush.py.

autopush/tests/test_integration.py
```diff
@@ -991,6 +997,155 @@ def test_message_update_unackd_delivered_message(self):
...
+
+        # Verify the move
+        c = yield deferToThread(self._settings.router.get_uaid, client.uaid)
+        eq_(c["current_month"], last_month)
+
+        # Move the clients channels back one month
+        exists, chans = yield deferToThread(
+            self._settings.message.all_channels, client.uaid
+        )
+        eq_(exists, True)
```
jrconlin: Ah, we're checking the `exists` value here. Probably want to replace `exists` with `_` in `test_save_channels` and `test_all_channels_no_uaid`

autopush/tests/test_main.py
```diff
@@ -45,6 +45,66 @@ def test_resolve_host_no_interface(self, mock_socket):
         eq_(ip, "google.com")
 
 
+class SettingsAsyncTestCase(trialtest.TestCase):
+    def test_update_rotating_tables(self):
+        from autopush.db import create_rotating_message_table, get_month
+        # Create the rotating table so the test passes
+        create_rotating_message_table()
+        settings = AutopushSettings(
+            hostname="google.com", resolve_hostname=True)
```
jrconlin: while popular, probably want to use "example.com". There's a lot of optimization to ensure example.com resolves.

autopush/websocket.py
```diff
@@ -792,6 +865,43 @@ def finish_webpush_notifications(self, notifs):
...
 
+    def _rotate_message_table(self):
+        """Function to fire off a message table copy of channels + update the
+        router current_month entry"""
+        self.transport.pauseProducing()
+        d = self.deferToThread(self.ps.message.all_channels, self.ps.uaid)
+        d.addCallback(self._register_rotated_channels)
+        d.addErrback(self.trap_cancel)
+        d.addErrback(self.err_overload, "notif")
+        d.addErrback(self.log_err)
```
jrconlin: do we want to return d here?

docs/architecture.rst
```diff
@@ -8,13 +8,93 @@ Endpoint nodes handle all notification PUT requests, looking up in DynamoDB to
...
+only moved to the new table on connect, and only after reading/acking all the
+notifications for the prior month.
+
+**Rules for Endpoints**
+
+1. Check the router table to see the current_month the client is on.
+2. Read the chan list entry from the appropriate month message table to see if
+   its a valid channel.
+
+   If its valid, move to step 3.
```
jrconlin: if not, it's an error. Is there any recovery mechanism or how is it signaled? (Thinking in terms of an Ops person reading this wondering what to look for.)

autopush/websocket.py
bbangert: The caller doesn't need a deferred back, so its not needed.

docs/architecture.rst
bbangert: If the channel isn't valid, then the behavior is identical to current behavior, ie, the channel is unregistered and the message is 404'd.

jrconlin:
r+

Issue #229, opened 2015-11-06
Add error handler for cyclone bits in autopush


bbangert:
I found a recent bug in the cyclone HTTP handler portion of websocket.py lately cause I was cruising through the kibana logs. We don't capture errors in the cyclone handler here, like we do in endpoint.py, since we don't have our own ``write_error`` method in there.

We should copy/adapt the write_error method from endpoint.py into the cyclone handlers in websocket.py.

jrconlin:
resolved by #257

PR #257, opened 2015-11-24
Add write_error override to other handlers

jrconlin:
issue #229

Since it was popping up in multiple files & classes, moved
write_error to utils and added pass-thoughs to classes.

@bbangert r?

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -53,18 +53,19 @@ def generate_hash(key, payload):
     h = hmac.new(key=key, msg=payload, digestmod=hashlib.sha256)
     return h.hexdigest()
 
+class Utils (object):
 
-def write_error(this, code, **kwargs):
-    """Write the error (otherwise unhandled exception when dealing with
-    unknown method specifications.)
+    def write_error(self, code, **kwargs):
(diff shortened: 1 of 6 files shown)


bbangert:
I can't say I'm a fan of multiple inheritance, but a mix-in class is really a much more Pythonic alternative and removes the very odd 'this' javascript smell to the error method and removes adding proxy methods to it inside each class.

autopush/utils.py
```diff
@@ -50,3 +52,21 @@ def generate_hash(key, payload):
     """
     h = hmac.new(key=key, msg=payload, digestmod=hashlib.sha256)
     return h.hexdigest()
+
+
+class Utils (object):
```
bbangert: nit: vague name, ErrorLogger maybe?

autopush/websocket.py
```diff
@@ -57,7 +57,7 @@
 
 from autopush import __version__
 from autopush.protocol import IgnoreBody
-from autopush.utils import validate_uaid
+from autopush.utils import (validate_uaid, Utils)
```
bbangert: nit: no parens unless dropping the line down

bbangert:
r+ pending nits

PR #262, opened 2015-12-01
bug: fix db test to do month addition properly

### LINKED ISSUE #261: Fix db test bug
It's december, so the poor month math in a db check now fails. It should be handled properly rather than just adding 1 to the current month.

bbangert:
Closes #261.

@jrconlin r?

### DIFF
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -56,7 +56,12 @@ class DbCheckTestCase(unittest.TestCase):
         from autopush.db import get_month
         month0 = get_month(0)
         month1 = get_month(1)
-        eq_(month0.month+1, month1.month)
+        this_month = month0.month
+        if this_month == 12:
+            next_month = 1
+        else:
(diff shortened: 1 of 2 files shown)


autopush/tests/test_db.py
```diff
@@ -56,7 +56,9 @@ def test_get_month(self):
         from autopush.db import get_month
         month0 = get_month(0)
         month1 = get_month(1)
-        eq_(month0.month+1, month1.month)
+        this_month = month0.month
+        next_month = 1 if this_month == 12 else this_month + 1
```
kitcambridge: How about `(this_month % 12) + 1`?

kitcambridge:
r+

autopush/tests/test_db.py
bbangert: Mainly because modulus operations aren't as obvious when skimming.

kitcambridge: :+1:

PR #267, opened 2015-12-03
Fix running documentation to indicate how to use the dynamodb local server

### LINKED ISSUE #265: Fix running docs
Docs indicate it should be run with moto_server, which is wrong. We require Amazon's local DynamoDB JAR, which implements the entire spec properly.

bbangert:
Closes #265.

@kitcambridge r?


jrconlin:
r+

PR #269, opened 2015-12-03
bug: Fix startup issue with rotating message table not being created.

### LINKED ISSUE #266: Autopush does call create_rotating_message_table on init
With the rotating table fix, autopush does not properly instantiate the rotating message tables, which can cause errors on route creation.

bbangert:
Closes #266.

@kitcambridge r?

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -13,6 +13,7 @@ from twisted.python import log
 from twisted.web.client import Agent, HTTPConnectionPool
 
 from autopush.db import (
+    create_rotating_message_table,
     get_router_table,
     get_storage_table,
     get_rotating_message_table,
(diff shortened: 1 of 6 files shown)


autopush/settings.py
```diff
@@ -204,6 +205,13 @@ def create_initial_message_tables(self):
...
         this_month = get_rotating_message_table(self._message_prefix)
+        # Verify the table exists
+        try:
+            this_month.describe()
+        except Exception:
+            # Create the message table
+            create_rotating_message_table(prefix=self._message_prefix)
+            this_month = get_rotating_message_table(self._message_prefix)
         self.message_tables = {
             last_month.table_name: Message(last_month, self.metrics),
```
kitcambridge: What if `last_month` doesn't exist, either?

bbangert: That's fine because it means this isn't a real production deploy. The current month is needed for a new start, ongoing deployments will run a set of scripts that hasn't landed yet to ensure tables are around.

kitcambridge:
r+

PR #272, opened 2015-12-04
bug: fix issue with gcm router not being passed appropriate params.

### LINKED ISSUE #271: Create Bridge test that verifies GCM interaction
The GCM test only verifies GCM code, the endpoint registration test fails to verify that the GCM code gets the right args from the endpoint. It needs a test that verifies the correct data is passed to the GCM bridge code.

bbangert:
Closes #271.

@kitcambridge r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Features
 Bug Fixes
 ---------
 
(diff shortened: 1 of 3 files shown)


kitcambridge:
r+

PR #273, opened 2015-12-04
Move blocking channel registration to a thread pool

kitcambridge:
I noticed this while reviewing #272. We were calling `make_endpoint` in the thread pool, but not `register_channel`. No changelog entry, since the HTTP API is new in this release.

@bbangert r?


bbangert:
Nice catch! r+

PR #276, opened 2015-12-11
bug: Add self param for metric sink. Resolves #274.

### LINKED ISSUE #274: autopush/metrics.py - missing self parameter
Hi!

I'm not using autopush but as I was looking at the code I noticed that there is missing `self` paremeter missing in `increment` method of `SinkMetrics` class:

    class SinkMetrics(IMetrics):
        """Exists to ignore metrics when metrics are not active"""
        def increment(name, count=1, **kwargs):
            pass

jrconlin:
@bbangert r?


bbangert:
r+

PR #277, opened 2015-12-14
feature: Pass Encryption-Key xor Crypto-Key. Fixes #258.

### LINKED ISSUE #258: Forward the `Crypto-Key` header
martinthomson/http-encryption#4 replaces the `Encryption-Key` header with `Crypto-Key`. We should:

1. Forward the header to the client as `crypto_key`.
2. Reject requests that include both crypto headers. It's still OK for messages to use `Encryption-Key`; the client is backward-compatible. But requests that include `Crypto-Key` *and* `Encryption-Key` are bogus.

jrconlin:
@bbangert || @kitcambridge  r?

Will need to notify @ncalexan before change is pushed to

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -72,6 +72,7 @@ Unless otherwise specified, all calls return the following error codes:
 
    - errno 101 - Missing neccessary crypto keys
    - errno 108 - Router type is invalid
+   - errno 110 - Invalid crypto keys specified
 
 -  401 - Bad Authorization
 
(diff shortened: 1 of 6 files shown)


kitcambridge:
LGTM. Thanks, JR!

autopush/endpoint.py
```diff
@@ -774,11 +776,17 @@ def _put_message(self, kind, uaid, chid):
                 encryption=self.request.headers["encryption"],
             )
             data = urlsafe_b64encode(self.request.body)
-            # AWS cannot store empty strings, so we only add the encryption-key
+            # AWS cannot store empty strings, so we only add the key
             # if its present to avoid empty strings.
-            if "encryption-key" in self.request.headers:
-                headers["encryption_key"] = \
-                    self.request.headers["encryption-key"]
+            if ("encryption-key" in self.request.headers and
```
kitcambridge: My impression is that the `Topic` header (webpush-wg/webpush-protocol#9) replaces the `PUT` mechanism for collapsing messages. I'll file a separate issue for removing `_put_message` and friends.

PR #281, opened 2015-12-17
bug: Fix delete error reporting. resolves bugzilla:1215253

jrconlin:
Checked over the delete code to ensure that proper states and errors are
being returned. Added tests to better guard against changes.

* Removed "all" flag for delete_message
* check state where possible and without increasing read costs.

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -298,13 +298,22 @@ class Message(object):
         db_key = self.encode({"uaid": uaid, "chidmessageid": " "})
         expr = "DELETE chids :channel_id"
         expr_values = self.encode({":channel_id": set([channel_id])})
-        conn.update_item(
+
+        result = conn.update_item(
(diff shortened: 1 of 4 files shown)


autopush/db.py
```diff
@@ -408,16 +417,14 @@ def delete_message(self, uaid, channel_id, message_id, updateid=None):
                 channel_id, message_id))
         return True
 
-    def delete_messages(self, uaid, chidmessageids, all=False):
+    def delete_messages(self, uaid, chidmessageids):
         with self.table.batch_write() as batch:
             for chidmessageid in chidmessageids:
```
bbangert: filter(None, chidmessageids) will remove the need for the if check on 423.

bbangert:
r+, minor nit though it's not needed so I'm merging as is

PR #282, opened 2015-12-19
refactor: Remove message updating

### LINKED ISSUE #279: Remove the PUT method from `MessageHandler`
This was originally added to support message collapsing, per webpush-wg/webpush-protocol#12. That proposal has been replaced with the `Topic` header (webpush-wg/webpush-protocol#62). We can add `Topic` in a separate issue, but let's remove `put`, `_put_message`, and `_put_completed`.

kitcambridge:
We'll probably want some of these tests when we implement `Topic`, but I just pruned them for now. Closes #279.

@bbangert r?


bbangert:
r+

PR #287, opened 2015-12-30
Config fixes

### LINKED ISSUE #246: Try to catch configArgParse weirdness
Spot potential odd returns from ConfigArgParse
e.g.
"Value = " sets value to "=", this is most often an error.

Need to test other potential config errors, e.g.

foo:
flag=False
etc.

jrconlin:
* deprecated unused args.
* tests no longer pull in local configs if present.
* added LOTS more commentary to config files (useful for  folks doing setups) and corrected default values.
* more flake8 fixes that I'm not sure why weren't picked up earlier.
* removed old cryptokey value. Don't be lazy & insecure, do what the docs are telling you.

Fixes #246 

@bbangert r?


CHANGELOG.rst
```diff
@@ -31,10 +31,26 @@ Bug Fixes
...
+  --log_level
+  --external_router (replaced by --apns_enabled)
+  --max_message_size
+* It is recommended that the following config options be moved to
+  .autopush_shared.ini
+  --gcm_enabled
+  --senderid_list
+  --senderid_expry
+
+
```
bbangert: Maybe a deprecation section? That way its more obvious they're going to be going away but not quite incompatible yet.

autopush/main.py
```diff
@@ -97,8 +97,6 @@ def add_shared_args(parser):
     parser.add_argument('--router_write_throughput',
                         help="DynamoDB router write throughput",
                         type=int, default=5, env_var="ROUTER_WRITE_THROUGHPUT")
-    parser.add_argument('--log_level', type=int, default=40,
-                        env_var="LOG_LEVEL")
```
bbangert: No way to change the log level at all then?

jrconlin: Honestly, these were deprecated because they weren't used by anything, so no. Nor has there been for quite some time.

CHANGELOG.rst
jrconlin: Fair enough. Mostly I figured I'd create the option so that oremj doesn't have to roll new config files right away.

bbangert:
r+, minor nits but nothing I'd consider blocking

PR #288, opened 2015-12-31
docs: Broke HTTP endpoint docs out into own doc.

jrconlin:
* Slight fixes to formatting of index.rst
* added a few more terms to the glossary
* Standard error codes now point to standard error codes.

@bbangert r?

(Last PR of 2015?)


docs/http.rst
```diff
@@ -0,0 +1,479 @@
...
+-  401 - Bad Authorization
+
+   - errno 109 - Invalid authentication
+
+-  404 - `{UAID}` not found or invalid
+
+   - errno 102 - Invalid URL endpoint
+   - errno 103 - Expired URL endpoint
+   - errno 105 - Endpoint became unavailable during request
+   - errno 106 - Invalid subscription
```
bbangert: We actually return a 413 too, but thats not documented for some reason.

docs/http.rst
```diff
@@ -0,0 +1,479 @@
...
+
+.. code-block:: json
+
+    {}
+
+Errors:
+^^^^^^^
+
+    :statuscode 404: `message_id` is not found.
+    :statuscode 200: Message has been updated.
```
bbangert: This isn't an error though?

docs/http.rst
```diff
@@ -0,0 +1,479 @@
...
+.. code-block:: json
+
+    {"message-id": {message-id}}
+
+Errors:
+^^^^^^^
+    :statuscode 404: `token` is invalid.
+    :statuscode 202: Message stored for delivery to client at a later
+                     time.
+    :statuscode 200: Message delivered to node client is connected to.
```
bbangert: Is it possible to remove some of these header sections and still maintain readability? They add a lot of padding and it seems like it should be obvious that the status codes are in the reply, and the call portion should be formatted appropriately so that its obvious that is the HTTP call being done.

bbangert:
r+, minor nits that could be addressed in a separate PR

jrconlin:
Good calls on these. I'll update to add the 413 and change Errors to Return
Values.

jonathanpmartins:
:+1:

PR #290, opened 2016-01-11
Convert the outbound encrypted GCM data to base64

### LINKED ISSUE #289: GCM: Ensure outboud encrypted stream is encoded as base64
Since the incoming encrypted message data is a byte-stream, it has been recommended that the data be converted to base64 before sending over GCM. 

Client work has already been

jrconlin:
* Fixes #289
* Throws 413 error if the encoded, outbound message is more than
  max_data (this should prevent unexpected developer confusion if the
  bridge rejects a block that is nearly max_data length.)
* Added note to docs.

@bbangert r?

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -88,8 +88,11 @@ class GCMRouter(object):
             udata = urlsafe_b64encode(notification.data)
             mdata = self.config.get('max_data', 4096)
             if len(udata) > mdata:
-                raise self._error("Converted buffer too long by %d bytes" % (
-                    len(udata) - mdata), 413)
+                raise self._error("This message is intended for a " +
+                                  "constrained device and is limited " +
+                                  "to 3070 bytes. Converted buffer too " +
+                                  "long by %d bytes" % (len(udata) - mdata),
+                                  413, errno=104)
             # TODO: if the data is longer than max_data, raise error
             data['body'] = udata
             data['con'] = con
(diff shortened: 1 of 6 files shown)


autopush/router/gcm.py
```diff
@@ -84,7 +85,13 @@ def _route(self, notification, router_data):
                 data['enckey'] = notification.headers.get('encryption-key')
             if ('crypto-key' in notification.headers):
                 data['cryptokey'] = notification.headers.get('crypto-key')
-            data['body'] = notification.data
+            udata = urlsafe_b64encode(notification.data)
```
kitcambridge: I wonder if we should enforce the encoded length for all routers [in this block of code](https://github.com/mozilla-services/autopush/blob/master/autopush/endpoint.py#L378-L389).

That violates the "Push services MUST NOT return a 413 status code in responses to an entity body that is 4k (4096 bytes) or less in size" provision in the spec...but we'd violate it consistently. :wink:

jrconlin: Honestly, it was something I was debating. We need to do the base64 decoding for GCM because the library exchanges the data as part of a JSON blob. For APNs, it's a binary exchange to begin with, so data conversion may not be required. Likewise, if we roll out some other bridge, there may be it's own considerations. 

That's why I'm limiting to just GCM for now.

autopush/router/gcm.py
```diff
@@ -84,7 +85,13 @@ def _route(self, notification, router_data):
...
+            mdata = self.config.get('max_data', 4096)
+            if len(udata) > mdata:
+                self._error("Converted buffer too long by %d bytes" % (
```
bbangert: I kind of wish I had mentioned earlier that self._error should return the exception, and the raise should go here, just so the code flow is more obvious. As is, it's not immediately apparent that self._error is intentionally going to break the flow here by raising an exception.

bbangert:
r+ pending test checks

PR #291, opened 2016-01-12
Allow UAIDs to be hashed against a key

### LINKED ISSUE #228: SHA256 UAID in database
UAIDs and CHIDs don't need to be stored in cleartext, since user agents and notifications resubmit them. It would be preferrable to store them hashed in database, and only do comparisons with what a user agent submits. This would prevent a database leak from revealing them.

jrconlin:
Fixes #228

* NOTE: Changing this key will invalidate all UAIDs currently stored.
There is no rollover function since reading multiple times from DynamoDB
for each potential transaction is STRONGLY discouraged for fiscal reasons.

@bbangert r?


autopush/db.py
```diff
@@ -39,6 +43,12 @@ def get_month(delta=0):
     return new
 
 
+def hasher(uaid):
+    if key_hash:
```
bbangert: Really necessary to use a global var?

bbangert:
Curious about the coverage drop.

Offhand, I think it'd be better to do the hashing outside of autopush.db. Mainly because a single request to various points can result in multiple db calls, and repeating the hashing over and over may not be ideal.

jrconlin:
The coverage drop is odd. The line is definitely getting called, but the coverage isn't picking it up. Possibly it's being JIT'd? 

I agree that the global is a bit ugly. It'd be cleaner to make the UAID its own object with accessor methods. I figured that would impact significantly larger amount of code, so short cut for now.

autopush/tests/test_main.py
```diff
@@ -180,6 +180,7 @@ class test_arg:
         message_write_throughput = 0
         senderid_list = '{"12345":{"auth":"abcd"}}'
         s3_bucket = "none"
+        key_hash = ""
```
bbangert: This evaluates to False, so the code checking for key_hash will not trigger, which is why the code coverage dropped.

jrconlin:
grr. github sync being difficult.

PR #293, opened 2016-01-12
feat: track last month connected using last_connect router field

### LINKED ISSUE #253: Change router table to use rotating table
Just like the message table in #191, the router table should be rotated as well. With the main difference being that entries from pre-rotating table should be migrated.

bbangert:
Utilizes the existing router last_connect global index to track the first
connect of the month for the client. The last_connect number is a number
joining the year/month/hour with a random int to reduce hot key contention
on the index.
    
Additional integration tests were added to verify the last_connect is updated
appropriately for simplepush and webpush clients.
    
This changes the get_rotating_message_table function to automatically create
the table if needed, reducing boiler-plate in the settings object wrapping
that behavior.
    
Registration endpoint handlers saving router data all update the last_connect
as well (new channels do not update the value).
    
Closes #253

@jrconlin r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 
 import datetime
 import logging
+import random
 import time
 import uuid
 from functools import wraps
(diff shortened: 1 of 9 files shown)


autopush/db.py
```diff
@@ -59,17 +60,31 @@ def create_rotating_message_table(prefix="message", read_throughput=5,
...
+
+    The last_connect index is a value used to determine the last month a user
+    was seen in. To prevent hot-keys on this table during month switchovers the
+    key is determined based on the following scheme:
+
+        (YEAR)(MONTH)(DAY)(HOUR)(0001-0010)
+
+    Note that the random key is only between 1-10 at the moment, if the key is
+    still too hot during production the random range can be increased at the
+    cost of additional queries during GC to locate expired users.
```
jrconlin: I'm a bit confused by this. last_connect is set to the YYMMDDHHrr value, but it's not a hash or range key, so it's an attribute. In has_connected_this_month, we only check for YYMM. Even in the message table, there's another attribute that records just the Month along with the last_connect. 

Why the high definition last_connect if we're throwing away more than half of it? Wouldn't YYMMr+

bbangert: Using a much larger r+ without the DD/HH would work fine, but I thought it'd be useful to retain the DD/HH knowledge since we could utilize that later to determine how long into the new month it takes on average for people to reconnect, etc. No use for it right now, but maybe later....

jrconlin: Ok, good point.

jrconlin:
r+

PR #298, opened 2016-01-20
 Fix GCM data block send. Fixes #297

### LINKED ISSUE #297: GCM rejecting data as "non-json"

jrconlin:
* pass "router_token" for senderid lookup
* Clean up comments to be more clear about what each token is.
* Add ability for clients to request "dry run"
* better exam of gcm data response block.
* fix test SenderID hash to match reality.

@bbangert r?

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -105,7 +105,7 @@ class GCMRouter(object):
             registration_ids=[router_data.get("token")],
             collapse_key=self.collapseKey,
             time_to_live=self.ttl,
-            dry_run=self.dryRun or "dryrun" in router_data,
+            dry_run=self.dryRun or ("dryrun" in router_data),
             data=data,
         )
         creds = router_data.get("creds", {"senderID": "missing id"})
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -43,12 +43,14 @@ class GCMRouter(object):
         msg["senderid"] = self.creds.get("senderID")
         return msg
 
-    def register(self, uaid, router_data):
-        """Validate that a token is in the ``router_data``"""
-        if not router_data.get("token"):
-            raise self._error("connect info missing 'token'", status=401)
+    def register(self, uaid, router_data, router_token=None, *kwargs):
+        """ Validate that the GCM Instance Token is in the ``router_data``"""
+        if "token" not in router_data:
+            raise self._error("connect info missing GCM Instance 'token'",
+                              status=401)
         # Assign a senderid
-        router_data["creds"] = self.creds = self.senderIDs.choose_ID()
+        router_data["creds"] = self.creds = \
+            self.senderIDs.get_ID(router_token)
         return router_data
 
     def route_notification(self, notification, uaid_data):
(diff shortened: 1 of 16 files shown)


autopush/router/gcm.py
```diff
@@ -43,12 +43,14 @@ def amend_msg(self, msg):
         msg["senderid"] = self.creds.get("senderID")
         return msg
 
-    def register(self, uaid, router_data):
-        """Validate that a token is in the ``router_data``"""
-        if not router_data.get("token"):
-            raise self._error("connect info missing 'token'", status=401)
+    def register(self, uaid, router_data, router_token=None, *kwargs):
+        """ Validate that the GCM Instance Token is in the ``router_data``"""
```
bbangert: Nit: leading space seems to have snuck in here.

autopush/router/gcm.py
```diff
@@ -43,12 +43,14 @@ def amend_msg(self, msg):
...
+        if "token" not in router_data:
+            raise self._error("connect info missing GCM Instance 'token'",
+                              status=401)
         # Assign a senderid
-        router_data["creds"] = self.creds = self.senderIDs.choose_ID()
+        router_data["creds"] = self.creds = \
```
bbangert: This didn't fit on one line?

bbangert: ``self.creds``, which is on a single router instance used concurrently from multiple paths keeps changing its state based on the router_token of the last register call?

autopush/router/gcm.py
```diff
@@ -98,12 +99,14 @@ def _route(self, notification, router_data):
             data['con'] = con
             data['enc'] = enc
 
+        # registration_ids are the GCM instance tokens (specified during
+        # registration.
         payload = gcmclient.JSONMessage(
-            registration_ids=[router_data["token"]],
+            registration_ids=[router_data.get("token")],
```
bbangert: If there's no token, this will be a list of [None]. Will that cause any issues?

autopush/router/gcm.py
```diff
@@ -98,12 +99,14 @@ def _route(self, notification, router_data):
...
             collapse_key=self.collapseKey,
             time_to_live=self.ttl,
-            dry_run=self.dryRun,
-            data=udata,
+            dry_run=self.dryRun or "dryrun" in router_data,
```
bbangert: Remembering the order of operation precedence isn't my strong-point so I usually () clauses like this just to be sure. Nick actually found an error in older GCM code due to operator precedence working in an unintentional way.

autopush/senderids.py
```diff
@@ -129,11 +129,11 @@ def senderIDs(self):
         """Return a list of senderIDs"""
         return self._senderIDs
 
-    def get_ID(self, id):
+    def get_ID(self, id=None):
         """Return the associated record for a given SenderID"""
+        if id is None:
```
bbangert: ``if not id:`` ? Or does this need to specifically check that its ``None``?

autopush/router/gcm.py
jrconlin: It will be converted to JSON [null] which will be handled by the GCM call.

autopush/senderids.py
jrconlin: I was originally thinking that we might reserve "" as a default defined in the senderid load table, but that might be overthinking and it's an uncommon use case at best.

bbangert:
r+ pending CI sign-off

PR #302, opened 2016-01-22
bug: fix extra db writes on new uaid connect

### LINKED ISSUE #300: Write out last_connect on new UAID
When a new UAID connects, there's no point in wasting another write to the router table to update last_connect, since its a new uaid we already know we should include that.

bbangert:
This also add's a flag to autopush.db allowing all provisioned method calls
to be logged and introspected for testing.

Fixes #300

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,16 +2,14 @@
 Autopush Changelog
 ==================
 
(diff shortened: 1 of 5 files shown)


jrconlin:
r+

PR #303, opened 2016-01-22
bug: fix the multiple write issue for new webpush users

### LINKED ISSUE #300: Write out last_connect on new UAID

bbangert:
This fixes the multiple write issue for new webpush and simplepush users by
updating the previous value the way moto does (incorrectly in moto's case).

Fixes #300, again.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 Autopush Changelog
 ==================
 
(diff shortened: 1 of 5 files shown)


jrconlin:
r+ pending travis

PR #305, opened 2016-01-23
bug: fix connections with invalid UAID from pre-uaid hashing

### LINKED ISSUE #304: Fix table rotation during sign-on to get the right value
By only getting updated old values before they were updated, we're failing to get the entire record which we actually need for the router, as it looks *now*. It actually doesn't matter what any of the prior values were before the update, since the update only updates 2 values during register to begin with.

bbangert:
Closes #304

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Autopush Changelog
 ==================
 
+1.9.3 (2016-01-23)
+==================
+
+* Fix issue with users connecting with an invalid UAID that didn't exist in
+  the database. Issue #304.
+
 1.9.2 (2016-01-22)
 ==================
 
--- a/autopush/__init__.py
+++ b/autopush/__init__.py
@@ -1,1 +1,1 @@
-__version__ = '1.9.2'  # pragma: nocover
+__version__ = '1.9.3'  # pragma: nocover
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -106,12 +106,12 @@ keyid="http://example.org/bob/keys/123;salt="XZwpw6o37R-6qoZjw6KwAw"\
         self.ws = websocket.create_connection(self.url)
         return self.ws.connected
 
-    def hello(self):
+    def hello(self, uaid=None):
         if self.channels:
             chans = self.channels.keys()
         else:
             chans = []
-        hello_dict = dict(messageType="hello", uaid=self.uaid or "",
+        hello_dict = dict(messageType="hello", uaid=uaid or self.uaid or "",
                           channelIDs=chans)
         if self.use_webpush:
             hello_dict["use_webpush"] = True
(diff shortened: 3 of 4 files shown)


jrconlin:
r+

kitcambridge:
Just to make sure I understand all the success cases, since I find it easier to think in prose...

## Case 1: Client sends UAID = `""`

* Generate a new UAID.
* Call `_register_user(existing_user=False)`.
* Set `user_item["last_connect"] = generate_last_connect()` and `user_item["current_month"] = self.ps.message_month`.
* Call `router.register_user(user_item)`.
* The conditional check passes, since `attribute_not_exists(router_type) and attribute_not_exists(node_id)` is true.
* `router.register_user` returns `(True, {}, user_item)`.
* `_copy_new_data` sets `previous["last_connect"] = user_item["last_connect"]` and `previous["current_month"] = user_item["current_month"]`.
* `_check_collision` sets `existing_webpush_rotator = True`, since we've set `current_month`.
* `_check_collision` calls `_check_other_nodes`.
* `_check_other_nodes` disconnects dupes, clears the hello timeout, and responds to the client.

## Case 2: client sends a UAID we've never seen before

* Call `_register_user(existing_user=True)`.
* Since `existing_user = True`, we don't set `last_connect` or `current_month`.
* Call `router.register_user(user_item)`.
* The conditional check passes (same as case 1).
* `router.register_user` returns `(True, {}, user_item)`.
* `_copy_new_data` does nothing.
* `_check_collision` sets `existing_webpush_rotator = False`, since we don't have a `current_month`.
* At this point, `registered == True and self.ps.use_webpush == True`, but we don't have rotation.
* Reset the UAID and call `_register_user(existing_user=False)`.
* Follow steps for case 1.

## Case 3: client sends a UAID we've seen before, but doesn't have rotation yet

This looks like the same as case 2, except the record returned by `router.register_user` has old values.

* Call `_register_user(existing_user=True)`.
* Since `existing_user = True`, we don't set `last_connect` or `current_month`.
* Call `router.register_user(user_item)`.
* The conditional check passes.
* `router.register_user` returns `(True, { "uaid", "node_id", "connected_at", "router_type" }, user_item)`.
* `_copy_new_data` does nothing.
* `_check_collision` sets `existing_webpush_rotator = False`, since we don't have a `current_month`.
* Reset the UAID and call `_register_user(existing_user=False)`.
* Follow steps for case 1.

## Case 4: client sends a UAID we've seen before, with rotation

* Call `_register_user(existing_user=True)`.
* Since `existing_user = True`, we don't set `last_connect` or `current_month`.
* Call `router.register_user(user_item)`.
* The conditional check passes.
* `router.register_user` returns `(True, { "uaid", "node_id", "connected_at", "router_type", "last_connect", "current_month" }, user_item)`.
* `_copy_new_data` does nothing.
* `_check_collision` sets `existing_webpush_rotator = True`, since the old record has a `current_month`.
* `_check_collision` calls `_check_other_nodes`.
* `_check_other_nodes` disconnects dupes, clears the hello timeout, and responds to the client.

## Case 5: client with `router_type = "simplepush"` now opts in to Web Push

This can happen when a user upgrades from Firefox 42 to 43+.

* Call `_register_user(existing_user=True)`.
* Since `existing_user = True`, we don't set `last_connect` or `current_month`.
* Call `router.register_user(user_item)`.
* The conditional check for `(router_type = :router_type)` fails.
* `router.register_user` returns `(False, {}, user_item)`.
* `_copy_new_data` does nothing.
* `_check_collision` sets `existing_webpush_rotator = False`.
* Reset the UAID and call `_register_user(existing_user=False)`.
* Follow steps for case 1.

PR #311, opened 2016-01-26
feat: add maintenance.py script for use in AWS Lambda

### LINKED ISSUE #254: Lambda functions for ensuring rotating tables are present
Issue #191 and #253 add rotating table usage for messages and the router table, a lambda function should ensure tables for the current and next month exist, as well as the prior month.

bbangert:
Add's the next months message table automatically, and updates if
necessary to the latest read/write units from the current month.

Closes #254

### DIFF
--- a/maintenance.py
+++ b/maintenance.py
@@ -1,0 +1,125 @@
+"""Maintenance scripts
+
+These scripts are isolated from ``main.py`` to reduce the package dependencies
+when running under AWS Lambda.
+
+This module avoids non-stdlib imports to work cleanly in AWS Lambda. Using the
+functions in the db module would be cleaner but autopush uses boto2, while
+AWS Lambda uses Boto 3. The db module also has additional imports requiring
+substantially more packages.
+
+The IAM execution context should have a policy similar to the following to
+ensure it can read/write DynamoDB properly:
+
+..
+
+    {
+        "Version": "2012-10-17",
+        "Statement": [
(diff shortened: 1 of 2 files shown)


maintenance.py
```diff
@@ -0,0 +1,125 @@
...
+    # Get the current provisioned limits for this months table
+    cur_mth_tbl = dynamodb.Table(make_rotating_tablename(MESSAGE_TABLE_PREFIX))
+    limits = cur_mth_tbl.provisioned_throughput
+    read_units = limits["ReadCapacityUnits"]
+    write_units = limits["WriteCapacityUnits"]
+
+    # Determine if we are making the table or just verifying limits
+    nxt_mth_tblname = make_rotating_tablename(MESSAGE_TABLE_PREFIX, delta=1)
+    existing_tables = client.list_tables()["TableNames"]
+    if nxt_mth_tblname in existing_tables:
```
jrconlin: Are we not dropping older tables or is that being done by some other function?
Just figured that we could auto drop tables[:-max_tables_to_hold] or something.

bbangert: My thought was that we leave that, along with downgrading the read/write ops as a manual ops task. This way there'd be no surprised if things disappeared. Dropping a table is a simple delete op in the console, while creating one is a bit more involved, thus this script.

jrconlin:
r+

PR #316, opened 2016-01-29
feat: tag webpush notifications for more thorough logging

### LINKED ISSUE #315: Tag logged notifications on the endpoint with webpush flag
If an incoming notification is for a webpush user, the notification should be tagged as such so we can differentiate webpush notifications in the system from simplepush.

bbangert:
This tags the router responses with additional information for webpush calls
so that our logging statements can determine the three cases under which we
return a 201: delivered, stored, dropped due to TTL 0.

Fixes #315 

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ Autopush Changelog
 Features
 --------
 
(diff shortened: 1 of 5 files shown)


jrconlin:
r+ pending travis

PR #317, opened 2016-01-29
bug: capture more types of exceptions that happen when nodes disappear

### LINKED ISSUE #306: Eat Response never recieved exceptions
During deployments, autopush nodes throw an error when attemping to tell the prior node a user was connected to that the user is now on the new node.

```
ResponseNeverReceived: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionLost'>>]
ResponseNeverReceived: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionDone'>>]
```

This is because the old node was shut-down. The HTTP call in autopush that sends the 'drop this user' command should gracefully ignore this error since its expected that the other node can disappear.

bbangert:
Fixes #306 

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,12 @@ Features
   Issue #205.
 * Add log message if routing connection is refused. Issue #283.
 
+Bug Fixes
+---------
+
+* Increase the type of connection loss exceptions caught by autopush that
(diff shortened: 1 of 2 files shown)


jrconlin:
r+

PR #321, opened 2016-02-01
bug: ensure message table objects for current month are available

### LINKED ISSUE #319: endpoint doesn't run table rotation looping task
There's a looping task that the connection nodes run to ensure that they have a message table object for the current month, the endpoint nodes are missing this call such that they fail to get a message table object for the new month. This causes them to keyerror.

The endpoint should also startup the same looping task that the connection nodes do.

bbangert:
This add's the looping task that the connection nodes have that ensures
the table objects for this months message table are available for use.

Fixes #319 

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,15 @@
 Autopush Changelog
 ==================
 
+1.10.1 (**dev**)
+================
+
+Bug Fixes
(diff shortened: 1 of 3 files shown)


jrconlin:
r+ pending travis.

PR #322, opened 2016-02-01
bug: overwrite records during monthly migration to avoid disconnect edge-case

### LINKED ISSUE #320: save_channels uses a conditional check
Apparently using `put_item` in boto causes a conditional check to exist on the update. It's rather important that the channels are just saved out, otherwise the rest of the router table update won't proceed, and the transport won't resume.

The put_item should be replaced with a direct table update like our other functions to ensure the channels are correctly saved. An errback should be added to the table rotation that disconnects users that rotation fails for, since the pauseProducing means we will be unable to do anything with the user until they reconnect.

bbangert:
During monthly rotation, if a user drops the connection after the monthly
rotation has begun, but before it completed, then the message table will be
updated but not the router table. This means that the next call will fail
since put_item wasn't called with overwrite=True. This patch fixes that
and includes an integration test to verify the fix.

Closes #320

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ Autopush Changelog
 Bug Fixes
 ---------
 
+* Use non-conditional update for save_messages as put_item relies on a flakey
+  conditional check that doesn't apply in our case. Issue #320.
 * Run looping task call to update message table objects on the endpoint as well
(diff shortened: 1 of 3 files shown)


jrconlin:
r+

PR #328, opened 2016-02-04
Add VAPID Auth header check and logging

### LINKED ISSUE #325: JWT token validation and logging
Per https://github.com/martinthomson/webpush-vapid, which has now identified JWT as the voluntary sender authentication scheme, we should validate JWT information in Push notifications and log it with the push message for use by our developer dashboard and analytics.

jrconlin:
* Adds token claims to _client_info block for logging.
* Also adds "verison" to client_info block.
* Adds logging for message deletion.

Resolves #325 

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -273,7 +273,8 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
     #############################################################
     def _db_error_handling(self, d):
         """Tack on the common error handling for a dynamodb request and
-        uncaught exceptions"""
+        uncaught exceptions
+        """
         d.addErrback(self._overload_err)
         d.addErrback(self._response_err)
         return d
@@ -301,8 +302,8 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         VAPID requires two headers to be present;
         `Authorization: bearer ...` and `Crypto-Key: p256ecdsa=..`.
         The problem is that VAPID is optional and Crypto-Key can carry
-        content for other functions."""
-
+        content for other functions.
+        """
         authorization = self.request.headers.get('authorization')
         # No auth present, so it's not a VAPID call.
         if not authorization:
@@ -337,7 +338,8 @@ class MessageHandler(AutoendpointHandler):
 
     def _token_valid(self, result, func):
         """Handles valid token processing, then dispatches to supplied
-        function"""
+        function
+        """
         info = result.split(":")
         if len(info) != 3:
             raise ValueError("Wrong message token components")
@@ -550,7 +552,6 @@ class RegistrationHandler(AutoendpointHandler):
         Endpoint generation and optionally router type/data registration.
 
         """
-
         self.start_time = time.time()
         self.add_header("Content-Type", "application/json")
         params = self._load_params()
@@ -721,7 +722,8 @@ class RegistrationHandler(AutoendpointHandler):
 
     def _return_endpoint(self, endpoint_data, new_uaid, router=None):
         """Called after the endpoint was made and should be returned to the
-        requestor"""
+        requestor
+        """
         if new_uaid:
             if self.ap_settings.bear_hash_key:
                 hashed = generate_hash(self.ap_settings.bear_hash_key[0],
@@ -755,7 +757,6 @@ class RegistrationHandler(AutoendpointHandler):
 
         Validate the given request bearer token
         """
-
         test, _ = validate_uaid(uaid)
         if not test:
             return False
@@ -780,7 +781,8 @@ class RegistrationHandler(AutoendpointHandler):
 
     def _load_params(self):
         """Load and parse a JSON body out of the request body, or return an
-        empty dict"""
+        empty dict
+        """
         try:
             return json.loads(self.request.body)
         except ValueError:
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -273,8 +273,7 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
     #############################################################
     def _db_error_handling(self, d):
         """Tack on the common error handling for a dynamodb request and
-        uncaught exceptions
-        """
+        uncaught exceptions"""
         d.addErrback(self._overload_err)
         d.addErrback(self._response_err)
         return d
@@ -303,6 +302,7 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         `Authorization: bearer ...` and `Crypto-Key: p256ecdsa=..`.
         The problem is that VAPID is optional and Crypto-Key can carry
         content for other functions.
+
         """
         authorization = self.request.headers.get('authorization')
         # No auth present, so it's not a VAPID call.
@@ -338,8 +338,7 @@ class MessageHandler(AutoendpointHandler):
 
     def _token_valid(self, result, func):
         """Handles valid token processing, then dispatches to supplied
-        function
-        """
+        function"""
         info = result.split(":")
         if len(info) != 3:
             raise ValueError("Wrong message token components")
@@ -356,6 +355,7 @@ class MessageHandler(AutoendpointHandler):
         The message will only be removed from DynamoDB. Messages that were
         successfully routed to a client as direct updates, but not delivered
         yet, will not be dropped.
+
         """
         self.version = self._client_info['version'] = token
         d = deferToThread(self.ap_settings.fernet.decrypt,
@@ -722,8 +722,7 @@ class RegistrationHandler(AutoendpointHandler):
 
     def _return_endpoint(self, endpoint_data, new_uaid, router=None):
         """Called after the endpoint was made and should be returned to the
-        requestor
-        """
+        requestor"""
         if new_uaid:
             if self.ap_settings.bear_hash_key:
                 hashed = generate_hash(self.ap_settings.bear_hash_key[0],
@@ -756,6 +755,7 @@ class RegistrationHandler(AutoendpointHandler):
         """Validates the Authorization header in a request
 
         Validate the given request bearer token
+
         """
         test, _ = validate_uaid(uaid)
         if not test:
@@ -781,8 +781,7 @@ class RegistrationHandler(AutoendpointHandler):
 
     def _load_params(self):
         """Load and parse a JSON body out of the request body, or return an
-        empty dict
-        """
+        empty dict"""
         try:
             return json.loads(self.request.body)
         except ValueError:
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -27,9 +27,6 @@ import time
 import urlparse
 import uuid
 import re
-import ecdsa
-import jws
-import base64
 
 from collections import namedtuple
 from base64 import urlsafe_b64encode
@@ -50,7 +47,8 @@ from autopush.utils import (
     generate_hash,
     validate_uaid,
     ErrorLogger,
-    parse_header
+    parse_header,
+    extract_jwt,
 )
 
 # Our max TTL is 60 days realistically with table rotation, so we hard-code it
@@ -66,6 +64,11 @@ status_codes = {
     500: "Internal Server Error",
     503: "Service Unavailable",
 }
+
+
+class VapidAuthException(Exception):
+    """Exception if the VAPID Auth token fails"""
+    pass
 
 
 class Notification(namedtuple("Notification",
@@ -118,9 +121,11 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         """Setup basic aliases and attributes"""
         self.uaid_hash = ""
         self._uaid = ""
+        self._chid = ""
         self.ap_settings = ap_settings
         self.metrics = ap_settings.metrics
         self.request_id = str(uuid.uuid4())
+        self._client_info = self._init_info()
 
     def prepare(self):
         """Common request preparation"""
@@ -155,8 +160,20 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         """Set the UAID and update the uaid hash"""
         self._uaid = value
         self.uaid_hash = hashlib.sha224(self.uaid).hexdigest()
+        self._client_info["uaid_hash"] = self.uaid_hash
 
-    def _client_info(self):
+    @property
+    def chid(self):
+        """Return the ChannelID"""
+        return self._chid
+
+    @chid.setter
+    def chid(self, value):
+        """Set the ChannelID and record to _client_info"""
+        self._chid = value
+        self._client_info["channelID"] = value
+
+    def _init_info(self):
         """Returns a dict of additional client data"""
         return {
             "request_id": self.request_id,
@@ -164,12 +181,6 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
             "remote-ip": self.request.headers.get("x-forwarded-for",
                                                   self.request.remote_ip),
             "authorization": self.request.headers.get("authorization", ""),
-            "uaid_hash": self.uaid_hash,
-            "router_key": getattr(self, "router_key", ""),
-            "channel_id": getattr(self, "chid", ""),
-            "audience": getattr(self, "aud", ""),
-            "submitter": getattr(self, "sub", ""),
-            "expires": getattr(self, "exp", ""),
         }
 
     #############################################################
@@ -199,13 +210,13 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         running.
 
         """
-        log.err(fail, **self._client_info())
+        log.err(fail, **self._client_info)
         self._write_response(500, 999)
 
     def _overload_err(self, fail):
         """errBack for throughput provisioned exceptions"""
         fail.trap(ProvisionedThroughputExceededException)
-        log.msg("Throughput Exceeded", **self._client_info())
+        log.msg("Throughput Exceeded", **self._client_info)
         self._write_response(503, 201)
 
     def _router_response(self, response):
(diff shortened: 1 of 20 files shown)


autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            try:
+                jws.verify(head, payload, sig, vk)
+            except Exception:
+                # failure. Probably should log this...
+                raise VapidAuthException("Invalid bearer token")
+        return payload
+
+    def _process_auth(self, result):
+        crypto_key = parse_header(
+            self.request.headers.get('crypto-key')).get('p256ecdsa')
```
bbangert: If there's no crypto-key, this returns None which errors on calling get.

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
         d.addErrback(self._response_err)
         return d
 
+    def _extract_jwt(self, token, crypto_key):
+        # first split and convert the jwt.
+        sig = None
+        payload = {}
+
+        def fix_padding(string):
+            return string + '===='[len(sig) % 4:]
```
bbangert: I assume this is for right padding? If so....
```python
sig = "{:=<4}".format(sig)
```
per https://docs.python.org/2/library/string.html#formatexamples under 'Aligning the text and specifying a width'.

As is, its quite confusing that this inline function uses sig from outside its scope, even though the string its passed is the sig on line 286, but then on line 292 it right-pads the crypto_key based on the sig length?

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+
+        try:
+            (ehead, epayload, sig) = token.split('.', 2)
+            sig = fix_padding(sig)
```
bbangert: This changes the behavior of the next call to fix_padding, since fix_padding appends = based on the length of sig, and this call changes the length of sig. Later calls will always fail to append anything since sig is now right-padded.

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
         d.addErrback(self._response_err)
         return d
 
+    def _extract_jwt(self, token, crypto_key):
```
bbangert: Nothing in this function uses `self`. The function should probably reside elsewhere, maybe utils.py.

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+        try:
+            (ehead, epayload, sig) = token.split('.', 2)
+            sig = fix_padding(sig)
+        except ValueError:
+            raise VapidAuthException("Invalid bearer token")
+        head = json.loads(base64.urlsafe_b64decode(ehead))
+        payload = json.loads(base64.urlsafe_b64decode(epayload))
+        if sig and crypto_key:
+            key = base64.urlsafe_b64decode(fix_padding(crypto_key))
+            vk = ecdsa.VerifyingKey.from_string(key, curve=ecdsa.NIST256p)
```
bbangert: This function doesn't throw exceptions if the key is invalid?

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            return string + '===='[len(sig) % 4:]
+
+        try:
+            (ehead, epayload, sig) = token.split('.', 2)
+            sig = fix_padding(sig)
+        except ValueError:
+            raise VapidAuthException("Invalid bearer token")
+        head = json.loads(base64.urlsafe_b64decode(ehead))
+        payload = json.loads(base64.urlsafe_b64decode(epayload))
+        if sig and crypto_key:
```
bbangert: If this function is being called, it means there's a crypto_key and authorization header, shouldn't the sig/crypto_key be required at this point?

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            key = base64.urlsafe_b64decode(fix_padding(crypto_key))
+            vk = ecdsa.VerifyingKey.from_string(key, curve=ecdsa.NIST256p)
+            try:
+                jws.verify(head, payload, sig, vk)
```
bbangert: The return value is not used in any way?

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            except Exception:
+                # failure. Probably should log this...
```
bbangert: Since this is a blanket exception, it should be logged. Specific exceptions should be caught when possible.

autopush/endpoint.py
jrconlin: crap, missed that one. thanks!

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            (auth_type, token) = authorization.split(' ', 1)
+            # if it's a bearer token containing what may be a JWT
+            if auth_type.lower() == 'bearer' and '.' in token:
+                try:
+                    jwt = self._extract_jwt(token, crypto_key)
+                    self.aud = jwt.get('aud')
+                    self.exp = jwt.get('exp')
+                    self.sub = jwt.get('sub')
+                except Exception:
+                    # It failed, for reasons, but for now let it slide.
```
bbangert: Blanket exception that should be logged, the specific exceptions should be caught individually.

autopush/endpoint.py
jrconlin: the token sig is passed as an argument to jws.verify(), along with the validating key generated from the crypto-key. I'll make that a bit more obvious when I move the code to utils.

autopush/endpoint.py
jrconlin: Oddly, no. jws.verify either returns "true" or throws an exception.

autopush/endpoint.py
jrconlin: Fair. One of the problems is that some functions return less than helpful errors like (and I'm not making this one up) "AssertionError: (3, 64)". There's not a huge amount one can do with that.

autopush/endpoint.py
bbangert: :grimacing:

autopush/endpoint.py
```diff
@@ -254,6 +273,49 @@ def _db_error_handling(self, d):
...
+            self.request.headers.get('crypto-key')).get('p256ecdsa')
+        authorization = self.request.headers.get('authorization')
+        if crypto_key and authorization:
+            (auth_type, token) = authorization.split(' ', 1)
+            # if it's a bearer token containing what may be a JWT
+            if auth_type.lower() == 'bearer' and '.' in token:
+                try:
+                    jwt = self._extract_jwt(token, crypto_key)
+                    self.aud = jwt.get('aud')
+                    self.exp = jwt.get('exp')
```
bbangert: Shouldn't we verify that the token hasn't expired using exp?

bbangert: Actually, I assume the verify function already verifies that. In which case, why do we need to log it?

autopush/utils.py
```diff
@@ -54,6 +58,65 @@ def generate_hash(key, payload):
     return h.hexdigest()
 
 
+def parse_header(header, major=";", minor="="):
+    """ Convert a multi-component header line (e.g. "a;b=c;d=e;...") to
+    a list.
+
+    For example, if the header content line is
+
+    `"a;c=1;b;d=2+3=5"`
```
bbangert: If cyclone doesn't strip quoting, it's also possible this could be called with:
```
'a;c="fred";d="smith"'
```
It should strip quotes from the value.

autopush/utils.py
```diff
@@ -54,6 +58,65 @@ def generate_hash(key, payload):
...
+    `['a', 'b', {'c': '1', 'd': '2+3=5'}]`
+
+    items defined with values will always appear as a dictionary at the
+    end of the list. If no items are assigned values, then no dictionary
+    is appended.
+
+    :param header: Header content line to parse.
+    :param major: Major item separator.
+    :param minor: Minor item separator.
+    """
```
bbangert: nit: line-break before the triple-quote, no line-break after it

autopush/endpoint.py
```diff
@@ -254,6 +271,41 @@ def _db_error_handling(self, d):
         d.addErrback(self._response_err)
         return d
 
+    def _process_auth(self, result):
```
bbangert:
```python
if 'crypto-key' not in self.request.headers:
    return result
```

Would tidy this up quite a bit, the if header_info indented block wouldn't be needed, nor the crypto_key check at 286, nor the None assign to crypto_key. Adding a block to check for authorization and punt would remove that whole layer of indentation as well.

Yea, I know I'm a big fan of doing a check and returning early rather than adding levels of indentation. :)

bbangert: Actually, if that was:
```python
if 'crypto-key' not in self.request.headers or 'authorization' not in self.request.headers:
    return result
```

Would wipe out both levels of if statements and indentation....

autopush/endpoint.py
```diff
@@ -254,6 +271,41 @@ def _db_error_handling(self, d):
...
+            # if it's a bearer token containing what may be a JWT
+            if auth_type.lower() == 'bearer' and '.' in token:
+                try:
+                    jwt = extract_jwt(token, crypto_key)
+                    if jwt.get('exp') and jwt.get('exp') < time.time():
+                        raise Exception("Auth expired")
+                    self.aud = jwt.get('aud')
+                    self.exp = jwt.get('exp')
+                    self.sub = jwt.get('sub')
+                except Exception, exc:
```
bbangert:
```python
except Exception as exc
```

autopush/endpoint.py
```diff
@@ -254,6 +271,41 @@ def _db_error_handling(self, d):
...
+                raise VapidAuthException("Invalid bearer token: "
+                                         "improperly specified crypto-key")
+        authorization = self.request.headers.get('authorization')
+        if crypto_key and authorization:
+            (auth_type, token) = authorization.split(' ', 1)
+            # if it's a bearer token containing what may be a JWT
+            if auth_type.lower() == 'bearer' and '.' in token:
+                try:
+                    jwt = extract_jwt(token, crypto_key)
+                    if jwt.get('exp') and jwt.get('exp') < time.time():
```
bbangert: The get isn't needed the second time since it must exist to get there the first time.

bbangert: Probably better to just get exp out on its own, so all the extra get() calls aren't needed (like the one below at 295).

autopush/endpoint.py
```diff
@@ -163,6 +170,10 @@ def _client_info(self):
             "uaid_hash": self.uaid_hash,
             "router_key": getattr(self, "router_key", ""),
             "channel_id": getattr(self, "chid", ""),
+            "audience": getattr(self, "aud", ""),
+            "submitter": getattr(self, "sub", ""),
+            "expires": getattr(self, "exp", ""),
+            "version": getattr(self, "version", ""),
```
bbangert: All these getattr calls are not cheap, we're adding more, and increasing our logging which is going to make the delivery of every message increasingly more CPU expensive. We should hook onto the initialization of the Handler class, and ensure all these default to "", so that all the getattr's can be removed.

autopush/endpoint.py
jrconlin: heh, that's coming out regardless

autopush/utils.py
jrconlin: Oh, it's better than that. It could be called as 
a;c="fred";d=""george "smith"""

Fortunately, python makes handling that remarkably simple

autopush/endpoint.py
```diff
@@ -114,9 +121,11 @@ def initialize(self, ap_settings):
         """Setup basic aliases and attributes"""
         self.uaid_hash = ""
         self._uaid = ""
+        self._chid = ""
```
bbangert: Where's router_key?

autopush/endpoint.py
```diff
@@ -151,8 +160,20 @@ def uaid(self, value):
...
+        self._client_info["uaid_hash"] = self.uaid_hash
+
+    @property
+    def chid(self):
+        """Return the ChannelID"""
+        return self._chid
 
-    def _client_info(self):
+    @chid.setter
+    def chid(self, value):
```
bbangert: Nice! I like this approach of updating the client_info when chid is set.

autopush/endpoint.py
```diff
@@ -254,6 +278,43 @@ def _db_error_handling(self, d):
         d.addErrback(self._response_err)
         return d
 
+    def _store_auth(self, jwt, result):
+        if jwt.get('exp') and jwt['exp'] < time.time():
+            raise VapidAuthException("Invalid bearer token: Auth expired")
+        self._client_info.update(jwt)
```
bbangert: How many entries are in the jwt dict? Do all of them need to be in the client_info? Might be useful to nest them in a sub-dict in case its possible any of the jwt entries might smash an existing entry.

autopush/endpoint.py
```diff
@@ -254,6 +278,43 @@ def _db_error_handling(self, d):
...
+    def _invalid_auth(self, fail):
+        if isinstance(fail.value, VapidAuthException):
+            raise fail.value
+        message = fail.value.message or repr(fail.value)
+        if isinstance(fail.value, AssertionError):
+            message = "A decryption error occurred"
+        log.msg("Invalid bearer token: " + message, **self._client_info)
+        raise VapidAuthException("Invalid bearer token: " + message)
+
+    def _process_auth(self, result):
```
bbangert: Did all my comments about how to refactor this function disappear from the client_info commit? I think they'd really clean this function up a bit more.

autopush/tests/test_endpoint.py
```diff
@@ -3,6 +3,10 @@
 import sys
 import time
 import uuid
+import base64
+import ecdsa
+import jws
+import hashlib
```
bbangert: PEP 8 has std lib, then 3rd party, then local package, line-separated.

autopush/utils.py
```diff
@@ -1,6 +1,10 @@
 """A small collection of Autopush utility functions"""
+import base64
+import ecdsa
 import hashlib
 import hmac
+import json
+import jws
 import socket
 import uuid
```
bbangert: Not PEP 8.

autopush/websocket.py
```diff
@@ -118,6 +118,11 @@ class PushState(object):
         'connected_at',
         'settings',
 
+        # Voluntary Auth
+        'public_key',  # site public key for vol. auth.
+        'host',  # site associated with this public key.
+        'contact',  # contact info associated with this feed.
```
bbangert: Shouldn't be in this PR.

bbangert:
r-, still needs some work as noted

groovecoder:
Sorry to abuse this PR to ask out of the blue, but ...

I don't see any code here that logs or stores the JWS signature (i.e., `Crypto-Key` `p256ecdsa` value?) with a message? I'm planning to let a developer store their JWS signature in the dev dashboard to query the push service for their message data.

Am I overlooking something that associates messages with the JWS signature? (The `PushState.public_key` has it and is persisted somewhere?) Or am I mis-understanding how I should query the push service for a developer's messages' data?

bbangert:
The vapid auth information is logged along with the notification version, so its captured by our logging infrastructure.

bbangert:
Sorry, I should note, we are not storing the raw crypto-key with that data. Only the validated attributes for the token (aud/exp/sub).

groovecoder:
Ah, okay. Then it's back to the simple implementation I originally thought it was? I.e., I ask devs to verify their authorization for domain(s), and then I query the push service data for messages with `aud` value matching the domain/origin I validated? Then, optionally, ask devs to verify their email address, and I can query the push service data for messages with `sub` value matching the email I validated?

bbangert:
Yup, that'd work.

autopush/endpoint.py
```diff
@@ -254,6 +278,58 @@ def _db_error_handling(self, d):
...
+        """Process the optional VAPID auth token.
+
+        VAPID requires two headers to be present;
+        `Authorization: bearer ...` and `Crypto-Key: p256ecdsa=..`.
+        The problem is that VAPID is optional and Crypto-Key can carry
+        content for other functions."""
```
bbangert: nit: triple quote needs a line break drop, no line break after it before the code starts

autopush/endpoint.py
```diff
@@ -639,10 +722,11 @@ def _create_endpoint(self, result=None):
 
     def _return_endpoint(self, endpoint_data, new_uaid, router=None):
         """Called after the endpoint was made and should be returned to the
-        requestor"""
+        requestor
+        """
```
bbangert: nit: no line-break for single-sentence summaries that happen to be too long for a single-line

autopush/endpoint.py
```diff
@@ -698,7 +781,8 @@ def _validate_auth(self, uaid):
 
     def _load_params(self):
         """Load and parse a JSON body out of the request body, or return an
-        empty dict"""
+        empty dict
+        """
```
bbangert: nit: no line-break for triple quote here

bbangert:
Looks good, there is one important thing I'd like to see added, which is an integration test. Mainly because this does add some flow bits, so it'd be nice to be re-assured this will work in the integration test.

jrconlin:
I'll note that the individual unit tests actually do pass valid decryption data as part of their test and check for the result in the _client_info. The encryption/decryption jws and ecdsa functions are not mocked. Any integration test would duplicate the function of these unit tests.

autopush/tests/test_integration.py
```diff
@@ -652,6 +686,16 @@ def test_basic_delivery(self):
         yield self.shut_down(client)
 
     @inlineCallbacks
+    def test_basic_delivery_with_vapid(self):
+        data = str(uuid.uuid4())
+        client = yield self.quick_register(use_webpush=True)
+        result = yield client.send_notification(data=data, vapid=True)
+        eq_(result["headers"]["encryption"], client._crypto_key)
+        eq_(result["data"], urlsafe_b64encode(data))
+        eq_(result["messageType"], "notification")
```
bbangert: Fetch the notification so we know it made it through?

docs/style.rst
```diff
@@ -0,0 +1,14 @@
...
+
+Autopush uses Python styling guides based on `PEP8<https://www.python.org/dev/peps/pep-0008/>`_ and `PEP257<https://www.python.org/dev/peps/pep-0257/>`_.
+
+Exceptions
+----------
+
+* Single sentence docstrings are formatted the same way as a single line docstring, but may not always include ending punctuation.
+
+* File level docstrings may not include a line break before the first line of code.
+
```
bbangert: :+1:

bbangert:
r+, :100: :+1:

Issue #334, opened 2016-02-08
Add routing identifier header to returned response


jrconlin:
To help determine cause of some errors like #255, add a header to indicate what routing a given channel uses. 

This header should only be added after a successful route or storage operation is performed.

jrconlin:
closed per #336

PR #336, opened 2016-02-11
Add Router header to successful outpound subscription updates

jrconlin:
In order to assist in determining errors with users, it would be useful to know how a given message was routed. The "Routing" header provides how the system believes the message should be routed ("simplepush", "webpush", "gcm", etc.)

Resolves: #334
@bbangert r?


jrconlin:
Closing in favor of always returning "Location". This provides a better token for lookup and diagnostic.

PR #339, opened 2016-02-12
bug: use appropriate 400, 410 instead of 404 for permanent errors

### LINKED ISSUE #312: Use 410 Gone instead of 404 Not Found
Per [RFC 7231](http://tools.ietf.org/html/rfc7231), [404 Not Found](http://tools.ietf.org/html/rfc7231#section-6.5.4) "does not indicate whether this lack of representation is temporary or permanent; the 410 (Gone) status code is preferred over 404 if the origin server knows, presumably through some configurable means, that the condition is likely to be permanent."

As far as I can tell, everything we return 404 for is indeed permanent. Perhaps we should switch to [410 Gone](http://tools.ietf.org/html/rfc7231#section-6.5.9), which "indicates that access to the target resource is no longer available at the origin server and that this condition is likely to be permanent. [...] The 410 response is primarily intended to assist the task of web maintenance by notifying the recipient that the resource is intentionally unavailable and that the server owners desire that remote links to that resource be removed."

bbangert:
Previously 404 was returned for multiple classes of errors when the endpoints
would never in fact be good, so a 410 is appropriate. In one case a 404 was
returned instead of a 400 which properly addresses the fact that the URL was
not validly constructed.

Closes #312 

@jrconlin r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,9 @@ Features
 Bug Fixes
 ---------
 
-*  Do not send useless 'ver' across GCM bridge. Issue #323.
+* Use appropriate 400, 404, 410 status codes for differing message endpoint
+  results, rather than always a 404. Issue #312.
+* Do not send useless 'ver' across GCM bridge. Issue #323.
 
 1.10.1 (2016-02-01)
 ===================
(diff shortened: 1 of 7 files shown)


jrconlin:
r+ pending travis

PR #340, opened 2016-02-12
Require TTL header for incoming Subscription events.

### LINKED ISSUE #329: Enforce TTL header on incoming Push publish requests
All incoming push publications should require a TTL header. 
Requests missing the TTL should return a 400 response ("Missing TTL header"). 
The TTL should also be included in the allowed CORS headers.

jrconlin:
Resolves #329 

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -381,7 +381,7 @@ class MessageHandler(AutoendpointHandler):
 class EndpointHandler(AutoendpointHandler):
     cors_methods = "POST,PUT"
     cors_request_headers = ["content-encoding", "encryption",
-                            "crypto-key",
+                            "crypto-key", "ttl",
                             "encryption-key", "content-type"]
     cors_response_headers = ["location"]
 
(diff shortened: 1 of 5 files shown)


autopush/endpoint.py
```diff
@@ -381,7 +381,7 @@ def _delete_completed(self, response):
 class EndpointHandler(AutoendpointHandler):
     cors_methods = "POST,PUT"
     cors_request_headers = ["content-encoding", "encryption",
-                            "crypto-key",
+                            "crypto-key", "ttl",
```
kitcambridge: Ooh, good catch!

autopush/tests/test_integration.py
```diff
@@ -215,14 +215,9 @@ def send_notification(self, channel=None, version=None, data=None,
...
-                    headers.update({
-                        'Crypto-Key': headers.get('Crypto-Key') + ';' + ckey
-                    })
-                else:
-                    headers.update({
-                        'Crypto-Key': ckey
-                    })
+                headers.update({
+                    'Crypto-Key': headers.get('Crypto-Key') + ';' + ckey
+                })
```
jrconlin: Why the reduced code? Apparently, we never call vapid with use_header == False.

bbangert: Yup, got a warning locally about missing code coverage, so I removed it.

bbangert:
r+, just needs rebase

PR #341, opened 2016-02-12
fix: Whitelist the "Authorization" header for CORS requests

kitcambridge:
Also add a "WWW-Authenticate" header for 401 responses, for maximum RFC 7235 compliance. :wink:

r=@bbangert

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -53,6 +53,7 @@ from autopush.utils import (
 
 # Our max TTL is 60 days realistically with table rotation, so we hard-code it
 MAX_TTL = 60 * 60 * 24 * 60
+AUTH_SCHEME = "bearer"
 status_codes = {
     200: "OK",
     201: "Created",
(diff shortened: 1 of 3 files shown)


bbangert:
I still see it as 'Bearer' in the vapid spec, not 'WebPush' for the Authorization header.

autopush/endpoint.py
```diff
@@ -320,7 +325,7 @@ def _process_auth(self, result):
             except ValueError:
                 raise VapidAuthException("Invalid Authorization header")
             # if it's a bearer token containing what may be a JWT
-            if auth_type.lower() == 'bearer' and '.' in token:
+            if auth_type.lower() == AUTH_SCHEME and '.' in token:
```
jrconlin: We probably want to make this 
   if auth_type.lower() in [AUTH_SCHEME, 'bearer'] and ...
for a few releases. There are a few folks in the wild using 'bearer' and we don't want to break them yet.

autopush/endpoint.py
```diff
@@ -768,7 +774,7 @@ def _validate_auth(self, uaid):
                                         header.strip()).split(" ", 2)
         except ValueError:
             return False
-        if "bearer" != token_type.lower():
+        if AUTH_SCHEME != token_type.lower():
```
jrconlin: See above.

bbangert:
Just missing a changelog.... :grin:

PR #342, opened 2016-02-12
Add "Location" to outbound webpush responses

### LINKED ISSUE #338: Always include Location header
Right now TTL 0 messages do not have a Location header. Per the spec they should though. We should also include a Location header of a similar nature that is logged for simplepush.

This would allow us to have a standard way to lookup messages in our system if an application developer believes there was a problem, and we could write some documentation for application developers submitting bugs about what information they should capture to report (the Location header).

jrconlin:
fixes #338 
@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -183,11 +183,6 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
             "authorization": self.request.headers.get("authorization", ""),
         }
 
-    def _location(self):
-        """Returns a location header value created from the version"""
-        return "%s/m/%s" % (self.ap_settings.endpoint_url,
-                            self.version)
-
     def _is_webpush(self):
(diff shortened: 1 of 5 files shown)


bbangert:
No changelog? ;)

autopush/endpoint.py
```diff
@@ -435,7 +440,7 @@ def _uaid_lookup_results(self, result):
 
         # Only simplepush uses version/data out of body/query, GCM/APNS will
         # use data out of the request body 'WebPush' style.
-        if router_key == "simplepush":
+        if not self._is_webpush():
```
bbangert: Function calls are still expensive on cpython, so I'd prefer to avoid easily avoidable ones like this in a hot path until we're on pypy.

bbangert:
r+ pending removal of the extra function call in endpoint

PR #343, opened 2016-02-16
feat: log notification acks

### LINKED ISSUE #331: Log notifications delivered via autopush
While we log incoming notifications for analytics, we don't know if/when they're actually delivered based on that. autopush should log notifications as they're delivered, as well as a flag indicating if they were delivered directly or out of storage.

bbangert:
Log all notification ack's for simplepush/webpush, along with the router
key, message/channel ids, and uaid hash. Update logging on autoendpoint
to match format and keys for correlation.

Closes #331 

@jrconlin r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -21,7 +21,6 @@ For a discussion on how to use the endpoints listed here, please refer
 to :ref:`http`.
 
 """
-import hashlib
 import json
 import time
 import urlparse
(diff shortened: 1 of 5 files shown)


autopush/endpoint.py
```diff
@@ -159,7 +161,7 @@ def uaid(self):
     def uaid(self, value):
         """Set the UAID and update the uaid hash"""
         self._uaid = value
-        self.uaid_hash = hashlib.sha224(self.uaid).hexdigest()
+        self.uaid_hash = hasher(value)
```
jrconlin: ah, cool! yeah, doing a fast abstraction on this is a good idea.

autopush/endpoint.py
```diff
@@ -437,7 +439,7 @@ def _uaid_lookup_results(self, result):
         # use data out of the request body 'WebPush' style.
         if router_key == "simplepush":
             self.version, data = parse_request_params(self.request)
-            self._client_info['version'] = self.version
+            self._client_info['message_id'] = self.version
```
jrconlin: oh, thank god

jrconlin:
r+ pending

PR #347, opened 2016-02-17
bug: handle well-known exported pub key formats

### LINKED ISSUE #346: Handle multiple "raw" public key formats
WebCrypto generates 'raw' public keys slightly differently than the base ecdsa library. There are two major forms for WebCrypto to export keys in: 'raw' (which prefixes the keys with a '0x4', and spki (which prefixes with a header of '0V0\x10\x06\x04+\x81\x04p\x06\x08*\x86H\xce=\x03\x01\x07\x03B\x00\x04'. It should be safe to check for:

true raw: 64 byte record.
WebCrypto 'raw': 65 bytes with a header of '0x4'
WebCrypto 'spki': 88 bytes with a header that begins with '0v0'

Additional public key formats may need to be handled as well.

jrconlin:
@bbangert r?

Fixes #346

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -108,13 +108,39 @@ def fix_padding(string):
     return string
 
 
+def decipher_public_key(key_data):
+    """A public key may come in several flavors. Attempt to extract the
+    valid key bits from keys doing minimal validation checks.
+
+    This is mostly a result of libs like WebCrypto prefixing data to "raw"
+    keys, and the ecdsa library not really providing helpful errors.
+
+    :param key_data: the raw-ish key we're going to try and process
+    :returns: the raw key data.
+    :raises: ValueError for unknown or poorly formatted keys.
+
+    """
+    # key data is actually a raw coordinate pair
+    key_len = len(key_data)
+    if key_len == 64:
+        return key_data
+    # Key format is "raw"
+    if key_len == 65 and key_data[0] == '\x04':
+        return key_data[-64:]
+    # key format is "spki"
+    if key_len == 88 and key_data[:3] == '0V0':
+        return key_data[-64:]
+    raise ValueError("Unknown public key format specified")
+
+
 def extract_jwt(token, crypto_key):
-    """ Extract the claims from the validated JWT. """
+    """Extract the claims from the validated JWT. """
     # first split and convert the jwt.
     if not token or not crypto_key:
         return {}
 
     key = base64.urlsafe_b64decode(fix_padding(crypto_key))
+    key = decipher_public_key(key)
     vk = ecdsa.VerifyingKey.from_string(key, curve=ecdsa.NIST256p)
     return jws.verify(token, vk, algorithms=["ES256"])
 
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -4,6 +4,7 @@ import sys
 import time
 import uuid
 import base64
+import random
 
 import ecdsa
 import twisted.internet.base
@@ -33,7 +34,7 @@ from autopush.db import (
 from autopush.settings import AutopushSettings
 from autopush.router.interface import IRouter, RouterResponse
 from autopush.senderids import SenderIDs
-from autopush.utils import generate_hash
+from autopush.utils import (generate_hash, decipher_public_key)
 
 mock_dynamodb2 = mock_dynamodb2()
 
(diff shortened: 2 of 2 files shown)


bbangert:
Commit should have 'Closes #346' two line-breaks down.

jrconlin:
According to the CONTRIBUTING doc: 

> In order to maintain a reference to the context of the commit, add fixes #<issue_number> if it closes a related issue or issue #<issue_number> if it's a partial fix.

Should we file a bug to fix the document?

bbangert:
Yes, we should update that. Technically it does mention what I said as well which makes it confusing:

> Footer
> 
> The footer should contain any information about Breaking Changes and is also the place to reference GitHub issues that this commit Closes.

bbangert:
Actually, I'm fine with 'Fixes', lets keep it that way. So that just needs to be moved down to the footer.

autopush/utils.py
```diff
@@ -108,13 +108,38 @@ def fix_padding(string):
...
+    """
+    # key data is actually a raw coordinate pair
+    if len(key_data) == 64:
+        return key_data
+    # Key format is "raw"
+    if len(key_data) == 65 and key_data[0] == '\x04':
+        return key_data[-64:]
+    # key format is "spki"
+    if len(key_data) == 88 and key_data[:3] == '0V0':
+        return key_data[-64:]
```
bbangert: Do the len call just once and store the result to avoid unnecessary repeat func calls.

bbangert:
r+

jrconlin:
Nope. UDP test failed again, causing testing lines to get skipped.

autopush/utils.py
```diff
@@ -108,13 +108,39 @@ def fix_padding(string):
...
+    """
+    # key data is actually a raw coordinate pair
+    key_len = len(key_data)
+    if key_len == 64:
+        return key_data
+    # Key format is "raw"
+    if key_len == 65 and key_data[0] == '\x04':
+        return key_data[-64:]
+    # key format is "spki"
+    if key_len == 88 and key_data[:3] == '0V0':
```
groovecoder: Sorry for the swoop-and-:poop: here, but how do we know the last 3 should be `0V0`? I'm asking because I'm trying to mirror [this code (and test code) in the push dashboard](https://github.com/mozilla-services/push-dev-dashboard/compare/add-vapid-key-to-apps-37?expand=1):
```
def _gen_keys():
    signing_key = ecdsa.SigningKey.generate(curve=ecdsa.NIST256p)
    verifying_key = signing_key.get_verifying_key()
    vapid_key = urlsafe_b64encode(verifying_key.to_string())
    return (signing_key, verifying_key, vapid_key)

...

    def test_extract_public_key_expected(self):
        signing_key, verifying_key, vapid_key = _gen_keys()
        self.assertEqual(vapid_key, extract_public_key(vapid_key)) 
```
But I'm getting:
```
ValueError: Unknown public key format specified
```
Because `key_data[:3]` is different each time? Is it because `utils.fix_padding` here always makes trailing characters into `=` character?

jrconlin: @groovecoder 
`key_data[:3]` should return the first three characters, not the last three. 

The code allows for three types of "raw" (or raw-ish) key specification. All the keys have been translated from URL Save base64 before any validation is done. Base64 requires strings that are on long boundaries so some strings may require padding. fix_padding checks the string and adds sufficient padding characters ( '=' ) so that the string decodes properly.  

A true raw key, (which is 64 bytes long), a tagged raw key (which is 65 bytes) and a SPKI key, (which is 88 bytes and begins with a prefix that starts with `0V0`) . Honestly, I wouldn't worry about the SPKI format. 

The tagged raw key format is what is exported from window.crypto.subtle.exportKey() and was included because i believe node.js may use WebCrypto (or at least a variant of crypto.subtle), and there are some RFCs that stipulate that format for "raw" keys.
Wow, that was a lot of extra information, but hopefully helps clarify what you're seeing.
Now, as for what you're getting:
I'm not certain I know where self.vapid_key is being set or what value it's being set to. Normally, the VAPID key is contained in
`Crypto-Key: p256ecdsa=..."
(Fun fact, `Crypto-Key` can contain multiple sub keys, each is tagged with a unique key. `autopush.utils.parse_header` attempts to deal with this. The last element ( which you can access via [:-1]) should contain a dictionary of values you can use.)

If that is the value, I'd suggest injecting a `import pdb;pdb.set_trace()` and seeing what the values are and where `extract_public_key` falls over. 

(Oh, and a PR or issue if you find a bug in my code would be delightful.)

Does that help?

groovecoder: I think so. Let me stew on it for a bit and try to integrate it into my code.

groovecoder: In my `test_extract_public_key_expected`, `vapid_key` comes from [the `_gen_keys()`](https://github.com/mozilla-services/push-dev-dashboard/compare/add-vapid-key-to-apps-37?expand=1#diff-a4ffa973bd2f6aa0f744e7a62018766eR16) helper, which makes it:
```
    signing_key = ecdsa.SigningKey.generate(curve=ecdsa.NIST256p)
    verifying_key = signing_key.get_verifying_key()
    vapid_key = urlsafe_b64encode(verifying_key.to_string())
    return (signing_key, verifying_key, vapid_key)
```
So, it's effectively `urlsafe_b64encode(ecdsa.SigningKey.generate(curve=ecdsa.NIST256p).get_verifying_key().to_string())`.

When it's passed to `extract_public_key`, it gets to `if key_len == 88 and key_data[:3] == '0V0'` and the `key_len` is indeed 88. Which makes me think that `ecdsa` library generates spki key formats? But the first 3 characters are *not* `0v0` ...

groovecoder: Bah! I was simply skipping a base64 decoding step.

PR #353, opened 2016-02-18
docs: update CONTRIBUTING.md doc to match our style

jrconlin:
This document was originally based on the angular.js work and thus
has a fair bit of dependency on angular style knowledge. Adding
examples and proper deep links makes the document easier to
understand for new-comers.

@bbangert r?


bbangert:
r+

Issue #358, opened 2016-02-22
Return a 400 for missing TTL only if the user is offline


bbangert:
To attempt to reduce the dropped messages missing a TTL header, we'll accept the message if the user is online, but warn if offline and include a better URL about why.

PR #360, opened 2016-02-22
fix: allow webpush w/no ttl & cleanup 400 logging

### LINKED ISSUE #357: Return a 400 for invalid `TTL` values
Currently, if you pass an invalid `TTL`, we'll default to `TTL: 0` (https://github.com/mozilla-services/autopush/blob/9b811eaaf5dd55d133849cce1164f0606d2148bb/autopush/endpoint.py#L476-L481). Since we're already requiring the header, I propose we reject invalid values.

bbangert:
Allows webpush notifications with no TTL header included as long
as the client is online. Otherwise a 400 is returned with a
descriptive URL including a link to the appropriate docs on how to
include a proper TTL.

Refactors how the TTL setup code is handled, and cleans up the 400
logging for consistency.

Closes #357, #358

@jrconlin r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -55,6 +55,7 @@ from autopush.utils import (
 
 # Our max TTL is 60 days realistically with table rotation, so we hard-code it
 MAX_TTL = 60 * 60 * 24 * 60
+VALID_TTL = re.compile(r'^\d+$')
 AUTH_SCHEME = "bearer"
 status_codes = {
     200: "OK",
@@ -250,6 +251,11 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
             log.msg("Success", status_code=exc.status_code,
                     logged_status=exc.logged_status or "",
                     **self._client_info)
+        elif 400 <= exc.status_code < 500:
+            log.msg("Client error", status_code=exc.status_code,
+                    logged_status=exc.logged_status or "",
+                    errno=exc.errno or "",
+                    **self._client_info)
         self._router_response(exc)
 
     def _uaid_not_found_err(self, fail):
@@ -439,6 +445,8 @@ class EndpointHandler(AutoendpointHandler):
         try:
             self.router = self.ap_settings.routers[router_key]
         except KeyError:
+            log.msg("Client error", status_code=400, errno=108,
+                    **self._client_info)
             return self._write_response(400, 108,
                                         message="Invalid router type")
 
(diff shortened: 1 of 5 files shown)


autopush/endpoint.py
```diff
@@ -447,39 +455,42 @@ def _uaid_lookup_results(self, result):
...
-                            **self._client_info)
-                    return self._write_response(
-                        400, 110, message="Invalid crypto headers")
-                self._client_info["message_size"] = len(data) if data else 0
 
-        try:
-            ttl = int(self.request.headers.get("ttl", "0"))
+        if "ttl" not in self.request.headers:
+            ttl = None
+        elif VALID_TTL.match(self.request.headers["ttl"]):
```
jrconlin: Wouldn't trapping the ValueError from int() be better? (less regex at least)

autopush/router/webpush.py
```diff
@@ -20,6 +20,8 @@
 from autopush.router.interface import RouterException, RouterResponse
 from autopush.router.simple import SimpleRouter
 
+TTL_URL = "https://webpush-wg.github.io/webpush-protocol/#rfc.section.6.2"
```
jrconlin: may want to toss these into a central file, only so we don't have to chase them down when specs change.

docs/http.rst
```diff
@@ -91,6 +91,7 @@ Unless otherwise specified, all calls return the following error codes:
    - errno 108 - Router type is invalid
    - errno 110 - Invalid crypto keys specified
    - errno 111 - Missing Required Header
+   - errno 112 - Invalid TTL header value
```
jrconlin: Do we want to call this out for each potential header? There could be a lot. Why not just say "Invalid header value" and return the header name in the message?

autopush/endpoint.py
bbangert: Throwing an exception is quite a bit more expensive than a compiled regex match call. Exception building requires a complete set of trace-back frames and all the associated objects. PyPy can optimize this out of the way, but cpython doesn't.

autopush/router/webpush.py
bbangert: I was considering that, I admit I was hoping this would be short-lived enough that it'll be removed entirely before we consider adding more (later we'll link to our docs, which will have rich information on all the error codes).

docs/http.rst
bbangert: I think it'll be ok to call it out on each header, then we can more easily sort by errno on our logging backend. Otherwise we have to be super careful to never change the text at all, or we won't be able to easily match them all for analytics. It also means we can include a rich section for each required header (or link to the MDN page for them) so that there's no chance of confusion by a developer.

jrconlin:
r+

PR #361, opened 2016-02-22
bug: include crypto-key when routing & storing messages

### LINKED ISSUE #355: Include `Crypto-Key` when routing and storing messages
Looks like we're accepting the `Crypto-Key` header (in `endpoint.py`), but not actually storing or forwarding it in https://github.com/mozilla-services/autopush/blob/master/autopush/router/webpush.py#L43-L54.

This breaks folks who use the new encryption scheme in Firefox 46+. (https://bugzilla.mozilla.org/show_bug.cgi?id=1249206)

We should also add an integration test for this.

jrconlin:
Closes #355

@bbangert r?

### DIFF
--- a/autopush/router/webpush.py
+++ b/autopush/router/webpush.py
@@ -47,10 +47,11 @@ class WebPushRouter(SimpleRouter):
             encoding=headers["content-encoding"],
             encryption=headers["encryption"],
         )
-        # AWS cannot store empty strings, so we only add the encryption-key if
-        # its present to avoid empty strings.
-        if "encryption-key" in headers:
-            data["encryption_key"] = headers["encryption-key"]
+        # AWS cannot store empty strings, so we only add these keys if
+        # they're present to avoid empty strings.
+        for name in ["encryption-key", "crypto-key"]:
+            if name in headers:
+                data[name] = headers[name]
         return data
 
     @inlineCallbacks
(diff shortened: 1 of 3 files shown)


bbangert:
r+

autopush/router/webpush.py
```diff
@@ -47,10 +47,11 @@ def _crypto_headers(self, notification):
...
         )
-        # AWS cannot store empty strings, so we only add the encryption-key if
-        # its present to avoid empty strings.
-        if "encryption-key" in headers:
-            data["encryption_key"] = headers["encryption-key"]
+        # AWS cannot store empty strings, so we only add these keys if
+        # they're present to avoid empty strings.
+        for name in ["encryption-key", "crypto-key"]:
+            if name in headers:
+                data[name] = headers[name]
```
tomoyukilabs: Perhaps should it modified to `data[name.replace('-', '_')] = headers[name]`?

jrconlin: The - to _ is more for logs than for data exchange. These values need to stay consistent between routing and eventual delivery to clients. Changing these values here could break dependency outside of the app. 

Still, nice spotting. Thanks!

kitcambridge: @tomoyukilabs You actually spotted a bug. As currently written, this will break data delivery for all clients. Thanks for catching that, and your comment on Bugzilla! :eyes: I'll send a PR soon.

Issue #349, opened 2016-02-18
Log user-agent in autopush logging


bbangert:
The ack's logged from autopush should include the user-agent, in the same key format as for autoendpoint.

PR #362, opened 2016-02-22
feat: add user-agent logging to acks

jrconlin:
Closes: #349 

@bbangert r?

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -155,7 +155,7 @@ class PushState(object):
             self._user_agent = None
         self._base_tags = []
         if self._user_agent:
-            self._base_tags.append("user-agent:%s" % self._user_agent)
+            self._base_tags.append("user_agent:%s" % self._user_agent)
         if host:
             self._base_tags.append("host:%s" % host)
 
@@ -195,6 +195,10 @@ class PushState(object):
     def message(self):
         """Property to access the currently used message table"""
         return self.settings.message_tables[self.message_month]
+
+    @property
+    def user_agent(self):
+        return self._user_agent or "None"
 
     def pauseProducing(self):
         """IProducer implementation tracking if we should pause output"""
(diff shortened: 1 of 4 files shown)


bbangert:
Type should be 'feat' not 'feature' per CONTRIBUTING.

autopush/websocket.py
```diff
@@ -1090,7 +1094,8 @@ def ver_filter(update):
             size = len(msg.data) if msg.data else 0
             log.msg("Ack", router_key="webpush", channelID=chid,
                     message_id=version, message_type="direct",
-                    message_size=size, uaid_hash=self.ps.uaid_hash)
+                    message_size=size, uaid_hash=self.ps.uaid_hash,
+                    **{'user-agent': self.ps.user_agent})
```
bbangert: All the logging keys here are using underscore, why not just user_agent=... and skip the ** dict?

jrconlin: Ah, I thought you wanted this to match how we're logging in endpoint. There, it's going in as "user-agent" thanks to pretty much the same thing via client_info.

Perfectly happy to use user_agent if that's not a problem with the logging stuff.

bbangert: Oh, my bad, ugh, we use underscore and dash randomly. How annoying. Let's file a separate bug to have a code-wide consistency to all logging keys cause this is getting out of hand already.

autopush/websocket.py
```diff
@@ -1090,7 +1094,8 @@ def ver_filter(update):
             size = len(msg.data) if msg.data else 0
             log.msg("Ack", router_key="webpush", channelID=chid,
                     message_id=version, message_type="direct",
-                    message_size=size, uaid_hash=self.ps.uaid_hash)
+                    message_size=size, uaid_hash=self.ps.uaid_hash,
+                    userr_agent=self.ps.user_agent)
```
bbangert: typo: userrrr

bbangert:
r+

Issue #366, opened 2016-02-24
Set TTL back to a valid int before sending to a push node


bbangert:
TTL is set to None for missing header, but needs to be set back to 0 before being sent to a push node.

PR #367, opened 2016-02-24
bug: silence missing TTL errors from sentry log

jrconlin:
Missing TTL headers are being logged resulting in high levels of noise
in the Sentry logs. Set the "log_exception" param to False to prevent
the message from going to the system log.

issue #366

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -246,7 +246,7 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         fail.trap(RouterException)
         exc = fail.value
         if exc.log_exception:
-            log.err(fail, **self._client_info)
+            log.err(fail, **self._client_info)  # pragma nocover
         if 200 <= exc.status_code < 300:
             log.msg("Success", status_code=exc.status_code,
                     logged_status=exc.logged_status or "",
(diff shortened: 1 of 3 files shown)


autopush/endpoint.py
```diff
@@ -246,7 +246,7 @@ def _router_fail_err(self, fail):
         fail.trap(RouterException)
         exc = fail.value
         if exc.log_exception:
-            log.err(fail, **self._client_info)
+            log.err(fail, **self._client_info)  # pragma nocover
```
jrconlin: Tests should mimic reality, and via this bug, we know this line works. Setting `nocover` for now.

bbangert:
Can you modify the commit so it follows the standard? Otherwise the changelog builder won't indicate what issue this closes. It needs to be: "Closes #366".

jrconlin:
CONTRIBUTING.md notes: 

> In order to maintain a reference to the context of the commit, add
`closes #<issue_number>` if it closes a related issue or *`issue #<issue_number>`
if it's a partial fix*.

(emphasis added)
This does not fully address #366, but only address the extra logging argument being generated.

jrconlin:
Ok, crafted #370 so that this fixes a stand alone bug. #366 didn't quite cover this case like I believed it to.

bbangert:
r+, pending travis

PR #372, opened 2016-02-25
bug: Normalize encryption headers

kitcambridge:
#361 was a bit too aggressive. Firefox expects `Crypto-Key` and `Encryption-Key` to be converted to `crypto_key` and `encryption_key`.

This should be part of tomorrow's deploy.

@jrconlin r?


jrconlin:
r+

We may want to add a unit test to verify that headers are set to lower case w/ s/-/_/g in a later PR.

PR #375, opened 2016-02-25
bug: change message_type to message_source

jrconlin:
Logging considers "message_type" to be a reserved term. This causes us
to lose some logging information.

@bbangert r?


bbangert:
We need to copy the message_type back into Type on the underlying object as well.

bbangert:
Also, 'bug:' -> 'fix:' for consistency.

bbangert:
r+

PR #376, opened 2016-02-25
bug: pass TTL Header value to GCM

### LINKED ISSUE #374: Pass user supplied TTL to GCM
Currently, gcm bridge uses a parameter provided TTL value. It should use the TTL provided by the user header. 

Be sure to normalize "None" to "0".

jrconlin:
The code was previously using the configuration option of gcm_ttl
instead of the user supplied TTL. Instead, use gcm_ttl as the minimum
TTL in case of message loss due to TTL 0.

closes #374 

@bbangert


bbangert:
r+

PR #382, opened 2016-02-29
feature: allow channels to register with public key

### LINKED ISSUE #326: Support channel association with a JWT key
Channel subscriptions will allow association with a sender key per:
https://github.com/w3c/push-api/pull/182

Our register command will need updating to support the additional information, and our channel-id scheme of a list of channels may need modification to store an additional crypto key with it.

The endpoint will need an additional check to ensure the sender key validates with the channel key if present.

Depends on: #325

jrconlin:
Channels can include a public key when registering a new subscription
channel. This public key should match the public key used to send
subscription updates later.

NOTE: this patch changes the format of the endpoint URLs, & the content
of the endpoint URL token. This change also requires that ChannelIDs be
normalized to dashed format, (e.g. a lower case, dash delimited string
"deadbeef-0000-0000-deca-fbad11112222") This is the default mechanism
used by Firefox for UUID generation. It is STRONGLY urged that clients
normalize UUIDs used for ChannelIDs and User Agent IDs. While this
should not break existing clients, additional testing may be required.

Closes #326

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -52,6 +52,18 @@ def hasher(uaid):
     return uaid
 
 
+def normalize_id(id):
+    if not id:
+        return id
+    if (len(id) == 36 and
+            id[8] == id[13] == id[18] == id[23] == '-'):
+        return id.lower()
+    raw = filter(lambda x: x in '0123456789abcdef', id.lower())
+    if len(raw) != 32:
+        raise ValueError("Invalid UUID")
+    return '-'.join((raw[:8], raw[8:12], raw[12:16], raw[16:20], raw[20:]))
+
+
 def make_rotating_tablename(prefix, delta=0):
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
@@ -261,7 +273,8 @@ class Storage(object):
             cond = "attribute_not_exists(version) or version < :ver"
             conn.put_item(
                 self.table.table_name,
-                item=self.encode(dict(uaid=hasher(uaid), chid=chid,
+                item=self.encode(dict(uaid=hasher(uaid),
+                                      chid=normalize_id(chid),
                                       version=version)),
                 condition_expression=cond,
                 expression_attribute_values={
@@ -281,10 +294,12 @@ class Storage(object):
         """
         try:
             if version:
-                self.table.delete_item(uaid=hasher(uaid), chid=chid,
+                self.table.delete_item(uaid=hasher(uaid),
+                                       chid=normalize_id(chid),
                                        expected={"version__eq": version})
             else:
-                self.table.delete_item(uaid=hasher(uaid), chid=chid)
+                self.table.delete_item(uaid=hasher(uaid),
+                                       chid=normalize_id(chid))
             return True
         except ProvisionedThroughputExceededException:
             self.metrics.increment("error.provisioned.delete_notification")
@@ -312,7 +327,8 @@ class Message(object):
         db_key = self.encode({"uaid": hasher(uaid), "chidmessageid": " "})
         # Generate our update expression
         expr = "ADD chids :channel_id"
-        expr_values = self.encode({":channel_id": set([channel_id])})
+        expr_values = self.encode({":channel_id":
+                                  set([normalize_id(channel_id)])})
         conn.update_item(
             self.table.table_name,
             db_key,
@@ -327,7 +343,8 @@ class Message(object):
         conn = self.table.connection
         db_key = self.encode({"uaid": hasher(uaid), "chidmessageid": " "})
         expr = "DELETE chids :channel_id"
-        expr_values = self.encode({":channel_id": set([channel_id])})
+        expr_values = self.encode({":channel_id":
+                                   set([normalize_id(channel_id)])})
 
         result = conn.update_item(
             self.table.table_name,
@@ -375,7 +392,7 @@ class Message(object):
         with the message id"""
         item = dict(
             uaid=hasher(uaid),
-            chidmessageid="%s:%s" % (channel_id, message_id),
+            chidmessageid="%s:%s" % (normalize_id(channel_id), message_id),
             data=data,
             headers=headers,
             ttl=ttl,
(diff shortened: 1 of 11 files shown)


autopush/db.py
```diff
@@ -52,6 +52,12 @@ def hasher(uaid):
     return uaid
 
 
+def normalize_id(id):
+    if not id:
+        return id
+    return str(uuid.UUID(id))
```
bbangert: This is an expensive operation. Might want to have a regex for the normalized ID, so that re-normalizing a normalized one isn't done, which would be faster if its generally the case that they're already normalized. If it's rare that they're normalized, the extra regex would be slower... so whichever.

autopush/endpoint.py
```diff
@@ -174,7 +175,7 @@ def chid(self):
     @chid.setter
     def chid(self, value):
         """Set the ChannelID and record to _client_info"""
-        self._chid = value
+        self._chid = normalize_id(value)
         self._client_info["channelID"] = value
```
bbangert: We don't want to log the normalized one?

autopush/main.py
```diff
@@ -467,7 +467,8 @@ def endpoint_main(sysargs=None, use_files=True):
 
     # Endpoint HTTP router
     site = cyclone.web.Application([
-        (r"/push/([^\/]+)", EndpointHandler, dict(ap_settings=settings)),
+        (r"/push/(v\d+)?\/?([^\/]+)", EndpointHandler,
```
bbangert: The / should be part of the optional bit. The re module has a non-capturing () which should do just fine rather than the extra ? on the trailing /, ie:
```
r"/push/(?:(v\d)/)?([^/]+)"
```

autopush/router/webpush.py
```diff
@@ -72,7 +73,8 @@ def preflight_check(self, uaid, channel_id):
             self.ap_settings.message_tables[month_table].all_channels,
             uaid=uaid)
 
-        if not exists or channel_id not in chans:
+        if (not exists or channel_id.lower() not
+                in map(lambda x: normalize_id(x), chans)):
```
bbangert: This is not going to be great to do at scale....

bbangert: We're prolly going to have to file a separate issue to replace the uaid lookup with a lookup that does a set membership test in dynamodb, so that we're not pulling the entire list of channels then checking them.

We can ensure the channels are normalized, by updating the migration code to normalize them on monthly rotation.

jrconlin: Yep. We're kinda stuck at the moment because of legacy records. Eventually, they should expire out, but for now, we have to account for them.

autopush/settings.py
```diff
@@ -248,7 +251,55 @@ def update(self, **kwargs):
...
+        if version == 'v1' and len(token) != 32:
+            raise ValueError("Corrupted push token")
+        if version == 'v2':
+            if len(token) != 64:
+                raise ValueError("Corrupted push token")
+            if not public_key:
+                raise ValueError("Invalid key data")
+            if sha256(public_key).digest() != token[32:]:
+                raise ValueError("Key mismatch")
+        return (token[:16].encode('hex'), token[16:32].encode('hex'))
```
bbangert: :open_mouth: :trophy:

autopush/endpoint.py
jrconlin: I'm confused, this will normalize the CHID whenever it's set. that should produce consistent log entries, no?

autopush/main.py
jrconlin: doh, yeah, look ahead is better for this so we don't accidentally eat ids that begin with "v\d..."

bbangert:
Pending a more efficient normalize func, this patch looks great. :clap:

autopush/endpoint.py
bbangert: We log the client_info["channelID"], but this sets that to value, the non-normalized one. It should prolly set it to self._chid which is now normalized.

autopush/db.py
```diff
@@ -52,6 +52,15 @@ def hasher(uaid):
     return uaid
 
 
+def normalize_id(id):
+    if not id:
+        return id
+    if (len(id) == 36 and id[8] == id[13] == id[18] == id[23] == '-'):
+        return id.lower()
+    raw = id.lower().replace('-', '')
```
bbangert: The - are removed so that they can be added again next?

autopush/db.py
```diff
@@ -52,6 +52,15 @@ def hasher(uaid):
     return uaid
 
 
+def normalize_id(id):
+    if not id:
+        return id
+    if (len(id) == 36 and id[8] == id[13] == id[18] == id[23] == '-'):
```
bbangert: Lack of parans here is confusing.

autopush/db.py
jrconlin: This is more a sanity thing. If there are misplaced dashes, strip them before we do any additional formatting. In reality, I should be stripping non-hex characters and checking for string length, but that gets me back precariously to just running the string through uuid.UUID().

bbangert:
Commit message has a semicolon after Closes that shouldn't be there, it seems to screw up multiple tools parsing the commit messages.

kitcambridge:
This looks really good, @jrconlin! :100: Are you planning to add `key` handling to `autopush/websocket.py` in a follow-up?

autopush/websocket.py
```diff
@@ -1000,7 +1000,7 @@ def process_register(self, data):
         self.transport.pauseProducing()
 
         d = self.deferToThread(self.ap_settings.make_endpoint, self.ps.uaid,
```
kitcambridge: Do you have an integration test for this, by chance? I don't want to block this from landing based on that, but it would be nice to see it end-to-end.

bbangert: I agree that a change like this should have an integration test.

autopush/settings.py
```diff
@@ -248,7 +251,68 @@ def update(self, **kwargs):
...
+        :param version: This is the API version of the token.
+        :param public_key: the public key (from Encryption-Key: p256ecdsa=)
+
+        :raises ValueError: In the case of a malformed endpoint.
+
+        :returns: a tuple containing the (UAID, CHID)
+
+        """
+
+        def comp_digest(a, b):
```
kitcambridge: I don't know anything about Python's internals to evaluate this function, but the `&=` and `a[i] == b[i]` seem suspicious. Coda's post recommends a [slightly less complicated form](https://codahale.com/a-lesson-in-timing-attacks/#well-crap-now-what), which is what [Django uses](https://github.com/django/django/blob/master/django/utils/crypto.py#L80-L105). (With a change to compare `ord(x) ^ ord(y)`).

bbangert: https://cryptography.io/en/latest/hazmat/primitives/constant-time/ has a func we should use.

bbangert:
r+

PR #383, opened 2016-03-01
bug: force header values to lowercase underscored values

### LINKED ISSUE #373: Clients expect header keys to be normalized lower case with _ replacing -
We should normalize all outbound headers before sending them out to the client. Optionally we should normalize headers before they're stored.

jrconlin:
The client expects all header values transmitted to be lower case with
underbars. (e.g."Encryption-Key" would become "encryption-key") The
current code enforces this rule for the header data that is included
as part of the WebPush update.

Closes #373 

@bbangert

### DIFF
--- a/autopush/router/webpush.py
+++ b/autopush/router/webpush.py
@@ -53,8 +53,9 @@ class WebPushRouter(SimpleRouter):
         # they're present to avoid empty strings.
         for name in ["encryption-key", "crypto-key"]:
             if name in headers:
-                # Normalize hyphens in header names.
-                data[name.replace("-", "_")] = headers[name]
+                # NOTE: The client code expects all header keys to be lower
+                # case and s/-/_/.
+                data[name.lower().replace("-", "_")] = headers[name]
         return data
 
     @inlineCallbacks


autopush/router/webpush.py
```diff
@@ -53,8 +53,9 @@ def _crypto_headers(self, notification):
         # they're present to avoid empty strings.
         for name in ["encryption-key", "crypto-key"]:
             if name in headers:
-                # Normalize hyphens in header names.
-                data[name.replace("-", "_")] = headers[name]
+                # NOTE: The client code expects all header keys to be lower
+                # case and s/-/_/.
+                data[name.lower().replace("-", "_")] = headers[name]
```
kitcambridge: This just makes it more explicit, right? It looks like the names are already lowercase in the list on line 54.

jrconlin: Yep. Just forcing the matter and adding the comment to prevent future potential screw ups.

kitcambridge: :+1:

bbangert:
r+

PR #384, opened 2016-03-04
bug: don't return 503 for disconnected user

### LINKED ISSUE #378: Reduce 503 occasions
When a user leaves, and a new notification comes in, we clear the node and return a 503 to try again. That can push a decent amount of work onto the app-server, it would likely be better to actually continue the logic in endpoint, such that:

- clear the node entry
- save the notification
- new node lookup
- tell node to fetch messages if there

So essentially all we'd do is remove the raise exception, and let the message be saved immediately. That would reduce this 503 path, without much of a db hit change since most of the time the user has left and we are going to be storing the message for awhile.

jrconlin:
When a user is not connected, store the message and do not return a 503.

Closes #378 

@bbangert r?

### DIFF
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -97,6 +97,7 @@ class SimpleRouter(object):
         #   - Error (Client gone, node gone/dead): Clear node entry for user
         #       - Both: Done, return 503
         if node_id:
+            result = None
             try:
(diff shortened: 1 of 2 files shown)


bbangert:
Nice, I like this behavior so much better than having the remote node retry.

r+

PR #386, opened 2016-03-04
bug: capture ValueError for empty notifications arrays

### LINKED ISSUE #385: Capture ValueError when client may leave
Per https://sentry.prod.mozaws.net/operations/autopush-prod/group/235881/

An AttributeError is caught, but not the ValueError that can occur. The ValueError should be caught as well.

jrconlin:
Empty pending notification arrays may be returned. Removing a value can
trigger a ValueError that can be safely ignored.

Closes #385 

@bbangert r?


bbangert:
r+

PR #388, opened 2016-03-04
fix: send raven calls to event loop

### LINKED ISSUE #387: Send raven call to event loop
raven should not run during logging exceptions as twisted may trigger those during garbage collection. Instead the call should be scheduled with callFromThread.

bbangert:
Deadlocks due to raven reporting can occur when exceptions
are logged during GC, as twisted may do. As such, the raven
call should be sent to the event loop so that twisted's
__del__ call can return uninterrupted.

Closes #387 

@jrconlin r?

### DIFF
--- a/autopush/logging.py
+++ b/autopush/logging.py
@@ -17,6 +17,7 @@ import sys
 import raven
 from eliot import (add_destination, fields,
                    Logger, MessageType)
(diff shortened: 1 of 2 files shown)


jrconlin:
r+ pending travis.

And TIL about reactor.callFromThread()

PR #391, opened 2016-03-07
bug: limit valid months to acceptable range

### LINKED ISSUE #350: Make message dict a defaultdict
Rather than adding a Message object as the month switches over, it should be a default-dict so that if there's greater clock skew in production, it won't matter.

The only twisted looping task to be to update the current_month var.

jrconlin:
A user that tries to connect from a period longer than we currently
allow for could cause a "KeyError" on the server. Instead, we should
require that the user use a new UAID, which shoud cause the client to
re-register older connections.

Closes #350 

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -58,10 +58,11 @@ def normalize_id(id):
     return '-'.join((raw[:8], raw[8:12], raw[12:16], raw[16:20], raw[20:]))
 
 
-def make_rotating_tablename(prefix, delta=0):
+def make_rotating_tablename(prefix, delta=0, date=None):
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
-    date = get_month(delta=delta)
+    if not date:
+        date = get_month(delta=delta)
     return "{}_{}_{}".format(prefix, date.year, date.month)
 
 
@@ -77,11 +78,11 @@ def create_rotating_message_table(prefix="message", read_throughput=5,
                         )
 
 
-def get_rotating_message_table(prefix="message", delta=0):
+def get_rotating_message_table(prefix="message", delta=0, date=None):
     """Gets the message table for the current month."""
     db = DynamoDBConnection()
     dblist = db.list_tables()["TableNames"]
-    tablename = make_rotating_tablename(prefix, delta)
+    tablename = make_rotating_tablename(prefix, delta, date)
     if tablename not in dblist:
         return create_rotating_message_table(prefix=prefix, delta=delta)
     else:
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -19,11 +19,10 @@ from autopush.db import (
     get_router_table,
     get_storage_table,
     get_rotating_message_table,
-    make_rotating_tablename,
     preflight_check,
     Storage,
     Router,
-    Message
+    Message,
 )
 from autopush.exceptions import InvalidTokenException
 from autopush.metrics import (
@@ -163,9 +162,13 @@ class AutopushSettings(object):
         self.router = Router(self.router_table, self.metrics)
 
         # Used to determine whether a connection is out of date with current
-        # db objects
-        self.current_msg_month = make_rotating_tablename(self._message_prefix)
-        self.current_month = datetime.date.today().month
+        # db objects. There are three noteworty cases:
+        # 1 "Last Month" the table requires a rollover.
+        # 2 "This Month" the most common case.
+        # 3 "Next Month" where the system will soon be rolling over, but with
+        #   timing, some nodes may roll over sooner. Ensuring the next month's
+        #   table is present before the switchover is the main reason for this,
+        #   just in case some nodes do switch sooner.
         self.create_initial_message_tables()
 
         # Run preflight check
@@ -204,18 +207,29 @@ class AutopushSettings(object):
         """Setter to set the current message table"""
         self.message_tables[self.current_msg_month] = value
 
+    def _tomorrow(self):
+        return datetime.date.today() + datetime.timedelta(days=1)
+
     def create_initial_message_tables(self):
         """Initializes a dict of the initial rotating messages tables.
 
-        An entry for last months table, and an entry for this months table.
+        An entry for last months table, an entry for this months table,
+        an entry for tomorrow, if tomorrow is a new month.
 
         """
+        today = datetime.date.today()
         last_month = get_rotating_message_table(self._message_prefix, -1)
         this_month = get_rotating_message_table(self._message_prefix)
+        self.current_month = today.month
+        self.current_msg_month = this_month.table_name
         self.message_tables = {
             last_month.table_name: Message(last_month, self.metrics),
-            this_month.table_name: Message(this_month, self.metrics),
+            this_month.table_name: Message(this_month, self.metrics)
         }
+        if self._tomorrow().month != today.month:
+            next_month = get_rotating_message_table(delta=1)
+            self.message_tables[next_month.table_name] = Message(
+                next_month, self.metrics)
 
     @inlineCallbacks
     def update_rotating_tables(self):
@@ -227,6 +241,15 @@ class AutopushSettings(object):
 
         """
         today = datetime.date.today()
+        tomorrow = self._tomorrow()
+        if ((tomorrow.month != today.month) and
+                sorted(self.message_tables.keys())[-1] !=
+                tomorrow.month):
+            next_month = get_rotating_message_table(
+                self._message_prefix, 0, tomorrow)
+            self.message_tables[next_month.table_name] = Message(
+                next_month, self.metrics)
+
         if today.month == self.current_month:
             # No change in month, we're fine.
             returnValue(False)
@@ -241,7 +264,6 @@ class AutopushSettings(object):
         self.current_msg_month = message_table.table_name
         self.message_tables[self.current_msg_month] = \
             Message(message_table, self.metrics)
-
         returnValue(True)
 
     def update(self, **kwargs):
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -776,9 +776,14 @@ class PushServerProtocol(WebSocketServerProtocol, policies.TimeoutMixin):
         self.transport.pauseProducing()
         # Check for table rotation
         cur_month = previous.get("current_month")
+        # Previous month user or new user, flag for message rotation and
+        # set the message_month to the router month
         if cur_month != self.ps.message_month:
-            # Previous month user or new user, flag for message rotation and
-            # set the message_month to the router month
+            if cur_month not in self.ps.settings.message_tables:
+                # This UAID has expired. Force client to reregister.
+                self.ps.uaid = uuid.uuid4().hex
+                self._finish_webpush_hello()
+                return
             self.ps.message_month = cur_month
             self.ps.rotate_message_table = True
 
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -70,6 +70,14 @@ class DbCheckTestCase(unittest.TestCase):
             'd8f614d06cdd592cb8470f31177c8331a')
         db.key_hash = ""
 
+    def test_normalize_id(self):
+        import autopush.db as db
+        abnormal = "deadbeef00000000decafbad00000000"
+        normal = "deadbeef-0000-0000-deca-fbad00000000"
+        eq_(db.normalize_id(abnormal), normal)
+        self.assertRaises(ValueError, db.normalize_id, "invalid")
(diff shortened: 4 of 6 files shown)


autopush/settings.py
```diff
@@ -160,6 +160,9 @@ def __init__(self,
         # db objects
         self.current_msg_month = make_rotating_tablename(self._message_prefix)
         self.current_month = datetime.date.today().month
+        # a range of months that we consider a message to be "valid"
+        self.valid_months = [self.current_msg_month,
+                             make_rotating_tablename(self._message_prefix, -1)]
```
bbangert: Technically 2 months in the past are valid. Otherwise our max message retention is only ~ 27 days. Also, if this value is not auto-updated, then the valid months will not be the current month when the day flips over to the new month.

bbangert:
I think maybe the bug wasn't quite clear why a defaultdict was better.

When the new month flips over, some nodes update their dict before others. Meaning some nodes know the current month is now 1 newer than other nodes. The other nodes therefore get keyerror when a user reconnects that was on one of already flipped-over nodes.

Technically the valid months are at any given check are:

- Next month, if we're within a day of the new month
- This month
- The past 2 months

At the same time, we can't create these all up-front, because as time progresses, all the calculations on "valid" change as well when we hit the new month. It's also not good to create next months table too far from the new month since it should exist in AWS first.

jrconlin:
Ok, so if I make PushState.settings.message_tables a default_dict, I'll add the logic to __missing__ handle those cases. How should __missing__ handle an entry that is beyond the limit of what we consider valid? (In the issue, we flag a value that is set to be outside of the current or prior month.) I'm presuming that we would have to wrap all property references with a try: handler so that we can detect the KeyError and reset the UAID or some other action, correct?

bbangert:
I think doing the check on websocket connect, and on the endpoint for messages should be fine so that the defaultdict doesn't need to do the logic. So if we get a message for a user that has too old of a message month, then we send a delete, and treat it as a nonexistent user.

And in websocket, we check if its invalid and assign a new one.

bbangert:
Typo in commit message, "which shoud cause"

autopush/settings.py
```diff
@@ -206,15 +210,28 @@ def message(self, value):
...
+        self.message_tables[last_month.table_name] = Message(last_month,
+                                                             self.metrics)
+        self.message_tables[this_month.table_name] = Message(this_month,
+                                                             self.metrics)
+        # Due to issues around mocking "datetime" and desire to
+        # reduce excessive hotpath function calls, we have to skip this
+        # line
+        if tomorrow.table_name != this_month.table_name:  # pragma: nocover
+            self.message_tables[tomorrow.table_name] = Message(
+                tomorrow, self.metrics)
```
bbangert: Shouldn't all this tomorrow logic go into the update_rotating_tables function below?

autopush/websocket.py
```diff
@@ -773,9 +773,14 @@ def _check_message_table_rotation(self, previous):
         self.transport.pauseProducing()
         # Check for table rotation
         cur_month = previous.get("current_month")
+        # Previous month user or new user, flag for message rotation and
+        # set the message_month to the router month
         if cur_month != self.ps.message_month:
-            # Previous month user or new user, flag for message rotation and
-            # set the message_month to the router month
+            if cur_month not in self.ps.settings.message_tables.keys():
```
bbangert: The 'in' test for Python dict by default checks the keys, calling .keys() on a dict isn't needed.

bbangert: Also, why not move this if block up out of its nesting, then make the block its nested in a "elif".

bbangert:
I don't quite understand why any change was needed in get_rotating_table_month. If the looping task of update_rotating_tables determines that one day from now is a new month and it hasn't already added the table for that month, it could call get_rotating_table_month with a delta of 1 and add that.

bbangert:
If there's checks to prevent key access to the messages dict if the key doesn't exist, why does it have a defaultdict at all? Alternatively, if something does manage to get access such that the defaultdict is triggered, what's the guarantee that only the single dummy function on it will be called?

autopush/websocket.py
jrconlin: I figured that the most common case would be that the previous "current_month" would equal the current month (as most clients would reconnect in the same month) thus saving the lookup for multiple values. Less likely would be: The connection is from last month, the connection is from a month that expired, the connection is one that was already forwarded (the tomorrow is another month addendum). If this is not the case, or if it's the same cost or cheaper to compare a value against a list than to compare against a single, I'm fine with creating an elif.

autopush/settings.py
jrconlin: It should be in both, in cases where the app is started at the end of the month. thanks!

jrconlin:
To be honest, the defaultdict isn't needed, although it was mentioned in the bug and I was encouraged to use a defaultdict as part of the solution. I filter out dates before calling the message property which triggered the original KeyError, so it shouldn't be needed. The only function call that existed in the original error path was fetch_messages() (called from websocket.process_notifications() as part of _finish_webpush_hello() ), and that one pulls from "current_month", which should be rotated in. 

I'm fine removing it if it's redundant and paranoid.

bbangert:
Yea, I'd prefer losing the extra defaultdict code, mainly cause I'm not sure about what code paths it introduces that could cause new bugs or oddities.

autopush/settings.py
```diff
@@ -206,15 +210,28 @@ def message(self, value):
...
+        self.message_tables[this_month.table_name] = Message(this_month,
+                                                             self.metrics)
+        # Due to issues around mocking "datetime" and desire to
+        # reduce excessive hotpath function calls, we have to skip this
+        # line
+        tomorrow = get_rotating_message_table(
+            self._message_prefix, 0, today + datetime.timedelta(days=1))
+        if tomorrow.table_name != this_month.table_name:  # pragma: nocover
+            self.message_tables[tomorrow.table_name] = Message(
+                tomorrow, self.metrics)
```
bbangert: Why is this in two places?

autopush/settings.py
```diff
@@ -206,15 +210,28 @@ def message(self, value):
...
+        self.message_tables[last_month.table_name] = Message(last_month,
+                                                             self.metrics)
+        self.message_tables[this_month.table_name] = Message(this_month,
+                                                             self.metrics)
+        # Due to issues around mocking "datetime" and desire to
+        # reduce excessive hotpath function calls, we have to skip this
+        # line
+        tomorrow = get_rotating_message_table(
+            self._message_prefix, 0, today + datetime.timedelta(days=1))
+        if tomorrow.table_name != this_month.table_name:  # pragma: nocover
```
bbangert: This definitely needs coverage, or it should be refactored into another function that is covered, that this and the update_rotating_tables function call.

autopush/websocket.py
bbangert: Ah, good point, I'm fine leaving it as is. The keys() bit should still be dropped.

autopush/db.py
```diff
@@ -49,6 +49,8 @@ def hasher(uaid):
 
 
 def normalize_id(id):
+    if id is None:
+        return None
```
bbangert: There was no test for this before, so I had removed it. Is there any case where this will be called on a None or empty string?

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
         """Setter to set the current message table"""
         self.message_tables[self.current_msg_month] = value
 
+    def add_tomorrow(self, today, this_month_table):
+        """check to see if tomorrow is a new month, in case we're a node that
+        has a slow clock."""
+        tomorrow = get_rotating_message_table(
+            self._message_prefix, 0, today + datetime.timedelta(days=1))
```
bbangert: We could avoid adding code to the get_rotating_message_table by just doing this check here.

```python
if (today + datetime.timedelta(days+1).month != today.month:
    next_month = get_rotating_message_table(delta=1)
    self.message_tables[next_month.table_name] = Message(next_month, self.metrics)
```

This also won't require passing this months table.

Why is it better to increase the complexity of the get_rotating_message_table function instead?

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
...
+        if tomorrow.table_name != this_month_table:
+            self.message_tables[tomorrow.table_name] = Message(
```
bbangert: update_rotating_tables is called once a minute. We should check to ensure tomorrow's table isn't already in there before making it again.

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
         """Setter to set the current message table"""
         self.message_tables[self.current_msg_month] = value
 
+    def add_tomorrow(self, today, this_month_table):
+        """check to see if tomorrow is a new month, in case we're a node that
+        has a slow clock."""
```
bbangert: Possible idea for a quick check to see if next months table was already added, since we know they sort ok...
```python
if sorted(self.message_tables.keys())[-1] != self.current_msg_month:
    return
```

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
         """Setter to set the current message table"""
         self.message_tables[self.current_msg_month] = value
 
+    def add_tomorrow(self, today, this_month_table):
```
bbangert: nit: so far we have last_month and this_month, maybe next_month and calling this add_next_month?

autopush/settings.py
```diff
@@ -163,9 +162,12 @@ def __init__(self,
...
         # Used to determine whether a connection is out of date with current
-        # db objects
-        self.current_msg_month = make_rotating_tablename(self._message_prefix)
-        self.current_month = datetime.date.today().month
+        # db objects. There are three noteworty cases:
+        # 1 "Last Month" the table requires a rollover.
+        # 2 "This Month" the most common case.
+        # 3 "Tomorrow" in the case where the system is actively rolling over
+        #   and one node is not yet using the latest clock, no rollover
+        #   required.
```
bbangert: "Next Month", where the system will soon be rolling over, but with timing, some nodes may roll over sooner. Ensuring the next month's table is present before the switchover is the main reason for this, just in case some nodes do switch sooner.

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
...
+        if tomorrow.table_name != this_month_table:
+            self.message_tables[tomorrow.table_name] = Message(
+                tomorrow, self.metrics)
+
     def create_initial_message_tables(self):
         """Initializes a dict of the initial rotating messages tables.
 
-        An entry for last months table, and an entry for this months table.
+        An entry for last months table, an entry for this months table,
+        and an entry for tomorrow, if it starts a new month.
```
bbangert: "an entry for next months table, if tomorrow is a new month."

autopush/settings.py
```diff
@@ -204,18 +206,35 @@ def message(self, value):
...
+        self.current_month = today.month
+        self.current_msg_month = this_month.table_name
         self.message_tables = {
             last_month.table_name: Message(last_month, self.metrics),
-            this_month.table_name: Message(this_month, self.metrics),
+            this_month.table_name: Message(this_month, self.metrics)
         }
+        # Due to issues around mocking "datetime" and desire to
+        # reduce excessive hotpath function calls, we have to skip this
+        # line
```
bbangert: What was the issue? I'm sure I could help get this fixed.

autopush/websocket.py
bbangert: :eyes: ``.keys()``

autopush/settings.py
```diff
@@ -227,6 +256,15 @@ def update_rotating_tables(self):
 
         """
         today = datetime.date.today()
+        tomorrow = self._tomorrow()
+        if ((tomorrow.month != today.month) and
+                sorted(self.message_tables.keys())[-1] !=
+                tomorrow.month):
+            next_month = get_rotating_message_table(
+                self._message_prefix, 0, tomorrow)
```
bbangert: Why not remove the date passing addition to get_rotating_message_table and just pass a delta of 1 in?

autopush/settings.py
```diff
@@ -204,18 +207,44 @@ def message(self, value):
...
+        tomorrow = (today + datetime.timedelta(days=1))
+        if tomorrow.month == today.month:
+            return
+        if sorted(self.message_tables.keys())[-1] != self.current_msg_month:
+            return
+        next_month = get_rotating_message_table(
+            self._message_prefix, 0, tomorrow)
+        self.message_tables[next_month.table_name] = Message(
+            tomorrow, self.metrics)
+    """
```
bbangert: Remove this?

bbangert:
r+ pending removal of the commented out block and travis

PR #396, opened 2016-03-10
Bug/issue 395

### LINKED ISSUE #395: Fix bug with version defaulting to 0
cyclone defaults to None being passed in for missing params, this should be accounted for in the handlers.

bbangert:
Default the api_ver to v0 for the message endpoint. Previously a
value of None could pass in causing earlier endpoints to be incorrectly
parsed out. parse_endpoint now throws InvalidTokenExceptions to avoid
being caught on accident as a ValueError.

Closes #395 

@jrconlin r?


jrconlin:
r+

PR #398, opened 2016-03-10
Bug/fix timing test

bbangert:
Occasionally a timing issue will prevent the last_connect test
from updating in time. Wait a more reasonable duration for the
database to update.

@jrconlin r?

Note: The other PR should be applied first.


jrconlin:
R+

PR #402, opened 2016-03-11
fix: clear only the node_id in the router record

### LINKED ISSUE #401: Clear node should only clear the node id
clear_node incorrectly clears the entire record. It should only clear out the node_id field.

bbangert:
Closes #401 

@jrconlin r?


jrconlin:
r+

jrconlin:
somebody was impatient.

PR #403, opened 2016-03-12
bug: Clear corrupted router records

### LINKED ISSUE #400: Clear corrupted router records
When a record is found only containing the uaid, it should be cleared entirely so it can be recreated properly by the client. We should then return a 410 for the notification.

jrconlin:
Records containing only a UAID are corrupt. Drop these records and
return a 410 to the server.

Closes #400 

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -539,7 +539,12 @@ class Router(object):
 
         """
         try:
-            return self.table.get_item(consistent=True, uaid=hasher(uaid))
+            item = self.table.get_item(consistent=True, uaid=hasher(uaid))
+            if item.keys() == ['uaid']:
+                # Incomplete record, drop it.
+                self.drop_user(uaid)
+                raise ItemNotFound("uaid not found")
+            return item
         except ProvisionedThroughputExceededException:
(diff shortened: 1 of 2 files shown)


autopush/db.py
```diff
@@ -539,7 +539,13 @@ def get_uaid(self, uaid):
 
         """
         try:
-            return self.table.get_item(consistent=True, uaid=hasher(uaid))
+            item = self.table.get_item(consistent=True, uaid=hasher(uaid))
+            if item.keys() == ['uaid']:
+                import pdb; pdb.set_trace()
```
bbangert: This shouldn't be in here.

autopush/db.py
```diff
@@ -539,7 +539,13 @@ def get_uaid(self, uaid):
...
+                # Incomplete record, drop it.
+                self.drop_user(uaid)
```
bbangert: :+1: I like dropping this here, while it's already in the other thread.

bbangert:
r+ pending drop of pdb

autopush/db.py
jrconlin: crap! must have gotten in the patch thanks!

PR #407, opened 2016-03-12
feat: validate v0 tokens more thoroughly

### LINKED ISSUE #406: Validate v0 token is correct
Some appservers seem to be sending v1 or v2 tokens in place of a v0 token. Most likely because they're storing only the token at the end and not the rest of the URL. We should consider these invalid tokens instead of throwing an exception on setting the chid from normalize_id.

bbangert:
Changes v0 token validation from simple check containing a : into
regex expecting valid uuid characters on either side of the : as
well.

Closes #406 

@jrconlin r?

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -1,5 +1,6 @@
 """Autopush Settings Object and Setup"""
 import datetime
+import re
 import socket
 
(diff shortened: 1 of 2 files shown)


jrconlin:
R+

PR #411, opened 2016-03-16
bug: decode and process crypto-key header correctly

### LINKED ISSUE #410: bug: Fix header parsing for Autopush
http://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-00#section-5.3
notes that Crypto-Key components have the following 
keyid="id";tag="value",keyid="id";tag="value";...

Since headers tend to have specific relationships with the data that they are encoding, parse_header needs to comply with the Crypto-Key format, and all references to the header output will need to be refactored.

jrconlin:
The Crypto-Key header has a different format than what was first
understood. A valid Crypto-Key header may look like `keyid="foo";
key1="data",key2="data"`. This patch specialized Crypto-Key parsing
to a class.

Closes #410

@bbangert r?

*Note*: I'd love to have a full integration test where we can check that the VAPID data was recorded correctly, as well as the data passed through. test_integration's TestWebpush.test_basic_delivery_with_vapid contains most of what would be required, but I don't know how to get the _client_info to check that the JWT content matches what we're expecting.

### DIFF
--- a/autopush/crypto_key.py
+++ b/autopush/crypto_key.py
@@ -1,0 +1,93 @@
+"""Crypto-Key header parser and manager"""
+
+
+class CryptoKeyException(Exception):
+    """Invalid CryptoKey"""
+
+
+class CryptoKey(object):
+    """Parse the Crypto-Key header per
+http://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-00#section-4
+
+    The Crypto-Key header is a data store that has it's own set of rules,
+    This class manages access to the Crypto-Key data. There are two ways that
+    data in the Crypto-Key can be used, one by the key-id (an optional
+    identifier that can be used to associate key data to other fields) and
+    the data label name.
+
+    There are a few additional functions that may be implemented. For
+    instance, removing a Crypto-Key component that may not need to be
+    passed on.
+
+    """
+    _values = []
+
+    def __init__(self, header):
+        """Parse the Crypto-Key header
+
+        :param header: Header content string.
+
+        """
+        chunks = header.split(",")
+        if self._values:
+            self._values = []
+        for chunk in chunks:
+            bits = chunk.split(";")
+            hash = {}
+            for bit in bits:
+                try:
+                    (key, value) = bit.split("=", 1)
+                except ValueError:
+                    raise CryptoKeyException("Invalid Crypto Key value")
+                hash[key.strip()] = value.strip(' "')
+            self._values.append(hash)
+
+    def get_keyid(self, keyid):
+        """Return the Crypto-Key hash referred to by a given keyid.
+
+        For example, for a CryptoKey specified as:
+        Crypto-Key: keyid="apple";foo="fruit",keyid="gorp";bar="snake"
+
+        get_keyid("apple") would return a hash of
+        {"keyid": "apple", "foo":"fruit"}
+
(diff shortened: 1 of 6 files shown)


autopush/crypto_key.py
```diff
@@ -0,0 +1,82 @@
...
+class CryptoKeyException(Exception):
+    """Invalid CryptoKey"""
+
+
+class CryptoKey(object):
+    """Parse the Crypto-Key header per
+    http://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-00#section-4
+
+    """
+    _values_ = []
```
bbangert: This is very unorthodox Python. Double underscores on both sides signify built-in override-able methods, I've never seen or heard of single underscores on both sides. Single leading underscore is generally a private var, and double leading underscore is a special private class var that gets Python class mangling applied. What's the rationale for this new approach?

autopush/crypto_key.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_label(self, label):
+        """Return the Crypto-Key value referred to by a given label.
+
+        For example, for a CryptoKey specified as:
+        Crypto-Key: keyid="apple";foo="fruit",keyid="gorp";bar="snake"
+
+        get_label("foo")
+        would return a value of "apple"
+
+        **NOTE** This presumes that "label" is unique. Otherwise it will
```
bbangert: Since these are generally formatted with Sphinx using RST, this should be a ..note directive.

autopush/settings.py
```diff
@@ -280,26 +281,34 @@ def make_endpoint(self, uaid, chid, key=None):
...
-        :returns: a tuple containing the (UAID, CHID)
+        :returns: a dict containing (uaid=UAID, chid=CHID, public_key=KEY)
 
         """
 
         token = self.fernet.decrypt(token.encode('utf8'))
+        public_key = None
+        if ckey_header:
+            try:
+                public_key = CryptoKey(ckey_header).get_label('p256ecdsa')
```
bbangert: Why do we make a CryptoKey class object, instantiate it, then call a method on it, just to ditch the object? That's quite a few function calls, object creations, and a method lookup just to run essentially a single function since the object is dumped.

bbangert:
I don't quite understand what I'm missing here. A CryptoKey object with a method (get_keyid) that is tested but never used by actual code. An object that is dumped immediately after creation.

Was there some other portion of this patch that isn't in here? Is the other method going to be used later, but just didn't make it into this PR?

jrconlin:
The Crypto-Key header is a data store that has it's own set of rules, thus the class to manage access to the Crypto-Key data. There are two ways that data in the Crypto-Key can (and are) used, one by the key-id (an optional identifier that can be used to associate key data to other fields) and the data label name. There has been some discussion regarding the server altering the content of the crypto-key. For example, to remove the p256ecdsa values which are currently only used for VAPID authorization and do not currently need to be stored or passed on to the client. Since the crypto-key is part of several evolving standards, I consider this class to be a building block that may become a future library. This is why there is also a separate test case for the class.

bbangert:
Ok, that makes sense, this is for future changes as well on the CryptoKey parser. Thanks.

bbangert:
r+ pending travis

jrconlin:
You raise a good point about adding that summary as a comment to the class. Adding that too.

PR #413, opened 2016-03-17
feat: bump autobahn/twisted reqs to 0.13/16.0 for utf8 fix

### LINKED ISSUE #351: Update autobahn requirement
autobahn has a new release that has the utf-8 fix we've been blocked on for quite awhile (Issue #180). We should update our requirement to use it.

bbangert:
Closes #351

@jrconlin r?

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -383,8 +383,6 @@ def connection_main(sysargs=None, use_files=True):
     proto = "wss" if args.ssl_key else "ws"
     factory = WebSocketServerFactory(
         "%s://%s:%s/" % (proto, args.hostname, args.port),
-        debug=args.debug,
-        debugCodePaths=args.debug,
     )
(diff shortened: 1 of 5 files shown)


autopush/main.py
```diff
@@ -383,8 +383,6 @@ def connection_main(sysargs=None, use_files=True):
     proto = "wss" if args.ssl_key else "ws"
     factory = WebSocketServerFactory(
         "%s://%s:%s/" % (proto, args.hostname, args.port),
-        debug=args.debug,
-        debugCodePaths=args.debug,
```
jrconlin: are these no longer valid or useful? (Don't know if they might be useful again in the future.)

bbangert: They're using the general Python logging flags now instead, so setting these is not supported anymore.

jrconlin:
r+

PR #418, opened 2016-03-18
fix: fix dockerfile for automated builds

### LINKED ISSUE #414: Restore dockerization of autopush
Autopush has a Dockerfile that used to work, but hasn't been maintained. It should be updated with automatic trusted Docker repo builds.

bbangert:
Dockerfile and Makefile were both broken. This fixes the Dockerfile
and uses the latest PyPy 5.0 for building the container.

Closes #414 

@jrconlin r?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@ SHELL := /bin/sh
 APPNAME = autopush
 DEPS =
 HERE = $(shell pwd)
-PTYPE=python
-BIN = $(HERE)/bin
+PTYPE=pypy
+BIN = $(HERE)/pypy/bin
 VIRTUALENV = virtualenv
 TESTS = $(APPNAME)/tests
 PYTHON = $(BIN)/$(PTYPE)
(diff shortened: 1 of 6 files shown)


bbangert:
Looks like the other Travis CI test is never going to run, for unknown reason it won't let us restart it.

Makefile
```diff
@@ -2,8 +2,8 @@ SHELL := /bin/sh
 APPNAME = autopush
 DEPS =
 HERE = $(shell pwd)
-PTYPE=python
-BIN = $(HERE)/bin
+PTYPE=pypy
```
kitcambridge: I wonder if we should eventually add a task that sets up a CPython virtualenv and symlinks `python` to `pypy` for local development.

Makefile
```diff
@@ -18,7 +18,7 @@ BUILD_DIRS = bin build deps include lib lib64 lib_pypy lib-python\
 
 all:	build
 
-travis: $(HERE)/ddb
+travis: $(HERE)/ddb $(BIN)/tox
```
kitcambridge: Hmm...don't we still need to install `$(BIN)/tox` at this point?

kitcambridge:
Please ignore the previous comment about tox; I didn't see that we already have a task for `$(BIN)/tox`. r+

PR #425, opened 2016-03-23
Test a different order of keys in Crypto-Key

marco-c:
A different order of keys is making the push service fail for me with 500 (Internal Server Error), so I wanted to try this test to make sure the CryptoKey parsing is working correctly.

I haven't run the tests locally, I'm not sure how to build and run tests.


marco-c:
It's passing, so the problem is somewhere else.
I think adding the test is still worth it, so I'm going to leave the PR open.

jrconlin:
marco: I think the problem was more with the difference between production and dev. This may have been resolved by #411. 

Still, adding the test is valuable, so going to let this PR stand

jrconlin:
r+

PR #427, opened 2016-03-23
bug: enforce fail.value.message to string

jrconlin:
Two things: 
1) fail.value.message can sometimes return a class, which will cause the logging to break.
2) Be a lot more rigorous about padding tests. There are some values that all the padding in the world will fail to decode, but we should be certain that only those actually fail.

@bbangert r?


bbangert:
r+

PR #430, opened 2016-03-28
fix: update cffi dep and ensure test reqs matches reqs

bbangert:
A cffi dependency pinning was missed because test-requirements was
not matching requirements.txt. This caused build issues for ops.


jrconlin:
r+

PR #439, opened 2016-04-06
bug: flatten JWT for logging.

jrconlin:
Logging does not handle object exports well. Flattening JWT by prefixing
"jwt_" to the names of values to prevent potential conflicts

@bbangert r?.


bbangert:
r+

PR #441, opened 2016-04-15
refactor: Remove duplicate validation logic in the GCM and APNs routers

kitcambridge:
I cleaned this up a bit, since we already validate the headers and encode the payload in `EndpointHandler`. While I was here, I also updated the APNs router to forward the crypto headers (even though we don't have any active consumers at the moment).

@jrconlin r?

### DIFF
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -65,12 +65,24 @@ class APNSRouter(object):
     def _route(self, notification, router_data):
         """Blocking APNS call to route the notification"""
         token = router_data["token"]
+        custom = {
+          "Chid": notification.channel_id,
+          "Ver": notification.version,
+        }
+        if notification.data:
+            custom["Msg"] = notification.data
+            custom["Con"] = notification.headers["content-encoding"]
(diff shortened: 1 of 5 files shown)


jrconlin:
Heh, well, we may have to completely redo APNs stuff to use the http2 interfaces, when the iOS client wants to get push messages.

autopush/router/apnsrouter.py
```diff
@@ -65,12 +65,25 @@ def route_notification(self, notification, uaid_data):
     def _route(self, notification, router_data):
         """Blocking APNS call to route the notification"""
         token = router_data["token"]
+        custom = {
+          "Chid": notification.channel_id,
+          "Ver": notification.version,
+        }
+        print("KITCAM: NOTIF REPR %r" % notification.data)
```
jrconlin: debug?

jrconlin:
r+ pending removal of the debugging print statement.

autopush/router/apnsrouter.py
kitcambridge: Oops. Thanks!

PR #442, opened 2016-04-16
fix: Canonicalize Base64 URL-encoded values per RFC 7515

kitcambridge:
For restricted subscriptions with an app server key, we include the SHA-256 hash of the *encoded* form in the endpoint URL. This can be a problem if one side uses padded input, and the other uses unpadded. Another complication is, while RFC 7515 forbids padding, Python's Base64 decoder requires it. As we discussed on IRC a few days ago, keeping track of these different forms is error-prone.

This PR adds two helper functions to encode and decode unpadded Base64. It also normalizes VAPID keys to the unpadded form, regardless of whether the sender included padding.

Finally, I noticed `self.assertRaises` can silence failures, so I switched to using Nose's `assert_raises` instead.

@bbangert r?

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -36,7 +36,7 @@ from autopush.router import (
     SimpleRouter,
     WebPushRouter,
 )
-from autopush.utils import canonical_url, resolve_ip
+from autopush.utils import canonical_url, resolve_ip, base64url_decode
 from autopush.senderids import SENDERID_EXPRY, DEFAULT_BUCKET
 from autopush.crypto_key import (CryptoKey, CryptoKeyException)
 
@@ -290,7 +290,7 @@ class AutopushSettings(object):
 
         :param uaid: User Agent Identifier
         :param chid: Channel or Subscription ID
-        :param key: Optional provided Public Key
+        :param key: Optional Base64 URL-encoded application server key
         :returns: Push endpoint
 
         """
@@ -301,7 +301,9 @@ class AutopushSettings(object):
         if key is None:
             return root + 'v1/' + self.fernet.encrypt(base).strip('=')
 
-        return root + 'v2/' + self.fernet.encrypt(base + sha256(key).digest())
+        raw_key = base64url_decode(key.encode('utf8'))
+        return root + 'v2/' + self.fernet.encrypt(base +
+                                                  sha256(raw_key).digest())
 
     def parse_endpoint(self, token, version="v0", ckey_header=None):
         """Parse an endpoint into component elements of UAID, CHID and optional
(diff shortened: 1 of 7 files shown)


autopush/tests/test_endpoint.py
```diff
@@ -1194,16 +1194,12 @@ def _assert_error_response(self, result):
...
-        eq_(utils.fix_padding("ab"), "ab==")
-        base64.urlsafe_b64decode(utils.fix_padding("ab"))
-        eq_(utils.fix_padding("abc"), "abc=")
-        base64.urlsafe_b64decode(utils.fix_padding("abc"))
-        eq_(utils.fix_padding("abcd"), "abcd")
-        base64.urlsafe_b64decode(utils.fix_padding("abcd"))
+        assert_raises(TypeError,
+                      utils.base64url_decode,
+                      "a===")
+        eq_(utils.base64url_decode("ab=="), "\x69")
```
jrconlin: https://twitter.com/kennwhite/status/721035633657909248

jrconlin:
while I get the efficiency of using raw values to compare against, I'm not a big fan of "mystery values". (thus the tweet I linked to.)

I see no fault in the code, and will defer to @bbangert for the value of `assert_raises`.

jrconlin:
will also adjust the test page and vapid libraries to follow form.

autopush/tests/test_endpoint.py
kitcambridge: I'm happy to change the test to use different vectors; I just used these because they were already there. It's more about laziness than efficiency. :wink:

kitcambridge:
Thanks, @jrconlin! :eyes:

autopush/tests/test_endpoint.py
jrconlin: Meh, don't worry about it. On second read, it's clear enough that these are the converted values.

jrconlin:
r+

autopush/settings.py
```diff
@@ -322,9 +324,15 @@ def parse_endpoint(self, token, version="v0", ckey_header=None):
...
         if ckey_header:
             try:
-                public_key = CryptoKey(ckey_header).get_label('p256ecdsa')
+                crypto_key = CryptoKey(ckey_header)
             except CryptoKeyException:
                 raise InvalidTokenException("Invalid key data")
+            label = crypto_key.get_label('p256ecdsa')
+            try:
+                public_key = base64url_decode(label)
+            except:
```
bbangert: What exception can be thrown here? Generally bare excepts are to be avoided when possible.

PR #443, opened 2016-04-16
feat: Support app server keys via the HTTP interface

### LINKED ISSUE #423: Add crypto-key handling to HTTP registration endpoint
Add Public-Key registration or Android and other clients that will not have web-socket availability, as well as check function parity for all other newly added features.

kitcambridge:
I think an optional `key` field in the body is all that's needed. Closes #423.

@jrconlin r?


jrconlin:
r+

PR #444, opened 2016-04-19
bug: Use static UAIDs for preflight, clean up after.

### LINKED ISSUE #434: Random HTTP probing creates users
Hitting the HTTP server in some cases seems to create router table records. Seen in test environments spun up randomly in AWS.

jrconlin:
Preflight would create new accounts which could throw off metrics.
This patch solves this in two ways; It uses static UAIDs matching
"deadbeef00000000deadbeef\d{8}" and deletes the created data once
the preflight is done.

Closes #434 

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -161,14 +161,14 @@ def get_storage_table(tablename="storage", read_throughput=5,
                        write_throughput)
 
 
-def preflight_check(storage, router):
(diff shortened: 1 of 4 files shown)


bbangert:
No test to see that nothing in the db was left afterward?

bbangert:
r+

PR #446, opened 2016-04-19
feat: Add extended err message for old encryption

### LINKED ISSUE #432: Idea: return a 400 with an explanation for `Content-Encoding: aesgcm128` + Crypto-Key
If https://bugzilla.mozilla.org/show_bug.cgi?id=1257821 is uplifted to Beta, Firefox will only support two content encodings:

* `Content-Encoding: aesgcm128` + `Encryption-Key`
* `Content-Encoding: aesgcm` + `Crypto-Key`

The intermediate state (`Content-Encoding: aesgcm128` + `Crypto-Key`) currently exists in Firefox 46. Since it's also incompatible with Chrome, I don't think too many developers have been using it...but it might be nice for us to detect it early and return a special error code and message ("You're using outdated encryption; please update to the format described in https://developers.google.com/web/updates/2016/03/web-push-encryption").

WDYT?

jrconlin:
Also fixes flake8 spacing error in router/apnsrouter.

closes #432 

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -410,6 +410,19 @@ class EndpointHandler(AutoendpointHandler):
         self.start_time = time.time()
         crypto_key_header = self.request.headers.get('crypto-key')
 
+        content_encoding = self.request.headers.get('content-encoding', "")
+        if content_encoding.lower() == 'aesgcm128' and crypto_key_header:
+            self.log.debug("Invalid crypto state; aesgcm128 + Crypto-Key",
+                           **self._client_info)
+            wpe_url = ("https://developers.google.com/web/updates/2016/03/"
+                       "web-push-encryption")
+            self._write_response(
+                400,
+                110,
+                message="You're using outdated encryption; "
+                "Please update to the format described in " + wpe_url)
+            return
+
         d = deferToThread(self.ap_settings.parse_endpoint,
                           token,
                           api_ver,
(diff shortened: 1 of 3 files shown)


autopush/router/apnsrouter.py
```diff
@@ -66,8 +66,8 @@ def _route(self, notification, router_data):
         """Blocking APNS call to route the notification"""
         token = router_data["token"]
         custom = {
-          "Chid": notification.channel_id,
-          "Ver": notification.version,
+            "Chid": notification.channel_id,
```
kitcambridge: Oops, sorry for introducing that! 🍇

autopush/endpoint.py
```diff
@@ -410,6 +410,20 @@ def put(self, api_ver="v0", token=None):
         self.start_time = time.time()
         crypto_key_header = self.request.headers.get('crypto-key')
 
+        content_encoding = self.request.headers.get('content-encoding')
+        if (content_encoding and content_encoding.lower() == 'aesgcm' and
```
kitcambridge: aesgcm128

kitcambridge:
r+ pending change to `aesgcm128`. `aesgcm` is the new hotness. :boom: :fire:

autopush/endpoint.py
```diff
@@ -410,6 +410,20 @@ def put(self, api_ver="v0", token=None):
         self.start_time = time.time()
         crypto_key_header = self.request.headers.get('crypto-key')
 
+        content_encoding = self.request.headers.get('content-encoding')
```
bbangert: Could default the get to "" to avoid the first bool check below.

bbangert:
r+ pending @kitcambridge's suggested fix

PR #447, opened 2016-04-20
Dockerfile changes:

oremj:
* remove MAINTAINER - it creates an extra layer
* FROM an offical docker image
* use /app by convention
* don't worry about trimming -dev packages


bbangert:
r+

PR #449, opened 2016-04-20
fix: Correct documents to strongly recommend well formatted UUIDs

### LINKED ISSUE #392: Docs: Update docs to reflect normalized UUID
All UUIDs are normalized to lower case with '-' separators. We need to note that in the docs so that app creators who use the UUID as a string index don't get random failures or unexpected results.

jrconlin:
The code normalizes UUIDs used for channelIDs and UIADs to lower case,
dashed format. The docs now state that and the code will reject UUIDs
that are not well formatted. This is important for clients that may
normalize differently.

BREAKING CHANGE: This patch will reject UUIDs that are not expressed as
lower case, dash formatted (e.g. "00001111-aaaa-bbbb-cccc-222233334444")

Closes #392

@bbangert , @kitcambridge r?

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -50,6 +50,7 @@ def tearDown():
 
 
 class WebsocketTestCase(unittest.TestCase):
+
     def setUp(self):
         from twisted.logger import Logger
         twisted.internet.base.DelayedCall.debug = True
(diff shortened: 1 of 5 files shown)


bbangert:
r+

kitcambridge:
Retroactive r+. Thanks, JR!

PR #452, opened 2016-04-21
bug: Strip padding from key content

### LINKED ISSUE #451: Reject padded Base64 in crypto headers
@martinthomson suggested rejecting padded params in https://bugzilla.mozilla.org/show_bug.cgi?id=1256488#c8, and [draft-ietf-httpbis-encryption-encoding](http://httpwg.org/http-extensions/draft-ietf-httpbis-encryption-encoding.html) cites RFC 7515, too.

Unfortunately, the [test page](https://jrconlin.github.io/WebPushDataTestPage/) used the padded form for a while, and I suspect others might, too. These senders will break once bug 1256488 lands.

I think we should detect these on the server and return a more helpful error, instead of having the client fail decryption.

+@jrconlin

jrconlin:
The client will soon reject any key content that includes padding.

The server will need to watch and strip it. We should also return a
warning message, however the likelihood of that warning being noticed
is minimal.

Closes #451 

@bbangert, @kitcambridge r?

(Let the great debate ensue!)

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -395,6 +395,9 @@ class EndpointHandler(AutoendpointHandler):
                             "encryption-key", "content-type",
                             "authorization"]
     cors_response_headers = ["location", "www-authenticate"]
+    # Remove trailing padding characters from complex header items like
+    # Crypto-Key and Encryption
+    strip_padding = re.compile('=+(?=[,;]|$)')
 
     #############################################################
     #                    Cyclone HTTP Methods
@@ -409,7 +412,6 @@ class EndpointHandler(AutoendpointHandler):
         api_ver = api_ver or "v0"
         self.start_time = time.time()
         crypto_key_header = self.request.headers.get('crypto-key')
-
         content_encoding = self.request.headers.get('content-encoding', "")
         if content_encoding.lower() == 'aesgcm128' and crypto_key_header:
             self.log.debug("Invalid crypto state; aesgcm128 + Crypto-Key",
@@ -518,6 +520,14 @@ class EndpointHandler(AutoendpointHandler):
 
     def _route_notification(self, version, result, data, ttl=None):
         self.version = self._client_info['message_id'] = version
+        warning = ""
+        # Clean up the header values (remove padding)
+        for hdr in ['crypto-key', 'encryption']:
+            if self.strip_padding.search(self.request.headers.get(hdr, "")):
+                warning = ("Padded content detected. Please strip"
+                           " base64 encoding padding.")
+                head = self.request.headers[hdr].replace('"', '')
+                self.request.headers[hdr] = self.strip_padding.sub("", head)
         notification = Notification(version=version, data=data,
                                     channel_id=self.chid,
                                     headers=self.request.headers,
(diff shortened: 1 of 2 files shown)


autopush/endpoint.py
```diff
@@ -395,6 +395,9 @@ class EndpointHandler(AutoendpointHandler):
                             "encryption-key", "content-type",
                             "authorization"]
     cors_response_headers = ["location", "www-authenticate"]
+    # Remove trailing padding characters from complex header items like
+    # Crypto-Key and Encryption
+    strip_pad = re.compile("=+(?=[,;]|$)")
```
kitcambridge: I think we'll want to account for the quote (`[,;"]`), too, in case they pass a quoted string.

autopush/endpoint.py
```diff
@@ -395,6 +395,9 @@ class EndpointHandler(AutoendpointHandler):
                             "encryption-key", "content-type",
                             "authorization"]
     cors_response_headers = ["location", "www-authenticate"]
+    # Remove trailing padding characters from complex header items like
+    # Crypto-Key and Encryption
+    strip_pad = re.compile('=+(?=[,;]|$)')
```
bbangert: strip_padding? Shortening it there makes it sound... odd.

autopush/endpoint.py
```diff
@@ -558,6 +569,9 @@ def _router_completed(self, response, uaid_data):
                               **self._client_info)
             time_diff = time.time() - self.start_time
             self.metrics.timing("updates.handled", duration=time_diff)
+            if warning:
+                response.response_body += (
+                    " "[:len(response.response_body) > 0] + warning)
```
bbangert: Ok, this is the most creative way I've seen to attach a conditional space. 

:clap:

bbangert: Though I am inclined to consider this marginally easier to reason about:
```python
response.response_body = (response.response_body + warning).strip()
```

bbangert:
r+ pending minor nits. strip_pad is meh, but the other method of conditionally including the space should at least have a comment indicating why its doing such a list splice with a boolean eval back to a int.

kitcambridge:
Thinking about this some more, I'm now wondering if we shouldn't do this.

Aurora 48 and Nightly 49 already reject padding, and we haven't seen breakage yet. The spec is also new enough to where we can discourage input like this, and Google will follow suit (they suggested prohibiting padding). Finally, we don't validate headers besides `Crypto-Key`; I could send `Encryption: chocolate=yes`, and the server would happily pass it through to the client.

But I could also see this being a common enough mistake.

jrconlin:
I view the padding issue as a needless annoyance to the developer. They already have several significant hurdles to overcome (encryption, user data management, scaling) and then us rejecting their submission because they included characters that literally signify nothing would certainly put me into a foul mood. It would be like having my thesis rejected because the staple was facing the wrong way.  I can understand where being more pedantic would make sense if we had elements that required strict formatting (A somewhat legitimate example is the VAPID JWT header, where extra padding existing in the header or claims absolutely impacts the generated signature) but that case is not true for Crypto-Key or other fields. 

Doing a simple, low cost action like ignoring extra padding makes lives slightly better. I'm ok with doing that.

bbangert:
r+

Issue #457, opened 2016-04-28
Log status_code for all endpoint errors


bbangert:
Only router failure logs status_code, all non-200 should include it.

PR #462, opened 2016-05-05
bug: Log status_code & errno for all errors

jrconlin:
fixes: #457 

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -219,13 +219,14 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         running.
 
         """
-        self.log.failure(fail, **self._client_info)
+        self.log.failure(fail, status_code=500, errno=999, **self._client_info)
         self._write_response(500, 999)
 
     def _overload_err(self, fail):
         """errBack for throughput provisioned exceptions"""
         fail.trap(ProvisionedThroughputExceededException)
-        self.log.info("Throughput Exceeded", **self._client_info)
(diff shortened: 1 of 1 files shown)


bbangert:
Commit message should be 'fix' instead of 'bug', and 'Closes #457', I don't believe fixes: like that will auto-close it.

autopush/endpoint.py
```diff
@@ -219,13 +219,14 @@ def _response_err(self, fail):
...
 
     def _overload_err(self, fail):
         """errBack for throughput provisioned exceptions"""
         fail.trap(ProvisionedThroughputExceededException)
-        self.log.info("Throughput Exceeded", **self._client_info)
+        self.log.info("Throughput Exceeded", status_code=503, errno=201,
+                      **self._client_info)
         self._write_response(503, 201)
 
     def _router_response(self, response):
```
bbangert: Is there any caller to _router_response that hasn't already logged out the 5xx?

jrconlin: Well, _router_fail_err wasn't if exc.log_exception was set to False for whatever reason. You've got a point that if `fail` isn't a RouterException, there's a real risk that we could be burying the real problem by trying to call `log_exception` on a general exception. I'll try a few tests on that.

jrconlin: Added an instance check in that method. Obviously, if there's something else getting triggered, we'll need to address that, but hopefully we shouldn't trigger a 500 because of an invalid attribute reference.

bbangert:
r+

PR #463, opened 2016-05-05
fix: Check connected month bounds for preflight

### LINKED ISSUE #461: Nonexisting message tables are still accessed, causing errors
```
May 05 16:52:04 ip-172-31-39-23 python[11560]: File "/data/autopush/app/autopush/router/webpush.py", line 74, in preflight_check
May 05 16:52:04 ip-172-31-39-23 python[11560]: self.ap_settings.message_tables[month_table].all_channels,
May 05 16:52:04 ip-172-31-39-23 python[11560]: exceptions.KeyError: u'prod.autopush.message_2016_3' 
```

It appears no handling is done to ensure the user has a new enough month when delivering messages. Since message tables 'too old' should be considered users who will be invalidated on a new connect, this issue should be handled as a 410 since the endpoint will never be valid again.

jrconlin:
Add full checks for preflight failure conditions.

Closes #461 

@bbangert r?

### DIFF
--- a/autopush/router/webpush.py
+++ b/autopush/router/webpush.py
@@ -66,17 +66,21 @@ class WebPushRouter(SimpleRouter):
         record = yield deferToThread(self.ap_settings.router.get_uaid, uaid)
 
         if 'current_month' not in record:
-            raise RouterException("No such subscription", status_code=404,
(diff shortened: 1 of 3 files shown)


bbangert:
r+, nice!

PR #464, opened 2016-05-06
bug: Wrap twisted formatEvent in a try handler

### LINKED ISSUE #460: Failure logging is incorrect
Logging errors is broken as instances are still sneaking in, better checks need to be done, and the format attempt itself should be in a try/except, with errors sentry reported.

jrconlin:
Declare the arguments for raven to hopefully prevent key corruption

Closes #460 
@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -219,14 +219,16 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
         running.
 
         """
-        self.log.failure(fail, status_code=500, errno=999, **self._client_info)
+        fmt = fail.value.message or 'Exception'
+        self.log.failure(format=fmt, failure=fail,
+                         status_code=500, errno=999, **self._client_info)
         self._write_response(500, 999)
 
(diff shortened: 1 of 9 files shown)


autopush/endpoint.py
```diff
@@ -219,14 +219,16 @@ def _response_err(self, fail):
         running.
 
         """
-        self.log.failure(fail, status_code=500, errno=999, **self._client_info)
+        format = fail.value.message or 'Exception'
```
bbangert: format is a builtin function

autopush/utils.py
```diff
@@ -124,8 +124,11 @@ def write_error(self, code, **kwargs):
         """
         self.set_status(code)
         if "exc_info" in kwargs:
-            self.log.failure(failure.Failure(*kwargs["exc_info"]),
-                             **self._client_info)
+            format = kwargs.get("format", "Exception")
```
bbangert: builtin

bbangert:
r+

PR #467, opened 2016-05-09
bug: Normalize padding handling for restricted subscriptions

### LINKED ISSUE #466: Padding for Subscription Locked endpoint causes bogus 400-201 error
There is a problem with subscription locked endpoints fernet decoding if the padding is altered.

jrconlin:
Since fernet passes it's content through base64 handling, it's just as
picky about padding as other base64 instances. For now, strip padding
from outbound endpoints and reapply it (if needed) to inbound endpoints.

Closes #466 

@bbangert r?

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -3,6 +3,7 @@ import sys
 import time
 import uuid
 import random
+import base64
 
 from hashlib import sha256
 
@@ -178,6 +179,7 @@ class EndpointTestCase(unittest.TestCase):
             hostname="localhost",
             statsd_host=None,
         )
+        self.old_fernet = settings.fernet
         self.fernet_mock = settings.fernet = Mock(spec=Fernet)
         self.metrics_mock = settings.metrics = Mock(spec=Metrics)
         self.agent_mock = settings.agent = Mock(spec=Agent)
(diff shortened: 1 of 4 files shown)


bbangert:
This changes generation of new endpoints, is there a test already that has old endpoints with the padding so we can be 100% sure existing endpoints work?

kitcambridge:
We should use BaseEmoji instead of Base64. @jrconlin suggests using 😭 for padding.

autopush/tests/test_endpoint.py
```diff
@@ -1230,6 +1232,30 @@ def test_padding(self):
...
+        dummy_key = ("BHolMkL36ucQsRe_0KRS70JyHB55H4C5Igv2YQEVNzCILN"
+                     "nedxFHSPtzI4KhzNtN2YPqHe7-mWW6_uvaIc5yEDk")
+        # Generate an endpoint, since the fernet key is not locked during
+        # testing.
+        while True:
+            ep = self.endpoint.ap_settings.make_endpoint(
+                dummy_uaid,
+                dummy_chid,
+                dummy_key)
+            (_, token) = ep.split("/")[-2:]
```
bbangert: Parens on left side not needed.

autopush/tests/test_endpoint.py
```diff
@@ -1230,6 +1232,30 @@ def test_padding(self):
...
+            off = len(token) % 4
+            if off:
```
bbangert: Combine? 'if len(token...'

bbangert:
r+, minor python nits

PR #468, opened 2016-05-10
feat: use gnureadline instead of readline

bbangert:
gnureadline is still maintained while readline has fallen behind and
has compile errors on some platforms.

@jrconlin r?


jrconlin:
r+

PR #470, opened 2016-05-13
chore: update all libs to latest versions and fix jws conflict

### LINKED ISSUE #453: Add requires.io dependency tracking
We should track the projects dependencies to ensure we stay up-to-date and don't miss security updates, etc.

The choices seem to be:

- gemnasium
- requires.io

bbangert:
Latest crypto library update introduced a change that this fixes as
well.

Closes #453 

@jrconlin r?


autopush/utils.py
```diff
@@ -115,7 +116,11 @@ def extract_jwt(token, crypto_key):
 
     key = decipher_public_key(crypto_key)
     vk = ecdsa.VerifyingKey.from_string(key, curve=ecdsa.NIST256p)
-    return jws.verify(token, vk, algorithms=["ES256"])
+    # jose offers jwt.decode(token, vk, ...) which does a full check
+    # on the JWT object. Vapid is a bit more creative in how it
+    # stores data into a JWT and breaks expectations. We would have to
+    # turn off most of the validation in order for it to be useful.
+    return json.loads(jws.verify(token, vk, algorithms=["ES256"]))
```
jrconlin: part of me wonders if/when this comes back to bite us, but looking at the jose code, they'd have to change a bunch of stuff to get the old behavior back.

jrconlin:
r+

PR #471, opened 2016-05-13
chore: add default.nix for nix/nixos users

bbangert:
A basic default for nix-shell for nix usage that eases development.

@jrconlin r?


jrconlin:
r+

PR #474, opened 2016-05-18
feat: Log naks from bridged connections via REST API

### LINKED ISSUE #469: Add an HTTP API for reporting decryption errors and NACKs
Similar to #380, but for the HTTP API, so that Android can use it.

/cc @bbangert

jrconlin:
Bridged devices do not use the websocket connection, so NAKs can be
lost. Our system generally ACKs messages when they're handed off to the
bridge system, but clients can 'retro-actively' NAK a packet.

Since connections should be limited to the minimum possible, and NAKs
are (hopefully) more rare and meaningful than ACKs, only need to log
NAKs.

Also fix up missing coverage for bad senderid JSON.

Closes #469

@bbangert r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -712,7 +712,8 @@ class RegistrationHandler(AutoendpointHandler):
         self.add_header("Content-Type", "application/json")
         d = Deferred()
         d.addCallback(router.register, router_data)
-        d.addCallback(self._save_router_data, router_type)
+        if params:
+            d.addCallback(self._save_router_data, router_type)
         d.addCallback(self._success)
         d.addErrback(self._router_fail_err)
         d.addErrback(self._response_err)
@@ -824,7 +825,7 @@ class RegistrationHandler(AutoendpointHandler):
                        **self._client_info)
         self.finish()
 
-    def _success(self, result):
+    def _success(self, result=None):
         """Writes out empty 200 response"""
         self.write({})
         self.finish()
@@ -867,3 +868,65 @@ class RegistrationHandler(AutoendpointHandler):
             return json.loads(self.request.body)
         except ValueError:
             return {}
+
+
+class BridgeMessageHandler(RegistrationHandler):
+    cors_methods = "PUT"
+
+    @cyclone.web.asynchronous
+    def put(self, uaid):
+        """HTTP PUT
+
+        Process message states for a UAID
+
+        """
+        self.start_time = time.time()
+        if not self._validate_auth(uaid):
+            return self._write_unauthorized_response(
+                message="Invalid Authentication")
+        params = self._load_params()
+        if 'nak' in params:
+            d = deferToThread(self._record_naks, uaid, naks=params['nak'])
+            d.addCallback(self._success)
+            d.addErrback(self._response_err)
+            return d
+        return self._success()
+
+    def _record_naks(self, o_uaid, naks):
+        """ Record the NAKs from the client
+
+        """
+        for nak in naks:
+            try:
+                result = self.ap_settings.fernet.decrypt(
+                    nak['id'].encode('utf8'))
+                kind, uaid, chid = result.split(":")
+                if kind != 'm':
+                    self.log.error(
+                        format="Improper messageID specified",
+                        message_id=result,
+                        uaid_hash=hasher(uaid),
+                        channel_id=chid,
+                        message=json.dumps(nak))
+                    continue
+                try:
+                    self.log.info(
+                        format="NAK",
+                        uaid_hash=hasher(uaid),
+                        channel_id=chid,
+                        user_agent=self.request.user_agent,
+                        message_id=result,
+                        message=nak.get("msg", None),
+                        code=nak.get("code", 0))
+                except Exception as e:
+                    self.log.failure(
+                        format="Unable to process record",
+                        uaid_hash=hasher(uaid),
+                        line=json.dumps(nak, sort_keys=True),
+                        error=repr(e))
+            except (ValueError, InvalidToken) as e:
+                self.log.failure(
+                    format="Invalid messageID specified",
+                    uaid_hash=hasher(o_uaid),
+                    line=json.dumps(nak, sort_keys=True),
+                    error=repr(e))
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -75,6 +75,7 @@ class GCMRouter(object):
             data['body'] = notification.data
             data['con'] = notification.headers['content-encoding']
             data['enc'] = notification.headers['encryption']
+            data['id'] = notification.version
 
             if 'crypto-key' in notification.headers:
                 data['cryptokey'] = notification.headers['crypto-key']
--- a/autopush/tests/test_main.py
+++ b/autopush/tests/test_main.py
@@ -235,10 +235,14 @@ class EndpointMainTestCase(unittest.TestCase):
             "--s3_bucket=none",
         ])
 
-    def test_bad_senderidlist(self):
+    @patch('autopush.main._parse_endpoint')
+    def test_bad_senderidlist(self, parse):
+        ff = Mock()
+        ff.senderid_list = "[Invalid"
+        parse.return_value = (ff, Mock())
         endpoint_main([
             "--senderid_list='[Invalid'"
-        ])
+        ], False)
 
     def test_ping_settings(self):
         ap = make_settings(self.test_arg)
--- a/docs/http.rst
+++ b/docs/http.rst
@@ -3,7 +3,7 @@
 HTTP Endpoints for Notifications
 ================================
 
-Autopush exposes three HTTP endpoints:
+Autopush exposes four HTTP endpoints:
 
 `/push/...`
 
@@ -22,6 +22,13 @@ and :ref:`update`.
 
 This is tied to the :class:`RegistrationHandler` (:ref:`endpoint_module`). This endpoint is used by
 apps that wish to use :term:`bridging` protocols to register new channels.
+See :ref:`bridge_api`.
+
+`/v1/.../.../messages/...`
+
+This is tied to the :class:`BridgeMessageHandler` (ref:`endpoint_module`).
+This endpoint is used by apps that wish to use the :term:`bridging`
+protocols to report the status of handled messages.
 See :ref:`bridge_api`.
 
 ---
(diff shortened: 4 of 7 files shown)


autopush/tests/test_main.py
```diff
@@ -235,10 +235,14 @@ def test_ssl(self):
             "--s3_bucket=none",
         ])
 
-    def test_bad_senderidlist(self):
+    @patch('autopush.main._parse_endpoint')
```
jrconlin: Huh, was getting local coverage misses off of this. My local install was using values pulled from local files, which is horribly invalid anyway. This ensures that log.critical is called.

bbangert:
There are no doc updates indicating how a mobile client should ack/nack, and this seems a bit problematic to overload the PUT for "Update router type/data for a UAID." to also be for ack/nack's.

This should likely be its own separate method, in a more RESTful manner, ie:

PUT /messages/{message-id} with nack/ack info or something.

jrconlin:
Are we saying that NAKs must be done one at a time rather than in bulk?

kitcambridge:
Considering we expect them to be rare, I don't see much value in batching them up, but it's your call.

jrconlin:
Ok, refactored to the new messageid method. 

Adding @kitcambridge to r? (to verify it meets his needs)
@bbangert can you spot check I am using defer.gatherresults() correctly?

autopush/endpoint.py
```diff
@@ -867,3 +868,69 @@ def _load_params(self):
...
+            if kind != 'm':
+                self.log.error(
+                    format="Improper messageID specified",
+                    message_id=result,
+                    uaid_hash=hasher(uaid),
+                    channel_id=chid,
+                    message=json.dumps(params))
+                return
+            try:
+                message_id, _ = params.get("ver").split(":")
```
kitcambridge: IIUC, `id` and `ver` are identical, so we can trim this down even more. :wink:

autopush/endpoint.py
```diff
@@ -867,3 +868,69 @@ def _load_params(self):
...
+        """
+        self.start_time = time.time()
+        if not self._validate_auth(uaid):
+            return self._write_unauthorized_response(
+                message="Invalid Authentication")
+        params = self._load_params()
+        if 'nak' in params:
+            naks = params['nak']
+            batch = []
+            for nak in naks:
```
kitcambridge: It might be slightly more efficient to decrypt everything on a single thread, instead of punting to the pool for every nack. Something like `deferToThread(map, lambda msg_id: self.ap_settings.fernet.decrypt(msg_id.encode('utf8')), naks)`. Then again, we're sending so few nacks that it probably doesn't matter.

:bike: :art: :house: (I ❤️ this emoji).

docs/http.rst
```diff
@@ -487,3 +495,63 @@ Remove a given ChannelID subscription from a UAID.
...
+    }
+
+
+For this example:
+
+    :id: The message ID included in the notification data.
+
+    :ver: The message version field included in the notification data.
+
+    :code: An optional return code indicating the type of error encountered. Error codes may be specific to individual bridge implementations.
```
kitcambridge: That reminds me; I think we still need to document the codes that the client sends (https://dxr.mozilla.org/mozilla-central/rev/c67dc1f9fab86d4f2cf3224307809c44fe3ce820/dom/push/PushServiceWebSocket.jsm#41-59).

kitcambridge:
Thanks so much, @jrconlin! Nice docs, too. r+ with feedback addressed.

autopush/endpoint.py
jrconlin: Yeah, I was debating that a fair bit. Frankly, if I was kicking to a thread, might as well have it just cycle through the naks, decrypt each id and log them. Breaking them into individual thread hits obviously gets through them faster at the cost of potentially chewing up the thread pool if we get a slew of NAKs. 

Meh, I'll stick to one thread to process the stream of naks, and burn the refactor bridge later.
 🐃 🔜 🐩

kitcambridge: > 🐃 🔜 🐩

👍 😁 😸

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+        self.start_time = time.time()
+        if not self._validate_auth(uaid):
+            return self._write_unauthorized_response(
+                message="Invalid Authentication")
+        params = self._load_params()
+        if 'nak' in params:
+            d = deferToThread(self._record_naks, uaid, naks=params['nak'])
+        else:
+            d = Deferred()
+            d.callback(True)
```
bbangert: Why not just return self._success() ?

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+                    continue
+                try:
+                    self.log.info(
+                        format="NAK",
+                        uaid_hash=hasher(uaid),
+                        channel_id=chid,
+                        user_agent=self.request.user_agent,
+                        message_id=result,
+                        message=nak.get("msg", None),
+                        code=nak.get("code", 0))
```
bbangert: Is there any utility in logging a nack that has no code or message? Under what case does that occur?

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+                        format="Unable to process record",
+                        uaid_hash=hasher(uaid),
+                        line=json.dumps(nak, sort_keys=True),
+                        error=repr(e))
+            except (ValueError, InvalidToken) as e:
+                self.log.error(
+                    format="Invalid messageID specified",
+                    uaid_hash=hasher(o_uaid),
+                    line=json.dumps(nak, sort_keys=True),
+                    error=repr(e))
```
bbangert: Is this an application level error of our software if the messageID is invalid?

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+                        user_agent=self.request.user_agent,
+                        message_id=result,
+                        message=nak.get("msg", None),
+                        code=nak.get("code", 0))
+                except Exception as e:
+                    self.log.error(
+                        format="Unable to process record",
+                        uaid_hash=hasher(uaid),
+                        line=json.dumps(nak, sort_keys=True),
+                        error=repr(e))
```
bbangert: Should this be .error, or .failure (which properly capture the entirety of the exception stack)?

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+        """ Record the NAKs from the client
+
+        """
+        for nak in naks:
+            try:
+                result = self.ap_settings.fernet.decrypt(
+                    nak['id'].encode('utf8'))
+                kind, uaid, chid = result.split(":")
+                if kind != 'm':
+                    self.log.error(
```
bbangert: Application error for a client giving us a bad messageID?

autopush/endpoint.py
```diff
@@ -867,3 +868,67 @@ def _load_params(self):
...
+                        message=nak.get("msg", None),
+                        code=nak.get("code", 0))
+                except Exception as e:
+                    self.log.error(
+                        format="Unable to process record",
+                        uaid_hash=hasher(uaid),
+                        line=json.dumps(nak, sort_keys=True),
+                        error=repr(e))
+            except (ValueError, InvalidToken) as e:
+                self.log.error(
```
bbangert: If we should be reporting this as a true error, .failure might be better.

autopush/router/gcm.py
```diff
@@ -75,6 +75,7 @@ def _route(self, notification, router_data):
             data['body'] = notification.data
             data['con'] = notification.headers['content-encoding']
             data['enc'] = notification.headers['encryption']
+            data['id'] = notification.version
```
bbangert: Does APNS need a similar id for its nack reporting? Or does it already add that?

docs/http.rst
```diff
@@ -487,3 +496,78 @@ Remove a given ChannelID subscription from a UAID.
 **Return Codes:**
 
 See :ref:`errors`.
+
+Report ACKs and NAKs
+~~~~~~~~~~~
```
bbangert: The ~ needs to match the line length above it. Does it match here and the mismatch is just the github font?

autopush/endpoint.py
jrconlin: I can see where a NAK may contain just the channelid. (e.g. the client either cannot determine the code or msg, or fails to properly send that data). This could still be useful for flow debugging.

autopush/endpoint.py
jrconlin: There may be any number of reasons that fenet might throw an InvalidToken (perhaps the crypto_key list removed a valid key too early, or blanked values, or specified an invalid key value). Likewise, if the decrypted result cannot be split into three parts, then whatever created the id should be examined.

autopush/endpoint.py
jrconlin: Since this value is usually echoed back from data provided in the subscription update, and it's encrypted and decrypted by our server, it's reasonable to presume it may have come from our server. Of course, there is the potential that it came from a client error, but since we have a tight relationship with the client code, it's worth noting.

autopush/router/gcm.py
jrconlin: It's fair to say that APNs is on "hold" for the time being. It's equally fair to say we should remove APNs support until such time as it's requested, then revisit the APNs bridge system.

autopush/endpoint.py
kitcambridge: Oh, I meant to ask: what's the message for? The WebSocket client sends a `code`, but I don't remember what the message does.

jrconlin: Ah, I probably misread that code and thought that a `msg` field was present. Yes, the role is just to provide additional information for logging if desired. I can remove it if this will never be used.

autopush/endpoint.py
bbangert: So if we see this logged, then it means our app is malfunctioning, and we'll need to open a bug to fix something that caused it, right?

I'm just trying to make sure we're logging actionable things, rather than adding more noise to our logstream (which is annoying noisy already).

bbangert:
r+

autopush/endpoint.py
jrconlin: If we see this crop up, we should definitely investigate. There's a chance it's coming from the client, but we'd need to chase that as well.

In any case, it should be sub digit probability of occurring.

PR #479, opened 2016-05-24
feat: Add endpoint to test logging

### LINKED ISSUE #478: Create http call point to test error reporting
With the low number of errors and info messages being generated, there is concern that logging may break and we'd be unaware of any problem.

Create an endpoint that generates a demo error to validate logging.

jrconlin:
call /v1/err to invoke a fake "error" for testing purposes.
Responds with a status of 418 and creates an appropriate log entry

Closes #478 

@bbangert  r?


autopush/log_check.py
```diff
@@ -0,0 +1,25 @@
...
+    @cyclone.web.asynchronous
+    def get(self):
+        """HTTP GET
+
+        Generate a dummy error message for logging
+
+        """
+        self.log.error(format="Test Error Message",
+                       status_code=418, errno=0,
+                       **self._client_info)
```
bbangert: This should throw an exception, then report it via log.failure so we can verify in dev/stage that Sentry gets it.

jrconlin: Going to do both.

`/v1/err` or `/v1/err/error` will return a standard error.
`/v1/err/critical` will throw the exception and log the failure.

This will let us do other triggered states in the future, rather than create a lot of one off handlers.

autopush/log_check.py
```diff
@@ -0,0 +1,49 @@
...
+            errType = "error"
+        else:
+            errType = errType.lower()
+        if 'error' in errType:
+            self.log.error(format="Test Error Message",
+                           status_code=418, errno=0,
+                           **self._client_info)
+            self._write_response(418, 999, message="ERROR:Success",
+                                 reason="Test Error")
+        if 'crit' in errType:
```
bbangert: Why not just...
```python
if 'crit' in errType:
    try:
        raise LogCheckError("Logcheck")
    except:
        self.log.failure("Test error message", ...)
```

log.failure should automatically pull out the active exception if no failure is passed to it (the docs seem to indicate that at least)

jrconlin: Didn't know how close to reality you wanted. I wasn't sure if twisted's defer might add additional context information you might want to check, but I can do the simpler version.

bbangert:
This looks good, just needs a rebasing as I had to include some of the commits included here in master already.

bbangert:
r+

Issue #379, opened 2016-02-26
Schema validation refactor in endpoint


bbangert:
Our code has slowly grown an if/else heap of validation checks over time. We should refactor to use a proper validation library, my thought is to use colander (http://docs.pylonsproject.org/projects/colander/en/latest/index.html).

We will need a slightly custom decorator for our cyclone functions that kicks the validation off to a thread, before calling back to the function if validation succeeds.

Because the validation should occur in a thread-pool, it can do additional checks all at once that we typically punt to the thread-pool for multiple times, such as:

- Fernet decryption of the token
- UAID lookup in AWS
- CHID lookup in AWS

These checks can all be implemented as custom validators in colander so that we end up with a single consistent schema of what headers/payloads/etc are required that is easier to maintain and verify.

PR #484, opened 2016-06-03
feat: refactor simplepush endpoint for validation schemas

bbangert:
This refactor splits out the base handler and several bits of
generic callbacks into a base.py module. Each web handler will
now get its own module, with simplepush.py as the initial
refactor candidate.

Marshmallow schemas are used to validate incoming requests and
format/coerce/validate data in a separate thread. This allows
the validation to make AWS calls or other expensive calls that
previously required punting calls back/forth between the twisted
thread-pool.

The SimplePush endpoints issued are now under a new spush prefix
so that they can go to a separate handler entirely that has the
schema validation. WebPush endpoints will similarly be sent to
a new prefix. An additional toggle is stored on the user object
to track whether their record was upgraded for this change so
that at a later point we can expire-on-connect old user records.

Issue #379 

@jrconlin, @kitcambridge: r?


autopush/web/validation.py
```diff
@@ -0,0 +1,166 @@
...
+            result = self.context["settings"].parse_endpoint(
+                token=d["token"],
+                version=d["api_ver"],
+            )
+        except InvalidTokenException:
+            raise InvalidRequest("invalid token", errno=102)
+        return result
+
+    @validates_schema
+    def validate_uaid_chid(self, d):
```
kitcambridge: OK...so Marshmallow runs the `@pre_load` decorators first, to build the validation object, then checks and converts the fields, then calls `validates_schema` with the modified object? And calling `schema.load()` kicks off the process and returns the object?

autopush/web/validation.py
```diff
@@ -0,0 +1,166 @@
...
+        body_string = d["body"]
+        version = data = None
+        if len(body_string) > 0:
+            body_args = urlparse.parse_qs(body_string, keep_blank_values=True)
+            version = body_args.get("version")
+            data = body_args.get("data")
+        else:
+            version = d["arguments"].get("version")
+            data = d["arguments"].get("data")
+        version = version[0] if version is not None else version
```
kitcambridge: It took a couple of reads to parse that this means `version = version[0] if version is not None else None`. How do you feel about writing this out as an `if` instead of a ternary?

autopush/websocket.py
```diff
@@ -647,6 +650,10 @@ def _register_user(self, existing_user=True):
         # users
         if not existing_user:
             user_item["last_connect"] = generate_last_connect()
+
+            # New users get a record_version so we can track changes that
+            # may require old user records to be expired on the fly
```
kitcambridge: Nice!

autopush/web/base.py
```diff
@@ -0,0 +1,234 @@
...
+        if self.ap_settings.cors:
+            self.set_header("Access-Control-Allow-Origin", "*")
+            self.set_header("Access-Control-Allow-Methods",
+                            self.cors_methods)
+            self.set_header("Access-Control-Allow-Headers",
+                            ",".join(self.cors_request_headers))
+            self.set_header("Access-Control-Expose-Headers",
+                            ",".join(self.cors_response_headers))
+
+    def write_error(self, code, **kwargs):
```
kitcambridge: We could eventually remove the `ErrorLogger` mix-in once the other handlers are converted, right?

autopush/web/base.py
```diff
@@ -0,0 +1,234 @@
...
+    #############################################################
+    #                    Cyclone HTTP Methods
+    #############################################################
+    def options(self, *args):
+        """HTTP OPTIONS Handler"""
+
+    def head(self, *args):
+        """HTTP HEAD Handler"""
+
+    #############################################################
```
kitcambridge: I'm assuming these are copied from `AutoendpointHandler`, so I only skimmed them. I can take another look later today.

autopush/web/base.py
```diff
@@ -0,0 +1,234 @@
...
+                   "version data channel_id headers ttl")):
+    """Parsed notification from the request"""
+
+
+class VapidAuthException(Exception):
+    """Exception if the VAPID Auth token fails"""
+    pass
+
+
+class BaseHandler(cyclone.web.RequestHandler):
```
kitcambridge: Thanks for splitting this out!

autopush/web/validation.py
```diff
@@ -0,0 +1,166 @@
...
+
+        """
+        return cls(schema)._decorator
+
+
+# Alias to the validation classmethod decorator
+threaded_validate = ThreadedValidate.validate
+
+
+class SimplepushSubscriptionSchema(Schema):
```
kitcambridge: Nit: SimplePushRequestSchema, to match the others.

autopush/tests/test_main.py
```diff
@@ -73,13 +74,15 @@ def test_update_rotating_tables(self):
...
         # Get the deferred back
+        e = Deferred()
         d = settings.update_rotating_tables()
 
         def check_tables(result):
             eq_(len(settings.message_tables), 1)
 
         d.addCallback(check_tables)
-        return d
+        d.addBoth(lambda x: e.callback(True))
```
kitcambridge: Interesting, so Twisted deferreds can't be chained like `d.addBoth(lambda x: True))`?

autopush/web/validation.py
bbangert: Yup, pre_load is allowed to modify/extract/etc as needed and the resulting dict is then coerced as indicated by the fields, then the validators run.

autopush/main.py
```diff
@@ -500,8 +501,10 @@ def endpoint_main(sysargs=None, use_files=True):
 
     # Endpoint HTTP router
     site = cyclone.web.Application([
-        (r"/push/(?:(v\d+)\/)?([^\/]+)", EndpointHandler,
+        (r"/push/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)", EndpointHandler,
          dict(ap_settings=settings)),
+        (r"/spush/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
```
kitcambridge: Do we want to invalidate existing Simple `/push` endpoints, or will your Web Push refactor handle them?

autopush/web/base.py
bbangert: The ErrorLogger is redundant with the new BaseHandler, yup. So the WebPush refactor to this setup will only inherit BaseHandler.

autopush/web/base.py
bbangert: Yup, the standard options/head bases.

kitcambridge:
Looks great, very clean! I'll probably read through this again later, but this is significantly better than the current state of things, so r+.

autopush/tests/test_main.py
bbangert: In this case, it was a race issue as the test was waiting on the deferred to fire as well as check_tables. Adding another deferred in the chain remedied the test consistently.

autopush/main.py
bbangert: We will eventually invalidate them, thus adding the version to new user records. And later, we can then force users without a record_version to get a new UAID which will invalidate existing simplepush /push endpoints.

autopush/tests/test_web_base.py
```diff
@@ -0,0 +1,257 @@
...
+        self.base.write_error(999)
+        self.status_mock.assert_called_with(999)
+        eq_(self.base.log.failure.called, True)
+
+    def test_init_info(self):
+        d = self.base._init_info()
+        eq_(d["request_id"], dummy_request_id)
+        h = self.request_mock.headers
+        h["user-agent"] = "myself"
+        d = self.base._init_info()
```
jrconlin: Since this is a hash lookup validation, and you're only testing one key item (without testing to see if other values in the hash are altered) Couldn't you set the mock headers, call `self.base._init_info()` and check the results instead of repeatedly calling `self.base._init_info()`?

autopush/web/validation.py
jrconlin: Might want to convert to int rather than just take the first character.

bbangert: It's a list actually, since it might come from the query which allows repeat values. This is code from the parse_ function in endpoint.py and functions the same. I let the field in the schema do the value coercion.

jrconlin:
r+

PR #485, opened 2016-06-03
chore: shorten delta output to just delta lines

bbangert:
Only during debugging sessinos does a full object summary of each
track statement assist. As such, by default only the deltas are
now shown in the summary.

@jrconlin r?


jrconlin:
r+

PR #489, opened 2016-06-28
feat: refactor webpush endpoint for validation schemas

### LINKED ISSUE #379: Schema validation refactor in endpoint

bbangert:
This refactor splits out the webpush endpoint to a new URL path
and uses the marshmallow schema to validate requests.

The webpush endpoints are issued under a new URL endpoint to
avoid conflicts with the existing.

Closes #379 

@jrconlin r?

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -1,1 +1,84 @@
-#
+import time
+
+from twisted.internet.defer import Deferred
+from twisted.internet.threads import deferToThread
+
+from autopush.web.base import (
+    BaseHandler,
+    Notification,
+)
+from autopush.web.validation import (
+    threaded_validate,
+    WebPushRequestSchema,
+)
+
+
+class WebPushHandler(BaseHandler):
+    cors_methods = "POST"
+    cors_request_headers = ["content-encoding", "encryption",
+                            "crypto-key", "ttl",
+                            "encryption-key", "content-type",
+                            "authorization"]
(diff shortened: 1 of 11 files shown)


autopush/web/webpush.py
```diff
@@ -1 +1,84 @@
...
+            else:
+                uaid_data["router_data"] = response.router_data
+            uaid_data["connected_at"] = int(time.time() * 1000)
+            d = deferToThread(self.ap_settings.router.register_user,
+                              uaid_data)
+            response.router_data = None
+            d.addCallback(lambda x: self._router_completed(response,
+                                                           uaid_data,
+                                                           warning))
+            return d
```
jrconlin: Coverage appears to be missing, but we've discussed this offline. We'll add back the coverage later.

jrconlin:
r+

PR #494, opened 2016-06-28
feat: add AMI instance ID to the logged information

jrconlin:
Introduces new shared option "--no_aws" which will prevent calling
AWS meta-information. This option defaults to "False" since autopush
generally always runs on AWS.

No breaking changes.

@bbangert r?


bbangert:
r+

PR #495, opened 2016-06-28
chore: add python27 dockerfile

oremj:


oremj:
Merging, this only adds an extra dockerfile and builds it in circle.

Issue #493, opened 2016-06-28
Errors aren't being sent to sentry


bbangert:
Sentry errors are not being reported when triggered, this bug started after the logging refactor.

PR #500, opened 2016-06-30
fix: remove failing key check from sentry logging

bbangert:
Twisteds new logging does not use the 'failure' key at all, the
prior assumptions it would exist caused key exceptions instead
of reporting errors. This fix changes to only the new style of
failure checking.

Issue #493 

@jrconlin r?


jrconlin:
r+ (with a sigh)

PR #503, opened 2016-06-30
fix: handle different behavior in rotating table tests at month-end

### LINKED ISSUE #502: Fix update_rotating_table tests for end of month
When its the end of the month, test coverage drops because of an error in our tests to not populate message_tables properly for the next month check.

bbangert:
Two of the tests for update_rotating_tables failed to account for
a different code-path that runs at the end of the month when the
next day is a new month. This change populates the next month
message table entry so that it won't matter if its the end of
the month or not for the tests to run properly.

Closes #502 

@jrconlin r?


jrconlin:
r+ pending travis kick.

PR #504, opened 2016-07-01
feat: change how bad bridge tokens are handled

### LINKED ISSUE #486: Revisit 301 handling for the GCM bridge
* The client treats non-200 responses as an error, so maybe we can remove the 301 instead of returning a new sender ID. Alternatively, if it's desirable for the client to change its sender ID or URL, we can make the client respect redirects.
* The `Location` header for the 301 is wrong; for example, `https://updates.push.services.mozilla.com/v1/gcm/965234145045/register` returns a 404. I think we might need to change it to `/registration`?

jrconlin:
Refactored things a bit to have router.check_token write out the
response. Added comment to discuss how clients fail on non 200, since
GCM may allow senderID swapping again in the future, or a new bridge
allows for token exchange.

Also changed up the relocation URL to be a bit more generic and handled
by the check_token function.

closes #486 

@bbangert r?


autopush/tests/test_router.py
```diff
@@ -836,5 +846,6 @@ def test_ammend(self):
         eq_(resp, self.router.amend_msg(resp))
 
     def test_check_token(self):
-        (t, v) = self.router.check_token("")
+        writer = Mock(spec=AutoendpointHandler._write_response)
+        (t, v) = self.router.check_token("", writer, "")
```
bbangert: The parens around t,v should be left off, like with t,w above.

autopush/tests/test_router.py
```diff
@@ -271,14 +272,15 @@ def test_register_bad(self):
...
 
-        (t, v) = self.router.check_token("test123")
+        (t, v) = self.router.check_token("test123", writer, None)
         ok_(t)
-        eq_(v, self.gcm_config['senderid_list'].keys()[0])
+        ok_(writer.called is False)
+        writer.reset_mock()
 
-        (t, v) = self.router.check_token("invalid")
+        (t, v) = self.router.check_token("invalid", writer, None)
```
bbangert: Parens should be dropped.

autopush/endpoint.py
```diff
@@ -649,12 +649,11 @@ def post(self, router_type="", router_token="", uaid="", chid=""):
             return self._write_response(
```
bbangert: There was an error here before then as we didn't error out?

jrconlin: Not sure I follow. This is the first instance in POST where we check the router type. The code modification below has router.check_token check token validity, as well as write any response, since each router may have a different method of responding to an invalid token. The only bit that can't be readily determined by the router is the relocation url, which we're building ahead of time and passing. (Few reasons for that, including the check token doesn't care how it was accessed and the URL requires content knowledge router doesn't need.)

bbangert: Sorry, think the highlighting was going wacky, it had this line flagged as green for code change, refresh changed it.

autopush/endpoint.py
```diff
@@ -649,12 +649,11 @@ def post(self, router_type="", router_token="", uaid="", chid=""):
...
                 400, 108, message="Invalid router")
         router = self.ap_settings.routers[router_type]
-        valid, router_token = router.check_token(router_token)
+        relo_url = self._relocate(router_type, uaid, chid)
+        valid, writer = router.check_token(router_token, self._write_response,
```
bbangert: I'm very nervous about inverting control of output writing to a router interface function. It drops the separation of responsibility for what chunk of code is responsible for which portion of the process.

autopush/endpoint.py
```diff
@@ -704,12 +703,11 @@ def put(self, router_type="", router_token="", uaid="", chid=""):
             return self._write_response(
                 400, 108, message="Invalid router")
         router = self.ap_settings.routers[router_type]
-        valid, router_token = router.check_token(router_token)
+        relo_url = self._relocate(router_type, uaid, chid)
+        valid, reply = router.check_token(router_token, self._write_response,
```
bbangert: In one spot its 'valid, writer', here its 'valid, reply'. I'm not sure why there's not consistency in the return naming.

autopush/router/gcm.py
```diff
@@ -34,10 +34,25 @@ def __init__(self, ap_settings, router_conf):
...
+            # are codifying to return a "400", which is technically both
+            # correct, and incorrect, depending on your point of view.
+            #
+            # if relocation were enabled...
+            #
+            # token = self.senderIDs.choose_ID().get('senderID')
+            # relo_url = relo % token
+            # return writer(301, 0 "Location: %s" + relo_url,
+            #               headers={"Location": relo_url)
+            return False, writer(400, 0, "Invalid Token")
```
bbangert: There's no point returning this since it was already called, which closed the reply by calling finish() on the cyclone RequestHandler.

autopush/endpoint.py
```diff
@@ -755,12 +753,11 @@ def delete(self, router_type="", router_token="", uaid="", chid=""):
...
+                                          relo_url)
         if not valid:
-            newUrl = self._relocate(router_type, router_token, uaid, chid)
-            return self._write_response(
-                301, 0, "Location: %s" % newUrl,
-                headers={"Location": newUrl})
+            return reply
```
bbangert: This is always a None value, since only the GCM check_token even calls self._write_response, and the return value of self._write_response is None.

bbangert:
r-, I think it's misleading to hand off control of output writing, then do something with the resulting return value as if there was meaning (return reply), when reply/writer is only ever a None value. The desired outcome for the GCM check_token call is to possibly interrupt processing to return a response in a certain case. It'd be more consistent with Router behavior to return a RouterResponse from check_token, and then "return self._router_response(reply)" from endpoint.py.

autopush/router/gcm.py
```diff
@@ -34,10 +34,27 @@ def __init__(self, ap_settings, router_conf):
...
         self.log.debug("Starting GCM router...")
 
-    def check_token(self, token):
+    def check_token(self, token, relo):
         if token not in self.senderIDs.senderIDs():
-            return (False, self.senderIDs.choose_ID().get('senderID'))
-        return (True, token)
+            # It is possible for some systems to allow the token to be
+            # changed, or a new token to be sent. At least one system
+            # historically allowed this. Unfortunately, that system no
```
kitcambridge: Could you please elaborate a bit in this comment? Which "one system historically allowed this?" What's a "mobile system"? (I don't have enough context to know if it's Fennec or the GCM router).

autopush/router/gcm.py
```diff
@@ -34,10 +34,27 @@ def __init__(self, ap_settings, router_conf):
...
+            # are codifying to return a "400", which is technically both
+            # correct, and incorrect, depending on your point of view.
+            #
+            # if relocation were enabled...
+            #
+            # token = self.senderIDs.choose_ID().get('senderID')
+            # relo_url = relo % token
+            # return writer(301, 0 "Location: %s" + relo_url,
+            #               headers={"Location": relo_url)
+            return False, dict(status_code=400,
```
kitcambridge: I'm wondering if this could throw a `RouterException` instead, and then this method wouldn't need to return anything. Or is that too magical?

kitcambridge:
Thanks! Sorry for the drive-by comments. 😄

autopush/router/gcm.py
jrconlin: I'd prefer if the function returned some value for reporting back. There can be instances where routers may want decidedly different behaviors or reports depending on how this value resolves.

bbangert:
r+

Issue #505, opened 2016-07-06
Increase GCM failure error level


jrconlin:
Currently logging GCM error replies as "debug", these should be higher so that they're logged and alert properly.

PR #506, opened 2016-07-06
bug: Elevate GCM reported error states

jrconlin:
Increase reporting priority for GCM errors so that they are logged to
Kibana. Made some messaging more unique and descriptive of actual case.

closes: #505 

@bbangert r?


jrconlin:
Adding a few explicit metric calls, since that should help ops.

bbangert:
r+

PR #513, opened 2016-07-13
Feature/issue 509

### LINKED ISSUE #509: Diagnosis CLI
We need a diagnosis CLI in autopush that lets us easily lookup full details for an endpoint URL, a UAID, and/or a UAID/CHID combo.

This CLI should let you put in a endpoint URL, and it'll then spit out:

1. UAID
2. CHID
3. Router record for UAID if found
4. Full CHID list for UAID and if the CHID is valid

bbangert:
Fixes a few bugs in existing tests that were using self.assert which failed to verify the value.

Fixes issue with the wpush endpoint not being registered in main.py for autoendpoint.

Add's the test_web_webpush which should've been included with the prior webpush endpoint refactor.

The last commit add's a endpoint_diagnostic CLI that will resolve an endpoint to its
UAID/CHID and router record info along with applicable valid channels.

Closes #509

@jrconlin r?

### DIFF
--- a/autopush/diagnostic_cli.py
+++ b/autopush/diagnostic_cli.py
@@ -1,0 +1,82 @@
+from __future__ import print_function
+
+import pprint
+import re
+import sys
+
+import configargparse
+from twisted.logger import Logger
+
+from autopush.main import (
+    add_shared_args,
+    shared_config_files,
+)
+from autopush.settings import AutopushSettings
+
+
+PUSH_RE = re.compile(r"push/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)")
+
+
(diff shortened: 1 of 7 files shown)


kitcambridge:
Nice! I wonder if it makes sense to add a little `__getattribute__` magic that fails a test if it tries to access `self.assert*`, so that this doesn't sneak in again in the future. On the other hand, we now know to look for it in code reviews, so maybe this is a case of policy instead of implementation.

bbangert:
I should note that there's still plenty of self.assert's in the code, these are just the ones that were failing that PyCharm picked up as it runs tox slightly differently. We should still replace the remainder so that detecting these aren't an accident.

jrconlin:
r+

PR #514, opened 2016-07-13
Feature/issue 433

### LINKED ISSUE #433: preflight check doesn't wait for table creation
On startup, autopush creates tables if needed. It should wait on their creation to complete so that the preflight check will run without error and the app can startup normally.

bbangert:
feat: wait for tables to be active in pre-flight check

On a fresh start, tables that are in the process of being created
may not be ready yet. We now wait for the tables to be active
and ready.

Includes a test commit fix for a time-out issue as well.

Closes #433

@jrconlin r?


jrconlin:
r+

PR #516, opened 2016-07-13
bug: remove shared file inclusion for tests

### LINKED ISSUE #515: fix local config file ignore for tests
Shared_config files crept back in to the unit tests. make sure they're segregated out from test runs.

jrconlin:
Parse-args can aggressively pull local config files. Tests are set to
exclude local files, but recent file changes added the shared_config
files back in causing local unit tests to fail.

Closes #515 

@bbangert r?


bbangert:
r+

PR #520, opened 2016-07-15
fix: log message errors as info instead of debug

### LINKED ISSUE #518: Switch 410 logging to info
We aren't logging many 410's because they're debug, they should be changed to info for logging.

bbangert:
Previously, several error classes were not logged as they were debug
level, not info level. This fix remedies that.

Closes #518 

@jrconlin r?


jrconlin:
r+

PR #521, opened 2016-07-15
bug: Invalidate UAIDs that are not lowercase or contain dashes

### LINKED ISSUE #519: UAID's supplied with dashes should be invalidated
Because /v1 and /v2 URL's all return uuid's with no dashes, old FF clients have endpoints generated that cannot be valid. As such we should invalidate the UAID upon connect if it has a dash in it, and treat it as invalid.

jrconlin:
Older UAIDs contain invalid characters and should be rejected.

Closes #519 

@bbangert r?


bbangert:
r+

PR #524, opened 2016-07-15
fix: refactor register uaid lookup to verify proper records

### LINKED ISSUE #523: Verify valid webpush record before considering it an existing user
Right now we just check to see if the UUID seems right for 'existing user'. We should instead do a router record read, and verify it has required attributes set before continuing.

One of the attributes we should check is that the current_month for the route record is in self.ps.settings.message_tables, and consider it an invalid record if not. This would let us remove the block of buggy code at https://github.com/mozilla-services/autopush/blob/1.15.1/autopush/websocket.py#L783 which fails to actually create a new record with the returned UAID.

bbangert:
When a client says hello, the register user method now performs a router lookup
to verify the record actually exists and has the appropriate necessary attributes.
A check is also done for webpush users to verify the server still has a table object
it can use to pull notifications for during verification so the uaid can be reset if
needed.

This cleans up the hello handling, removing the need for a separate hello finish
and simplifying  the callbacks.

Closes #523 

@jrconlin r?

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -84,10 +84,15 @@ class WebsocketTestCase(unittest.TestCase):
     def _send_message(self, msg):
         self.proto.onMessage(json.dumps(msg).encode('utf8'), False)
 
-    def _wait_for_message(self, d):
+    def _wait_for_message(self, d, count=0.0):
         args = self.send_mock.call_args_list
         if len(args) < 1:
-            reactor.callLater(0.1, self._wait_for_message, d)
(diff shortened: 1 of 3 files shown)


autopush/tests/test_websocket.py
```diff
@@ -484,6 +503,8 @@ def test_hello_tomorrow(self):
         def fake_msg(data):
             return (True, msg_data, data)
 
+        # Add mock return value for the router here!
```
jrconlin: todo?

jrconlin:
r+

PR #527, opened 2016-07-17
fix: ensure router_type is present in all records

### LINKED ISSUE #526: Ensure router_type is sent in register_user
The refactor of registering a user fails to include the router_type, but it must.

bbangert:
It was possible previously for a corrupted record to be missing a router_type
which caused AWS ValidationError's. This change ensures all records have
a router_type and connected_at which are required values for the register
user expression condition that AWS checks.

Closes #526 

@jrconlin r?


jrconlin:
r+

PR #532, opened 2016-07-18
bug: fix up AWS conditional binding and assoc. conditions

jrconlin:
AWS throws an exception if conditional args are not defined, this
impacted tests by failing, but throwing an exception and causing fun
havoc. register_user now fails if all conditionals are not met
(preserving previous behaviors). Added tests for older records that may
be incomplete.

Some formatting fixes included as well

@bbangert r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -18,6 +18,7 @@ from boto.dynamodb2.layer1 import DynamoDBConnection
 from boto.dynamodb2.table import Table
 from boto.dynamodb2.types import NUMBER
 
+from autopush.exceptions import AutopushException
 from autopush.utils import generate_hash
 
 key_hash = ""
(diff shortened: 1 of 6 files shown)


autopush/db.py
```diff
@@ -582,6 +582,9 @@ def register_user(self, data):
         conn = self.table.connection
         db_key = self.encode({"uaid": hasher(data["uaid"])})
         del data["uaid"]
+        if "router_type" not in data or "connected_at" not in data:
+            # Not specifying these values will generate an exception in AWS.
+            return (False, {}, data)
```
bbangert: Is this going to accidentally mask a call that thinks the condition is failing vs. being called with incorrect data?

autopush/tests/test_db.py
```diff
@@ -473,19 +474,24 @@ def raise_error(*args, **kwargs):
...
+        router.drop_user = Mock()
+        router.table.get_item.return_value = {"uaid": uuid.uuid4().hex}
         try:
             router.register_user(dict(uaid=uaid))
         except:
             pass
         self.assertRaises(ItemNotFound, router.get_uaid, uaid)
-        self.assertRaises(ItemNotFound, router.table.get_item,
-                          consistent=True, uaid=uaid)
+        assert(router.drop_user.called)
```
bbangert: ok_

autopush/tests/test_endpoint.py
```diff
@@ -624,7 +624,7 @@ def test_put_router_needs_change(self):
         def handle_finish(result):
             self.assertTrue(result)
             self.endpoint.set_status.assert_called_with(500, None)
-            assert(self.router_mock.register_user.called)
+            assert(not self.router_mock.register_user.called)
```
bbangert: ok_

bbangert:
r+ with 2 minor nits

PR #533, opened 2016-07-19
fix: remove use of buggy moto from websocket/db tests

### LINKED ISSUE #531: Update moto to return proper information
Two tests in websocket.py can fail due to odd heisenbugs in moto. In one case the Attributes response is a list, which is completely invalid per AWS. In the second case a conflict occurs for no apparent reason.

Moto is quite a few versions ahead of my fork, and has added some dynamodb features, but not the expression attributes we use, so my changes need to be merged back into the latest moto.

bbangert:
moto's testing harness is a bit buggy, it was returning errors in
a semi-random manner during test-runs and failing to do
validation checks that DynamoDB performs. As such we now
spin up the Java DynamoDB local server for these test files to
ensure consistent responses.

Closes #531 

@jrconlin r?

### DIFF
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -10,7 +10,6 @@ from boto.dynamodb2.exceptions import (
 from boto.dynamodb2.layer1 import DynamoDBConnection
 from boto.dynamodb2.items import Item
 from mock import Mock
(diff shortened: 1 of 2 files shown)


jrconlin:
r+

PR #536, opened 2016-07-19
fix: overwrite existing messages instead of throwing an error

### LINKED ISSUE #535: store_message should overwrite any existing records
Sentry Error:
```
ConditionalCheckFailedException: ConditionalCheckFailedException: 400 Bad Request
{u'message': u'The conditional request failed', u'__type': u'com.amazonaws.dynamodb.v20120810#ConditionalCheckFailedException'}
```

bbangert:
DynamoDB throws a condition check failure if you put_item and
it already exists. This can occur when we save a message that
may already exist when a user hops servers without ack'ing. We
should allow it since it consumes the same resources.

Closes #535 

@jrconlin r?

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -411,7 +411,7 @@ class Message(object):
         if data:
             item["headers"] = headers
             item["data"] = data
-        self.table.put_item(data=item)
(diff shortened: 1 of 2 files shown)


jrconlin:
r+

PR #537, opened 2016-07-20
test: don't hide wait_for_times test failures

pjenvey:
was probably unintentional in 874f0757


jrconlin:
r+
@bbangert want to verify?

bbangert:
r+

PR #539, opened 2016-07-20
feat: Add FCM router support

### LINKED ISSUE #517: Add FCM router support
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1286912, there are a few changes that need to be done to move from GCM to FCM, one of which is to change the endpoint the server calls from 
http.googleapis.com/gcm/ to fcm.googleapis.com/fcm/

We'll want to create a new router type of "fcm", that points to the new endpoint URL. This new router can mostly be a clone of the GCM one.

jrconlin:
Adds 'fcm' as a client selectable router, to allow 'gcm' to persist or
act as a fallback.

closes #517 

@bbangert r?

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -165,6 +165,25 @@ def add_external_router_args(parser):
                         env_var="GCM_COLLAPSEKEY")
     parser.add_argument('--senderid_list', help='SenderIDs to load to S3',
                         type=str, default="{}")
+    # FCM
+    parser.add_argument('--fcm_enabled', help="Enable FCM Bridge",
+                        action="store_true", default=False,
+                        env_var="FCM_ENABLED")
+    label = "FCM Router:"
+    parser.add_argument('--fcm_ttl', help="%s Time to Live" % label,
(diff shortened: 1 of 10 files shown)


bbangert:
Moar reviewers...

pjenvey:
There's just so little difference between this & GCM's router (just the constructor) I'd suggest consolidating them a bit somehow. Might they diverge much in the future?

jrconlin:
Yeah, I thought about combining them, except we hit problems with doing something like that in the past. The way I look at it is that the "GCM" router would be locked while any future changes would hit the "FCM" router. This would also mean that older clients (which may or may not update to use FCM) could continue for a while. Plus, as I note in a few locations, there's a LOT more work the client has to do to use FCM. This, at least, lets clients determine when they want to switch over without adding undue complexity to our code.  

Fortunately, these routers pretty much just hand off data. Still, I definitely understand the sentiment.

bbangert:
r+, not a fan of the duplication, but in theory we lose the GCM at some point and FCM can improve independently (or hopefully we just use SNS).

PR #561, opened 2016-07-27
Bug: Fix webpush to reject endpoints which have an invalid GCM senderid

### LINKED ISSUE #556: Return 410 on GCM message delivery failure due to SenderIDMismatch
Since Jul, 16, 2016 we've seen a spike in GCM delivery failure messages (e.g.
```
GCM failures: [["eHpQ5cm5hdI:APA91bG-...YLfBv", "MismatchSenderId"]]
```
The SenderID we use on the server and client has not changed and OPs has verified that it matches the ID provided by the Google Developer Console. In addition, only older installs seem to be impacted.

Turns out, that the solution is to uninstall the effected version of Firefox, and reinstall so that the SenderID on the application can be reinitialized. (I have personally proven that this approach works.) Updating the application or restarting does NOT appear to solve the issue. It's my understanding that currently GCM SenderIDs for applications are established in the manifest file and may be stored by the OS on install. I do not know if it's currently possible for applications to reset or re-assert their SenderIDs after install.

Complicating matters, Clients do not frequently check with the server once a GCM bridged push connection is established. We rely on GCM being stable to deliver content. There is, however, one event that we may be able to use to indicate to the remote client that there is a problem with their Push bridge. When a client wishes to create a new channel, it sends an HTTP registration request to the server, and a reply message containing the prospective endpoint is returned. We could piggy-back an error code to the client in that message (along with not returning an endpoint) which could allow the client to either request a re-install, or try some other mechanism to re-register their SenderID. 

For now, a failed GCM message should now return a 410, indicating that the client is no longer available. It is to be determined if additional information should be relayed to the subscription provider so that further measures could be taken by the Subscription service as well.

jrconlin:
Adds logic to fail endpoints with invalid SenderIDs so that they fail fast on
post. Reduced logging level of GCM SenderID errors, since there's going
to be a lot of those for a while.

Closes #556 

@bbangert @pjenvey r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -53,6 +53,7 @@ from autopush.utils import (
     base64url_encode,
 )
 from autopush.web.base import DEFAULT_ERR_URL
+from autopush.websocket import ms_time
 
 
 # Our max TTL is 60 days realistically with table rotation, so we hard-code it
@@ -581,7 +582,7 @@ class EndpointHandler(AutoendpointHandler):
         # Were we told to update the router data?
         if response.router_data:
             uaid_data["router_data"] = response.router_data
-            uaid_data["connected_at"] = int(time.time() * 1000)
+            uaid_data["connected_at"] = ms_time()
             d = deferToThread(self.ap_settings.router.register_user,
                               uaid_data)
             response.router_data = None
@@ -766,7 +767,7 @@ class RegistrationHandler(AutoendpointHandler):
             uaid=self.uaid,
             router_type=router_type,
             router_data=router_data,
-            connected_at=int(time.time() * 1000),
+            connected_at=ms_time(),
             last_connect=generate_last_connect(),
         )
         return deferToThread(self.ap_settings.router.register_user, user_item)
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -6,6 +6,7 @@ from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
 
 from autopush.router.interface import RouterException, RouterResponse
+from autopush.websocket import ms_time
 
 
 class GCMRouter(object):
@@ -24,6 +25,7 @@ class GCMRouter(object):
         self.senderIDs = router_conf.get("senderIDs")
         self.metrics = ap_settings.metrics
         self._base_tags = []
+        self.router_table = ap_settings.router
         try:
             sid = self.senderIDs.keys()
             self.senderID = sid[0]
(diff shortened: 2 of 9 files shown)


autopush/router/gcm.py
```diff
@@ -138,10 +142,18 @@ def _process_reply(self, reply):
...
-            self.log.critical("GCM failures: {failed()}",
-                              failed=lambda: json.dumps(reply.failed.items()))
-            raise RouterException("GCM failure to deliver", status_code=503,
-                                  response_body="Please try request later.")
+            self.log.info("GCM failures: {failed()}",
+                          failed=lambda: json.dumps(reply.failed.items()))
+            self.router_table.register_user(
+                {"uaid": uaid_data.get('uaid'),
+                 "router_type": uaid_data.get("router_type", "gcm"),
+                 "connected_at": time.time(),
```
bbangert: Shouldn't this be the ms_time function?

jrconlin: Huh, we don't seem to be consistent about that. A quick grep shows that we've got time.time() in endpoint, ms_time() in websocket, I'll correct all of the instances to use ms_time.

autopush/router/gcm.py
```diff
@@ -43,18 +45,20 @@ def register(self, uaid, router_data, router_token=None, *kwargs):
         if "token" not in router_data:
             raise self._error("connect info missing GCM Instance 'token'",
                               status=401)
+        if not (router_token == router_data['token'] == self.senderID):
```
bbangert: I don't really understand why the router_token needs to match the self.senderID, maybe a link to some GCM docs or something highlighting why this needs to match here? Does this also imply that every single GCM client stores a token in our DB that matches a hard-coded value of self.senderID?

autopush/router/gcm.py
```diff
@@ -138,10 +142,18 @@ def _process_reply(self, reply):
...
+            self.log.info("GCM failures: {failed()}",
+                          failed=lambda: json.dumps(reply.failed.items()))
+            self.router_table.register_user(
+                {"uaid": uaid_data.get('uaid'),
+                 "router_type": uaid_data.get("router_type", "gcm"),
+                 "connected_at": time.time(),
+                 "critical_failure": "Client is unreachable due to a "
+                                     "configuration error. Unable to "
+                                     "send message.",
+                 })
```
bbangert: Why are we updating the user record every time we have a config error?

autopush/router/gcm.py
jrconlin: It's a bit complex and a hang-over from when we were going to use multiple senderids to deliver content. That got dropped, but it's a useful check for situations such as this. "Router_Token" is a client side, hard coded. value that is the GCM SenderID/Project Number. It needs to match much in the same way as our UAID needs to match. I could pare the value out of the router data if that's a concern.

autopush/router/gcm.py
bbangert: Ok, I see later on where that's used to halt processing before hitting this again. But why not just delete the user record entirely?

jrconlin: We're not, just the first time we hit this error. It allows us to later not go through the effort of even trying a record we know will fail.

jrconlin: We also may need to report the error to the subscriber, thus the amended record. Otherwise the subscriber might wonder why their client can subscribe but the server refused the record as "missing".

autopush/router/gcm.py
bbangert: The time.time calls for connected_at in endpoint.py all end up with ms_time since they multiply by 1000 though. This one doesn't.

jrconlin: Right, but we should be consistent to avoid problems like this.

bbangert: Agreed, there's prolly a call to replace in web/webpush.py as well.

autopush/endpoint.py
```diff
@@ -53,6 +53,7 @@
     base64url_encode,
 )
 from autopush.web.base import DEFAULT_ERR_URL
+from autopush.websocket import ms_time
```
jrconlin: I will leave moving ms_time() to utils for a future patch.

bbangert:
r+

Issue #553, opened 2016-07-25
Webpush preflight check doesn't catch ItemNotFound and throw RouterException


bbangert:
Per sentry issue:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/341729/

bbangert:
Closed by #566.

PR #566, opened 2016-07-28
fix: avoid uaid error in webpush preflight

### LINKED ISSUE #542: Don't call get_uaid twice for webpush
webpush.py preflight_check does a get_uaid call, this is unnecessary since we already called it before routing.

bbangert:
Avoids the uaid error on fetch in the preflight by re-using the existing fetched
data. If the current_month doesn't exist, this is an invalid record so the uaid
is removed as it should've been.

closes #542, #553 

@jrconlin, @pjenvey r?

### DIFF
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -64,7 +64,7 @@ class SimpleRouter(object):
     def amend_msg(self, msg, router_data=None):
         return msg
 
-    def preflight_check(self, uaid, channel_id):
+    def preflight_check(self, uaid_data, channel_id):
         """Verifies this routing call can be done successfully"""
         return True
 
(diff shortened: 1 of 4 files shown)


autopush/router/simple.py
```diff
@@ -64,7 +64,7 @@ def register(self, uaid, connect, *kwargs):
     def amend_msg(self, msg, router_data=None):
         return msg
 
-    def preflight_check(self, uaid, channel_id):
+    def preflight_check(self, uaid_data, channel_id):
```
jrconlin: +💯

jrconlin:
+r

Really wonder if it might be worth auditing parameters to see where we can simplify to the dict.

autopush/router/webpush.py
```diff
@@ -60,21 +60,20 @@ def _crypto_headers(self, notification):
...
+            yield deferToThread(self.ap_settings.router.drop_user, uaid)
             raise RouterException("No such subscription", status_code=410,
                                   log_exception=False, errno=106)
 
-        month_table = record["current_month"]
+        month_table = uaid_data["current_month"]
         if month_table not in self.ap_settings.message_tables:
             self.log.info("'current_month' out of scope: {record}",
-                          records=json.dumps(record))
+                          records=json.dumps(uaid_data))
```
pjenvey: While you're here, records= is broken (per {record}) -- maybe change it to uaid instead (if so, "UAID missing .." above too)? channelid= a few lines below is also broken

PR #567, opened 2016-07-28
bug: Do not record the GCM data overflow error

### LINKED ISSUE #552: Catch RouterException from gcm
Per exception:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/341792/

It appears the router exception is being reported when it shouldn't be.

jrconlin:
Return the GCM data overflow error to the subscription server, but
there's no need to record it into Sentry.

Closes #552 

@bbangert , @pjenvey r?


bbangert:
r+

pjenvey:
might this be applicable to FCM too?

PR #569, opened 2016-07-28
bug: Add ValueError trap for extract_jwt

### LINKED ISSUE #562: Validation fails to catch ValueError in extract_jwt
Per https://sentry.prod.mozaws.net/operations/autopush-prod/issues/342238/ the new validation.py is not catching a ValueError bubbled up by decipher_public_key from the extract_jwt call.

```
ValueError: Unknown public key format specified
  File "twisted/python/threadpool.py", line 246, in inContext
    result = inContext.theWork()
  File "twisted/python/threadpool.py", line 262, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "autopush/web/validation.py", line 59, in _validate_request
    return schema.load(data)
  File "marshmallow/schema.py", line 542, in load
    result, errors = self._do_load(data, many, partial=partial, postprocess=True)
  File "marshmallow/schema.py", line 646, in _do_load
    result = self._invoke_load_processors(POST_LOAD, result, many, original_data=data)
  File "marshmallow/schema.py", line 767, in _invoke_load_processors
    data=data, many=many, original_data=original_data)
  File "marshmallow/schema.py", line 865, in _invoke_processors
    data = utils.if_none(processor(data), data)
  File "autopush/web/validation.py", line 313, in fixup_output
    self.validate_auth(d)
  File "autopush/web/validation.py", line 299, in validate_auth
  File "autopush/utils.py", line 132, in extract_jwt
    key = decipher_public_key(crypto_key)
  File "autopush/utils.py", line 123, in decipher_public_key
    raise ValueError("Unknown public key format specified")
```

jrconlin:
A bad or improperly formatted VAPID object may generate a ValueError.
While important, and notable to the originator, this does not need to go
into Sentry.

Also fixes some logging formatting errors

closes #562 

@bbangert, @pjenvey r?

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -215,9 +215,11 @@ class AutoendpointHandler(ErrorLogger, cyclone.web.RequestHandler):
                 self.set_header(header, headers.get(header))
         self.finish()
 
-    def _write_unauthorized_response(self, message="Invalid authentication"):
+    def _write_unauthorized_response(self, message="Invalid authentication",
+                                     **kwargs):
         headers = {"www-authenticate": AUTH_SCHEME}
-        self._write_response(401, errno=109, message=message, headers=headers)
+        self._write_response(401, errno=109, message=message, headers=headers,
+                             **kwargs)
 
     def _response_err(self, fail):
         """errBack for all exceptions that should be logged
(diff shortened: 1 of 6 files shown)


autopush/endpoint.py
```diff
@@ -336,7 +336,8 @@ def _invalid_auth(self, fail):
         if isinstance(fail.value, VapidAuthException):
             raise fail.value
         message = fail.value.message or repr(fail.value)
-        if isinstance(fail.value, AssertionError):
+        if (isinstance(fail.value, AssertionError) or
+                isinstance(fail.value, ValueError)):
```
pjenvey: Is "A decryption error occurred" the right message for this ValueError?

jrconlin: There are a number of reasons that either exception could be raised, that would be difficult to relay. The ValueError in the exception is due to an bad JWT key, for instance. The message is already pointing that the bearer token (and thus VAPID) is failing. It may be worth pointing them toward documentation, though.

pjenvey:
r+

bbangert:
r+

Issue #565, opened 2016-07-28
Update Autopush install documentation


jrconlin:
Note ALL requirements (TLS, configurations, etc.) for setting up a test server. Presume no prior knowledge, but add links to relevant documentation when required.

PR #570, opened 2016-07-28
docs: updates to intros

pjenvey:
Some of #565 

@bbangert, @jrconlin r?


bbangert:
r+ :100:

docs/install.rst
```diff
@@ -33,7 +33,7 @@ For instance, if installing on an Amazon EC2 machine:
     libncurses5-devel openssl-devel patch python-devel \
     python-virtualenv readline-devel -y
 
-Autopush uses the `Boto python library <https://boto.readthedocs.io/en/latest/>`_. Be sure to `properly set up <https://boto.readthedocs.io/en/latest/boto_config_tut.html>`_ your ``.boto`` configuration file.
+Autopush uses the `Boto python library <https://boto.readthedocs.io/en/latest/>`_. Be sure to `properly set up <https://boto.readthedocs.io/en/latest/boto_config_tut.html>`_ your boto configuration file.
```
jrconlin: While I appreciate that ``.boto`` is a bit opaque, do you think it might be useful to remind folks that they need to set up the config file in their home directory as ``.boto``. 

People are often gleefully illiterate.

docs/running.rst
```diff
@@ -33,8 +33,48 @@ To setup the server locally:
     $ mkdir ddb
     $ curl -sSL http://dynamodb-local.s3-website-us-west-2.amazonaws.com/dynamodb_local_latest.tar.gz | tar xzvC ddb/
     $ java -Djava.library.path=./ddb/DynamoDBLocal_lib -jar ./ddb/DynamoDBLocal.jar -sharedDb -inMemory
-    $ cp automock/boto.cfg ~/.boto
 
-Note the last line copies a boto config over ``~/.boto`` in your home dir. If
-you have existing AWS Credentials in this file, you should move it elsewhere
-first before running ``autopush`` and ``autoendpoint``.
+An example `boto config file`_ is provided in ``automock/boto.cfg`` that
```
jrconlin: Huh, wonder if we should copy this to ./configs, since that's where we've collected other, similar config files. Probably ought to comment the file as well. (Not that it's helped in some cases, but it's consistent)

docs/running.rst
```diff
@@ -33,8 +33,48 @@ To setup the server locally:
...
+
+    ; Uncomment to provide stdout logging in a more human readable format.
+    #human_logs
+
+Running the nodes
+=================
+
+.. code-block:: bash
+
+    $ BOTO_CONFIG=automock/boto.cfg bin/autopush --config-shared=./autopush_shared.ini
```
jrconlin: hrm. configs are weird. 

We have an app specific file, plus the shared. We also default to reasonably well known locations. It seems odd that we're using the CLI to specify just one config, but I agree that specifying both could be verbose and potentially confusing. I'll file a bug to consider simplifying this by adding a `config_dir` arg or something.

jrconlin:
minor edits, but nothing that's a show-stopper. 

r+

docs/install.rst
bbangert: I think the reason for omitting it is good, which is that .boto is just one of several locations you config may reside. It could be in /etc as well. We should ensure people are aware of the various places boto config may be by pointing them to the config docs.

docs/running.rst
bbangert: There are 5 possible config locations unfortunately, if supplying BOTO_CONFIG up front forces just that one to be used I think documenting that as the best way is ideal. I know I've run into issues more than once because I forgot about a config file somewhere that boto grabbed.

jrconlin: Yeah, I agree being explicit about the BOTO_CONFIG is good, because we don't control the boto library. I was more thinking about the three potential locations for our own config files.

pjenvey: BOTO_CONFIG does indeed override the other locations. This is more of a quickstart so the simpler BOTO_CONFIG is easier and added benefit of not clashing with the test runner's env if you're so inclined to run them.

Otherwise it's a more generic description "you need some kind of boto config or AWS env credentials" -- if you're deploying to AWS you can decide whichever of those works better for your setup

PR #573, opened 2016-07-29
bug: Fix up FCM library to reject enpdoints with invalid FCM senderids

### LINKED ISSUE #556: Return 410 on GCM message delivery failure due to SenderIDMismatch

jrconlin:
Closes #556 

(picks up spare from https://github.com/mozilla-services/autopush/pull/561)

@bbangert @pjenvey r?

### DIFF
--- a/autopush/router/fcm.py
+++ b/autopush/router/fcm.py
@@ -50,6 +50,14 @@ class FCMRouter(object):
         if "token" not in router_data:
             raise self._error("connect info missing FCM Instance 'token'",
                               status=401)
+        # router_token and router_data['token'] are semi-legacy from when
+        # we were considering having multiple senderids for outbound
+        # GCM support. That was abandoned, but it is still useful to
+        # ensure that the client's senderid value matches what we need
+        # it to be. (If the client has an unexpected or invalid SenderID,
+        # it is impossible for us to reach them.
+        if not (router_token == router_data['token'] == self.senderID):
+            raise self._error("Invalid SenderID", status=410, errno=105)
         # Assign a senderid
         router_data["creds"] = {"senderID": self.senderID, "auth": self.auth}
         return router_data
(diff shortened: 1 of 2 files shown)


autopush/router/fcm.py
```diff
@@ -50,6 +50,8 @@ def register(self, uaid, router_data, router_token=None, *kwargs):
         if "token" not in router_data:
             raise self._error("connect info missing FCM Instance 'token'",
                               status=401)
+        if not (router_token == router_data['token'] == self.senderID):
```
pjenvey: You may or may not want to copy the associated comment over from gcm.py, up to you

pjenvey:
r+

bbangert:
r+

PR #574, opened 2016-07-30
Update dev endpoint

wh0:
Was this move documented somewhere? updates-autopush-dev.stage.mozaws.net doesn't resolve anymore.

URLs from this page work https://wiki.mozilla.org/QA/Push_Notifications


jrconlin:
r+

PR #576, opened 2016-08-01
bug: Do not remove router_type from UAID info

### LINKED ISSUE #549: Ensure router_type is present in _router_completed
A previous PR fixed the return handing in _router_completed, this fix was not applied to the _router_completed in web/webpush.py however.

jrconlin:
Webpush web handler should not remove router_type from UAID information.

Also fixed error message formatting and added appropriate test to ensure
that router_type persists.

closes #549 

@bbangert, @pjenvey r?


pjenvey:
r+

bbangert:
I don't get it, why doesn't this match how _router_completed in endpoint.py works?

PR #577, opened 2016-08-02
fix: Normalize endpoint and web _router_completed calls and tests

### LINKED ISSUE #549: Ensure router_type is present in _router_completed

jrconlin:
closes #549 

@bbangert r?


bbangert:
Which way is it supposed to be? The endpoint.py change to remove the del statement was done first, so I had thought the endpoint.py change was going to propagate to webpush.py, not vice versa. Is it supposed to delete the router_data if its not set, or not?

bbangert:
r+

autopush/router/fcm.py
```diff
@@ -45,18 +45,19 @@ def amend_msg(self, msg, data=None):
...
-        # ensure that the client's senderid value matches what we need
-        # it to be. (If the client has an unexpected or invalid SenderID,
+        # senderid is the remote client's senderID value. This value is
+        # very difficult for the client to change, and there was a problem
+        # where some clients had an older, invalid senderID. We need to
+        # be able to match senderID to it's corresponding auth key.
+        # If the client has an unexpected or invalid SenderID,
         # it is impossible for us to reach them.
-        if not (router_token == router_data['token'] == self.senderID):
+        if not (senderid == self.senderID):
```
jrconlin: router['token'] is the registrationID and should NOT match senderid. I'm an idiot.

autopush/web/webpush.py
```diff
@@ -56,13 +56,21 @@ def post(self, api_ver="v1", token=None):
...
+                # be deleted. There is no longer valid route information for
+                # this record.
+                self.ap_settings.router.drop_user(self.uaid)
+                return self._router_response(response)
+            # The router data needs to be updated to include any changes
+            # requested by the bridge system
+            uaid_data["router_data"] = response.router_data
+            # set the AWS mandatory data
             uaid_data["connected_at"] = ms_time()
+            rt = self.valid_input.get(
```
jrconlin: Aye! Dig! Dig ya miserable bastard! There be a router_type down there!

autopush/endpoint.py
```diff
@@ -584,11 +584,21 @@ def _route_notification(self, version, result, data, ttl=None):
...
         # TODO: Add some custom wake logic here
-
         # Were we told to update the router data?
-        if response.router_data:
+        if response.router_data is not None:
+            if not response.router_data:
+                # An empty router_data object indicates that the record should
+                # be deleted. There is no longer valid route information for
+                # this record.
+                self.ap_settings.router.drop_user(self.uaid)
```
pjenvey: I think we need to deferToThread drop_user

jrconlin: I was wondering about that. In theory, we're already in a deferred thread from _route_notification.
I'll add a defer anyway.

autopush/router/gcm.py
```diff
@@ -12,7 +12,7 @@
 class GCMRouter(object):
     """GCM Router Implementation"""
     log = Logger()
-    gcm = None
+    gcm = {}
```
pjenvey: We only instantiate one GCMRouter per app it looks like, but let's avoid this evil mutable class variable regardless

jrconlin:
ready for re-review

@bbangert @pjenvey r?

autopush/router/gcm.py
```diff
@@ -22,39 +21,45 @@ def __init__(self, ap_settings, router_conf):
         self.min_ttl = router_conf.get("ttl", 60)
         self.dryRun = router_conf.get("dryrun", False)
         self.collapseKey = router_conf.get("collapseKey", "simplepush")
-        self.senderIDs = router_conf.get("senderIDs")
+        self.gcm = {}
+        self.senderIDs = {}
```
pjenvey: You could actually combine the gcm and senderIDs dicts into one at this point, {sid: (auth, GCM()),} .. but wait a second..

pjenvey:
r+

autopush/router/gcm.py
jrconlin: well, really, I could create a map of {sid: GCM()} Holding onto the auth key is probably not needed. I divided things up for now mostly to keep it clear (at least for me). We can collapse things down if we need to.

autopush/web/webpush.py
bbangert: Why not just record the router_type on self from the post method above that already dug it out?

bbangert: Also, the valid_input was guaranteed to have a router_type and subscription attribute already, the benefit of validation up front means we aren't guessing if it exists and using get calls like this.

autopush/web/webpush.py
```diff
@@ -56,13 +56,23 @@ def post(self, api_ver="v1", token=None):
...
+                d.addCallback(lambda x: self._router_response(response))
+                return d
+            # The router data needs to be updated to include any changes
+            # requested by the bridge system
+            uaid_data["router_data"] = response.router_data
+            # set the AWS mandatory data
             uaid_data["connected_at"] = ms_time()
+            rt = self.valid_input.get(
+                'subscription', {}).get('user_data', {}).get('router_type')
+            uaid_data["router_type"] = uaid_data.get("router_type", rt)
```
bbangert: What's the rationale for a router being able to change the router type?

bbangert:
r-, need some clarification on what's occuring with router type resets, and get calls that aren't needed.

autopush/web/webpush.py
jrconlin: We're setting it if it's not already set.

bbangert: How would we get to this point if it wasn't set? The validation logic will toss out the request if there's no router_type before starting this process.

bbangert:
r+ fe0d19c

PR #580, opened 2016-08-02
docs: updates to installation

pjenvey:
The Makefile assumes PyPy and we're going in that direction so defaulting to it


pjenvey:
@bbangert @jrconline r?

bbangert:
r+

jrconlin:
r+

PR #581, opened 2016-08-02
fix: readline isn't needed

### LINKED ISSUE #558: Remove the gnureadline dependency
AFAICT this isn't needed. If it was ever needed it was likely only for OS X -- but even stock OS X python includes a limited libedit based readline module. That stock python isn't an option for autopush anyway because it's a UCS2 build.

pjenvey:
closes #558 

r? @bbangert


bbangert:
r+

PR #584, opened 2016-08-04
refactor: clarify intent of write methods returning nothing

pjenvey:
these async write calls don't return anything so this makes the intent more clear, especially amid callbacks

r? @bbangert @jrconlin


jrconlin:
r+

bbangert:
r+

PR #590, opened 2016-08-09
bug: Update FCM handler to more accurately reflect API

### LINKED ISSUE #589: Add FCM library support
FCM can return status as a JSON object which may not be fully readable by the gcmclient library. 
Recommend using or creating a FCM compliant library for the FCM router.

jrconlin:
FCM, while compatible with GCM, has a much richer error handler. We
should also use a fcm specific bridging library.

closes #589 

@pjenvey r?

### DIFF
--- a/autopush/router/fcm.py
+++ b/autopush/router/fcm.py
@@ -1,6 +1,5 @@
 """FCM Router"""
-import gcmclient
-import json
+import pyfcm
 
 from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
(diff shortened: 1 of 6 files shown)


autopush/router/fcm.py
```diff
@@ -32,8 +31,87 @@ def __init__(self, ap_settings, router_conf):
         self.auth = router_conf.get("auth")
         self.metrics = ap_settings.metrics
         self._base_tags = []
+        self.creds = {}
+        self.regid = ""
+        self.reasonTable = {
```
pjenvey: I'd prefer reasonTable as a global var, or as a class var like APNSRouter.errors since it's a big chunk of immutable data

autopush/router/fcm.py
```diff
@@ -32,8 +31,87 @@ def __init__(self, ap_settings, router_conf):
...
+                "msg": ("'to' or 'registration_id' is blank or"
+                        " invalid: {regid}"),
+                "err": 500,
+                "errno": 1,
+            },
+            "InvalidRegistration": {
+                "msg": "registration_id is invalid: {regid}",
+                "err": 410,
+                "errno": 105,
+                "crit": True,
```
pjenvey: I'd argue these are more error log level vs critical as they're all somewhat expected. fcm.py already has 1 log.critical call for one case, I'll argue that that one switch to error as well.

autopush/router/fcm.py
```diff
@@ -72,11 +151,17 @@ def route_notification(self, notification, uaid_data):
...
+            raise self._error("No registration token found. "
+                              "Rejecting message.",
+                              410, errno=106, log_exception=False)
+        self.regid = router_data.get("token")
         # Payload data is optional. The endpoint handler validates that the
         # correct encryption headers are included with the data.
         if notification.data:
             mdata = self.config.get('max_data', 4096)
-            if len(notification.data) > mdata:
+            self.notifLen = len(notification.data)
```
pjenvey: PEP8 notifLen, and assigning it to self isn't necessary

autopush/router/fcm.py
```diff
@@ -96,79 +181,74 @@ def _route(self, notification, router_data):
...
+                )
+                raise RouterException("FCM failure to deliver",
+                                      status_code=err['err'],
+                                      response_body="Please try request "
+                                                    "later.")
+            self.log.info("{msg} : {info}",
+                          msg=err['msg'],
+                          info={"senderid": self.creds.get('registration_id'),
+                                "reason": reason,
+                                "error": err['msg']})
```
pjenvey: this is redundant w/ msg=err['msg'] a few lines above

autopush/router/fcm.py
jrconlin: I'll agree about InvalidRegistration & DeviceMessageRateExceeded (since I missed those when I was cleaning up) but the other ones imply something seriously wrong with either our code or that the FCM endpoint has changed. I'm willing to say those are critical.

autopush/router/fcm.py
```diff
@@ -96,79 +181,74 @@ def _route(self, notification, router_data):
...
+                dry_run=self.dryRun or ('dryrun' in router_data),
+                registration_id=self.regid,
+                time_to_live=self.router_ttl,
+            )
+        except pyfcm.errors.AuthenticationError as e:
             raise self._error("Authentication Error: %s" % e, 500)
         except Exception as e:
             raise self._error("Unhandled exception in FCM Routing: %s" % e,
                               500)
         self.metrics.increment("updates.client.bridge.gcm.attempted",
```
pjenvey: Might as well switch this to ..bridge.fcm.attempted while you're here?

autopush/router/fcm.py
jrconlin: Crap. yeah, leftover from a previous thought. I can revert this back and drop notifLen entirely, since I'm passing notification to _process_reply()

autopush/router/fcm.py
```diff
@@ -32,8 +31,87 @@ def __init__(self, ap_settings, router_conf):
         self.auth = router_conf.get("auth")
         self.metrics = ap_settings.metrics
         self._base_tags = []
+        self.creds = {}
```
pjenvey: I'd like to kill self.creds and self.regid because they get mutated during register/route calls. We can explicitly pass the creds/regid from router_data to _process_reply (or just pass the whole router_data dict itself), then the vars on self won't be necessary

pjenvey: Similar comment for self.router_ttl -- it's really just a local variable used in the _route method -- so a self.router_ttl is not needed

pjenvey:
r+

PR #591, opened 2016-08-09
feature: better documentation for config files & locations

### LINKED ISSUE #572: Simplify config system
Right now we have three config files (four if you include boto). Two are specific to the type of app instance we're starting and one is shared. While we keep the files in reasonably well known locations, there can be some confusion because the names for the config file name changes subtly (e.g. ./configs/autopush_shared.ini, ~/.autopush_shared.ini, /etc/autopush_shared.ini)

We may want to include a "--config_dir" option, and rename the ./config files as *.sample or something.

jrconlin:
Add documentation about the value of the configuration files and
where they can be located on the system.

Closes #572 

@pjenvey r?


pjenvey:
Did you git mv the configs? I don't see the .sample files here (they weren't staged?)

docs/install.rst
```diff
@@ -56,6 +56,58 @@ to specify the ``-p <path to pypy>`` option):
...
+line. if an unexpected option is present in a configuration file, the application
+will fail to start.
+
+Configuration files can be located in:
+
+* in the /etc/ directory
+
+* in the $HOME or current directory (prefixed by a period '.')
+
+* in the configs subdirectory
```
pjenvey: Mention configs before $HOME/current since that's the order they're checked

pjenvey:
Also there's a lingering reference to configs/autopush_shared.ini in running.rst

jrconlin:
Weird, I did git mv the configs. Might have gotten stomped on the rebase.

pjenvey:
r+

PR #593, opened 2016-08-10
feat: Allow both "Bearer" and "WebPush" as Auth tokens

### LINKED ISSUE #592: VAPID auth scheme name is now "WebPush", not "Bearer"
I raised this in webpush-wg/webpush-vapid#7 a few months ago, and we decided to keep `Bearer` then...but that's since changed in  webpush-wg/webpush-vapid#18. The correct scheme name is now `WebPush` (though we'll probably want to allow both for back-compat).

See also: web-push-libs/web-push#200

jrconlin:
VAPID spec recently changed to specify "WebPush" as valid Authorization
token ID. Code allows either Bearer or WebPush while external code
transitions.

Closes #592 

@pjenvey, @kitcambridge r?


pjenvey:
r+

Issue #595, opened 2016-08-10
Remove s3/misc remnants


pjenvey:
We forgot to remove a few obsoleted config options from the sample config files (the s3 options and a couple others)

We can also remove the mock_s3 calls from tests.

PR #596, opened 2016-08-11
fix: remove obsoleted options from sample config files, and mock_s3 w…

pjenvey:
kill some old stuff, #595

r? @jrconlin


jrconlin:
Huh, for some reason, I thought I had dropped these earlier. Thanks!

r+

PR #600, opened 2016-08-12
 feat: Validate Encryption/C-Key headers in preflight

### LINKED ISSUE #456: Validate crypto headers
@jrconlin brings up a great point in https://github.com/mozilla-services/autopush/pull/452#issuecomment-215485775:

> They already have several significant hurdles to overcome (encryption, user data management, scaling) and then us rejecting their submission because they included characters that literally signify nothing would certainly put me into a foul mood.

We can also help developers by rejecting invalid or incomplete `Encryption` (and `Content-Encoding`) headers, like we already do for `Crypto-Key`. May as well reject early with an informative error instead of storing and delivering a message that the client will drop immediately.

jrconlin:
Check the encryption headers to make sure they're roughly valid before
passing them on to the client.

Closes #456

@pjenvey r?

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -219,7 +219,7 @@ class EndpointTestCase(unittest.TestCase):
         frouter.route_notification = Mock()
         frouter.route_notification.return_value = RouterResponse()
         self.endpoint.chid = dummy_chid
-        self.request_mock.headers["encryption"] = "stuff"
+        self.request_mock.headers["encryption"] = "salt=stuff"
         self.request_mock.headers["content-encoding"] = "aes128"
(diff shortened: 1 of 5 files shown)


autopush/tests/test_endpoint.py
```diff
@@ -697,11 +734,12 @@ def handle_finish(result):
             self.endpoint.set_status.assert_called_with(400, None)
 
         self.finish_deferred.addCallback(handle_finish)
+        # Wait, is this dropping to simplepush!?
```
pjenvey: is this comment still needed?

pjenvey:
I'm wondering if that's a stray comment in test_endpoint but
r+

PR #601, opened 2016-08-12
docs: describe configuration of SSL/TLS

pjenvey:
This is probably enough for #565, for now

r? @jrconlin

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -65,6 +65,43 @@ You may also want to enable human readable logs in the shared config file:
     ; Uncomment to provide stdout logging in a more human readable format.
     #human_logs
 
+Configuring SSL/TLS
+-------------------
+
+The Push protocol mandates use of SSL/TLS connections to ensure the integrity
+of push subscription data. This rule can be ignored for testing purposes
+only. This requires a valid, trusted TLS certificate (for example, from `Let's
+Encrypt`_).
+
+The ``ssl_key`` and ``ssl_cert`` options configure a TLS certificate for
+endpoint's HTTPS and or connection node's WebSocket Secure (WSS) connections:
+
+.. code-block:: ini
+
+    ; TLS private key and certificate file paths, used for the main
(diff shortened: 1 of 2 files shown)


docs/running.rst
```diff
@@ -65,6 +65,39 @@ You may also want to enable human readable logs in the shared config file:
...
+-------------------
+
+The Push protocol mandates use of SSL/TLS connections to ensure the integrity
+of push subscription data. This rule can be ignored for testing purposes
+only. This requires a valid, trusted TLS certificate (for example, from `Let's
+Encrypt`_).
+
+The ``ssl_key`` and ``ssl_cert`` options configure a TLS certificate for both
+the endpoint's HTTPS and connection node's WebSocket Secure (WSS) connections:
+
```
jrconlin: Wonder if we should note that different creds can be used for endpoint and connection handlers by moving these option definitions to the appropriate config files? (Could be consideration if you don't have a wild card cert.)

docs/running.rst
```diff
@@ -65,6 +65,39 @@ You may also want to enable human readable logs in the shared config file:
...
+    ; TLS private key and certificate file paths, used for the main
+    ; connection and endpoint listeners. If omitted, TLS will not be
+    ; used.
+    ssl_key = /etc/letsencrypt/live/EXAMPLE.COM/privkey.pem
+    ssl_cert = /etc/letsencrypt/live/EXAMPLE.COM/cert.pem
+    ; Path to file containing any additional parameters for the Ephemeral
+    ; Diffie-Hellman
+    #ssl_dh_param =
+
+The connection between the endpoint and connection nodes should not be exposed
```
jrconlin: "exposed" or "accessible"? Really, it should be locked to only other push server machines, but that might be more complicated than necessary.

docs/running.rst
pjenvey: Sounds good

jrconlin:
r+

Issue #501, opened 2016-06-30
Limit HTTP request body sizes at the protocol level


kitcambridge:
We currently limit body sizes to 4k, but I think we check that after the entire request body has already been read.

IIUC, it's still possible for a malicious client to send gigabytes of data, which we'll buffer into memory (or a temporary file, from the looks of https://github.com/fiorix/cyclone/blob/master/cyclone/httpserver.py#L168-L179) before our request handler gets called.

jrconlin:
This might be an odd one.

The code above relies on the data already having been read and stored in the `data` var. `data` is filled by [`lineReceived`](https://github.com/fiorix/cyclone/blob/master/cyclone/httpserver.py#L90), which is subclassed from Twisted's [`basic.LineReceiver`](https://twistedmatrix.com/documents/current/api/twisted.protocols.basic.LineReceiver.html) which may top out at [16,384](https://twistedmatrix.com/documents/current/api/twisted.protocols.basic.LineReceiver.html#MAX_LENGTH) (This might be an interesting issue for things like VAPID, but that's beside the point right now.)  The parent process reads data from the source in 65,535 byte chunks, minus the max header size,  leaves a fair bit of slop. If the `content-length` header [is specified](https://github.com/fiorix/cyclone/blob/master/cyclone/httpserver.py#L100) cyclone limits the data to that length. 

The HTTPConnection class that handles the rawDataRecieved() function is set as [a class variable](https://github.com/fiorix/cyclone/blob/master/cyclone/web.py#L1364) for Application. We could monkey patch it to reject any connection that exceeds 65K, but I'm not really sure that's a good idea. 
I'm not really sure there's a point where we can inject ourselves into that flow to, say, restrict twisted's TCP handler to only accept one 65K block. 

Continuing to dig into this.

pjenvey:
FYI protocol class vars is a typical pattern for twisted factories. It's made for overwriting -- the preferable way is to do it is via subclassing the factory (vs monkeypatching).

jrconlin:
Everytime I see that, I keep thinking it's more an anti-pattern. Reaching into objects directly just feels unclean. Still, if it's the expected behavior...

PR #602, opened 2016-08-16
feat: Limit the size of allowed HTTP bodies & headers

jrconlin:
Limits the amount of data preventing a potential DDoS. This extends much
of what's already in place. Twisted limits TCP read chunks to 65K. It
also already limits the max length of header lines to 16,384 bytes. This
forces connections closed if too much data is attempted to be sent.

NOTE: currently too much data events are not being logged, mostly due
to the fact that errors would lack a good deal useful info.

@pjenvey r?

### DIFF
--- a/autopush/web/limitedhttpconnection.py
+++ b/autopush/web/limitedhttpconnection.py
@@ -1,0 +1,54 @@
+from cyclone import httpserver
+from twisted.logger import Logger
+
+
+class LimitedHTTPConnection(httpserver.HTTPConnection):
+    """
+    Limit the amount of data being sent to a reasonable amount.
+
+    twisted already limits TCP streamed chunk reads to 65K, with
+    ~16k per header line. By default, we'll limit the number of
+    header lines to 100, and the maximum amount of data for the body
+    to be 4K.
+
+    """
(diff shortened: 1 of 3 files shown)


autopush/web/limitedhttpconnection.py
```diff
@@ -0,0 +1,54 @@
...
+            Logger().warn("Too much data sent, terminating connection")
+            return self.lineLengthExceeded(data)
+        if self.content_length is not None:
+            data, rest = data[:self.content_length], data[self.content_length:]
+            self.content_length -= len(data)
+        else:
+            rest = ''
+
+        self._contentbuffer.write(data)
+        if self.content_length <= 0:
```
pjenvey: nice bugfix (I noticed this is == 0 in cyclone) =]

pjenvey:
r+

I'll note (per your comment on #501) you can also override a factory's buildProtocol to pass the protocol values at runtime (like maxData). That requires customizing your own factory, usually subclassing, though, so it's extra work (I'm totally fine with this patch though, we already do this unclean assignment w/ PushServerProtocol in main.py)

PR #603, opened 2016-08-16
bug: Check tokens in constant time

### LINKED ISSUE #571: Make token check in endpoint.py constant time
in endpoint.*._validate_auth we should make token comparison constant time (see settings.py)

jrconlin:
Closes #571 

@pjenvey r+


pjenvey:
r+

PR #608, opened 2016-08-18
fix: return 404 for invalid URL's for consistency

### LINKED ISSUE #578: Report invalid subscription_ids more specifically
In testing error responses from autopush, we modified valid subscription endpoint URLs (e.g., `https://updates-autopush.stage.mozaws.net/wpush/v1/gAAAAABXm9zL46k0BF-HXEwueq...`) by altering one character of the last component (beginning `gAAAAA...`). The intention was to prompt a 410 GONE or 404 NOT FOUND response, specific to that token. Instead we got a 400 BAD REQUEST, with no content or other explanation.

Since there are a number of cases that can cause an "empty" (no content) 400 response (mostly meaning that the headers for the POST are not correct), our application server code isn't able to interpret the 400 as meaning the subscription_id is incorrect, so it remains in our system, eligible to have future messages sent.

It would be preferable if either the response status code were 410 or 404, or if a specific message could be included in the content to explain **why** the 400 response was generated.

bbangert:
URL's that are invalid should return a 404, just like any other
invalid URL.

Closes #578

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -5,7 +5,6 @@ from txstatsd.metrics.metrics import Metrics
 
 import datadog
 from datadog import ThreadStats
-from datadog.util.hostname import get_hostname
 
 
 class IMetrics(object):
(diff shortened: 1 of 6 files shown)


jrconlin:
update http://autopush.readthedocs.io/en/latest/http.html#error-codes ?

jrconlin:
r+

pjenvey:
r+

autopush/metrics.py
```diff
@@ -71,12 +71,13 @@ def timing(self, name, duration, **kwargs):
 class DatadogMetrics(object):
     """DataDog Metric backend"""
     def __init__(self, api_key, app_key, flush_interval=10,
-                 namespace="autopush"):
+                 namespace="autopush", hostname="localhost"):
```
jrconlin: hrm. tempted to say this should be None instead of "localhost". Settings sets it to a value, and that's pretty much the only init location.

That or we drop `get_hostname()` and treat the fact that it's been forced to None as an error.

jrconlin:
r++

PR #611, opened 2016-08-19
bug: trap JWS/JWT errors from being reported as Sentry Errors

### LINKED ISSUE #610: Trap JWT/JWS error exceptions
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/345791/

Some non-critical JWT errors are leaking into Sentry. We should trap those and just reject the submission with appropriate error codes and info.

jrconlin:
Silence Jose errors from handling the Auth header.

Closes #610 

@bbangert , @pjenvey r?

### DIFF
--- a/autopush/web/validation.py
+++ b/autopush/web/validation.py
@@ -79,6 +79,7 @@ class ThreadedValidate(object):
             d.addCallback(self._call_func, func, request_handler, *args,
                           **kwargs)
             d.addErrback(request_handler._overload_err)
+            d.addErrback(request_handler._boto_err)
             d.addErrback(request_handler._validation_err)
             d.addErrback(request_handler._response_err)
         return wrapper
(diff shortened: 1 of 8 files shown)


autopush/web/validation.py
```diff
@@ -320,8 +321,8 @@ def validate_auth(self, d):
 
         try:
             jwt = extract_jwt(token, public_key)
-        except ValueError:
-            raise InvalidRequest("Invalid Authorization Header",
+        except (ValueError, JOSEError) as ex:
+            raise InvalidRequest("Invalid Authorization Header: %s" % str(ex),
```
bbangert: Do we really want to tell the user the exact jose error message?

jrconlin: I was kinda debating that. In interests of clarity and developer relationship, it wouldn't hurt. From a more paranoid PoV, it's never a good idea to give more data than absolutely necessary. I'm trying to be less paranoid, but am absolutely fine with not disclosing the actual error.

bbangert:
r+

pjenvey:
r+

Issue #612, opened 2016-08-19
General Code Cleanup


jrconlin:
There are a number of out-of-scope or unused variables in the code, missing docstrings, and other bits of legacy that it'd be nice to clean up. IDEs like PyCharm can help identify many of these.

PR #614, opened 2016-08-19
feat: Code cleanup

jrconlin:
* replaces python reserved word var names
* replaces mixed case var names
* fixes setting chain breaks (message_read/write_throughput)
* removes unneeded var instantiations
* normalize HealthHandler to base from AutoendpointHandler, so
ap_settings is known
* instantiate from dict() rather than {}
* limit except handlers to known exceptions

issue #612

### DIFF
--- a/autopush/crypto_key.py
+++ b/autopush/crypto_key.py
@@ -31,14 +31,14 @@ http://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-00#section-4
         chunks = header.split(",")
         for chunk in chunks:
             bits = chunk.split(";")
-            hash = {}
+            ck_hash = {}
             for bit in bits:
                 try:
                     (key, value) = bit.split("=", 1)
                 except ValueError:
                     raise CryptoKeyException("Invalid Crypto Key value")
-                hash[key.strip()] = value.strip(' "')
-            self._values.append(hash)
+                ck_hash[key.strip()] = value.strip(' "')
+            self._values.append(ck_hash)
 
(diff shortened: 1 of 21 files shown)


pjenvey:
HealthHandler doesn't need to subclass AutoendpointHandler, the initialize method's from cyclone RequestHandler, so HealthHandler(.., ap_settings=ap_settings) in the test should just work as is (that makes a bit more sense too because it's not just for the endpoint).

Also the cyclone add_handlers calls in main should accept the same style as Application() -- so we cleanup the .ap_settings class vars dance in main.py by doing add_handlers([(.., HealthHandler, dict(ap_settings=ap_settings)) like we do here:

https://github.com/mozilla-services/autopush/blob/54c4526/autopush/main.py#L540

jrconlin:
Yeah, this is going to be bikeshedding on an unbelievable scale.

Most of these changes were driven by wanting to make PyCharm shut up. In the case of HealthHandler, it's due to the fact that PyCharm can't work out the source for `ap_settings`, unless you specify the direct class chain. In some respects, I understand since cyclone does a fair bit of magic to get that to happen that's not very obvious.

Happy to revert it if that's a problem. If so, I'll probably add a comment to the docstring explaining what's going on.

pjenvey:
ah ok, no worries. let's get it in before more conflicts pop up

r+

bbangert:
This appears to include other commits that are already on master, are we going to get more dupe commits in the history if this is merged as-is?

bbangert:
r+

Issue #613, opened 2016-08-19
Improve sentry reports w/out context


pjenvey:
Some sentry reports lack context due to twisted Failures not including a traceback.

One cause is twisted.web.client's _WrapperException which wraps other exceptions (possibly even multiple exceptions) and commonly lack a tb. We should special case these: log their wrapped exceptions instead, which'll likely include a context.

e.g. ResponseNeverReceived is one:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/345377/

There's others without a tb for unknown reasons:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/345112/events/8063796/

We can special case the logger to forcefully include the current stack trace when coming across these. That *might* provide enough context to track down their cause so we can improve their handling.

pjenvey:
Er, before special casing _WrapperExceptions inside the logger itself, let's ensure they're all trapped/handled for all our twisted.web.client calls. So they never propagate down to the logger in the first place.

(The logger special casing might still be useful in case we forget to handle new ones in the future).

PR #617, opened 2016-08-22
feat: send sentry the current stack when lacking a Failure tb

pjenvey:
it's sent as sentry's top level stacktrace vs its explicit exception stacktrace

issue #613

r? @bbangert @jrconlin

### DIFF
--- a/autopush/logging.py
+++ b/autopush/logging.py
@@ -12,6 +12,7 @@ import threading
 import boto3
 import raven
 from raven.transport.twisted import TwistedHTTPTransport
+from raven.utils.stacks import iter_stack_frames
 from twisted.internet import reactor
 from twisted.logger import (
     formatEvent,
@@ -90,11 +91,7 @@ class PushLogger(object):
 
     def __call__(self, event):
         if self.raven_client and 'log_failure' in event:
-            f = event["log_failure"]
-            reactor.callFromThread(
-                self.raven_client.captureException,
-                exc_info=(f.type, f.value, f.getTracebackObject())
-            )
+            self.raven_log(event)
 
         if event["log_level"] < self.log_level:
             return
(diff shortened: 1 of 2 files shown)


autopush/logging.py
```diff
@@ -108,6 +105,23 @@ def __call__(self, event):
             self._output.write(unicode(text))
             self._output.flush()
 
+    def raven_log(self, event):
+        f = event["log_failure"]
+        stack = extra = None
+        tb = f.getTracebackObject()
+        if not tb:
+            # include the current stack for at least some context
+            stack = list(iter_stack_frames())[4:]  # approx.
```
jrconlin: Will there ever be a chance that the iter_stack_frames could ever be less than 4? (I know, with the frameworks, fairly remote if possible, but still, wouldn't want to lose a report to an index error.)

bbangert: Ah, so this will catch the function that actually log.failure(SomeException())?

bbangert:
r+

autopush/logging.py
pjenvey: iter_stack_frames isn't "officially" public but I see it being used enough elsewhere that I'd consider it public enough (so likely stable enough). it's very simple, get the current stack in its particular format. raven strips frames from its result in a couple different ways so I expect it to not change.

I'll note 4: is more about the number of twisted logging frames (+ PushLogger's) involved, which could change too. 5 cleans it all up but I approximated to 4 just in case yet still getting rid of most of it

jrconlin:
r+

autopush/logging.py
pjenvey: Yea @bbangert. Of course your function could still be the reactor or Deferred's loop. We have a few of those in websocket.py -- addErrback(self.log_failure..). But at least now you would know PushServerProtocol.log_failure called it which is better than absolutely nothing

PR #619, opened 2016-08-23
fix: trap UnicodeEncodeError

### LINKED ISSUE #606: Capture processHandshake error
processHandshake does not handle headers happily when a unicode error occurs. We should capture this better since our override calls the main one.

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe5' in position 9: ordinal not in range(128)
  File "autopush/websocket.py", line 111, in wrapper
    return func(self, *args, **kwargs)
  File "autopush/websocket.py", line 383, in processHandshake
    return self.parent_class.processHandshake(self)
  File "autobahn/websocket/protocol.py", line 2552, in processHandshake
    protocols = [str(x.strip()) for x in self.http_headers["sec-websocket-protocol"].split(",")]
```

bbangert:
Closes #606 

r? @pjenvey @jrconlin

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -245,6 +245,20 @@ class WebsocketTestCase(unittest.TestCase):
         self.proto.processHandshake()
         eq_(self.proto.factory.externalPort, 80)
 
+    def test_handshake_decode_error(self):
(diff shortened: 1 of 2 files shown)


pjenvey:
r+

jrconlin:
r+

PR #620, opened 2016-08-23
fix: disable log_exception capturing in tests

### LINKED ISSUE #618: Disable log_exception for tests
In websocket.py, @log_exception eats exceptions rather aggressively. This leads to the side_effects in several processHandshake tests not reporting their output in the tests. That can lead us to not notice eq_ failures in the side_effects.

We should disable the decorator so that it's only run for its own test, and won't interfere in results.

bbangert:
closes #618 

r? @jrconlin @pjenvey


jrconlin:
r+

pjenvey:
r+

Issue #554, opened 2016-07-25
autoendpoint throwing a ResponseNeverReceived exception


pjenvey:
Per sentry:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/341754/

Unfortunately there's no context attached to it, so the first step is finding out where it's happening.

pjenvey:
Closing out for now per 3cc24fe which hopefully solves it

PR #621, opened 2016-08-23
fix: handle more errors to connection nodes

### LINKED ISSUE #613: Improve sentry reports w/out context

pjenvey:
* prefer base classes ConnectError/ConnectionClosed/ResponseFailed,
  covering ConnectionRefused/UserError, ConnectionLost/Done,
  ResponseNeverReceived
* handle ResponseFailed (ResponseNeverReceived) in the routers as
  a broken connection
* trap these in PushServerProtocol._notify_node

closes #613, #554

r? @bbangert @jrconlin

### DIFF
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -23,10 +23,11 @@ from twisted.internet.defer import (
 )
 from twisted.internet.error import (
     ConnectError,
+    ConnectionClosed,
     ConnectionRefusedError,
-    UserError
 )
 from twisted.logger import Logger
+from twisted.web._newclient import ResponseFailed
 from twisted.web.client import FileBodyProducer
 
 from autopush.protocol import IgnoreBody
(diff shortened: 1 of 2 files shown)


jrconlin:
much cleaner. r+

bbangert:
r+

PR #624, opened 2016-08-23
fix: kill dead_cache, it's no longer used

### LINKED ISSUE #623: kill simple router's dead_cache
It's now dead code and according to @bbangert, probably didn't do much for us anyway

pjenvey:
closes #623 

r? @jrconlin @bbangert


jrconlin:
r+ 
(Although I wonder if I should give an r- since this is mostly removal of code, and I wouldn't want to break the trend.)

bbangert:
r+

PR #626, opened 2016-08-24
break out base deps into their own req file

### LINKED ISSUE #625: create a base requirements file
We should kill the duplication of requirements file dependencies by putting them all in a base requirements file. Other req files can source it w/ a **-r base** line.

I'm also tempted to move them all into a requirements/ dir since we have a number of them, plus a new base

pjenvey:
closes #625 

r? @jrconlin @bbangert


pypy-requirements.txt
```diff
@@ -1,60 +1 @@
-ConfigArgParse==0.10.0
```
jrconlin: Do we really need this file if there's no difference?

jrconlin:
r+

pypy-requirements.txt
pjenvey: Currently no but since one more file doesn't matter, it's always possible we may need it later. We did have the differing readline dep recently which would have needed it

bbangert:
+r

PR #628, opened 2016-08-24
bug: clarify token references in HTTP endpoint docs

### LINKED ISSUE #627: Clarify bridge documentation
There is clear confusion regarding how to use the bridge HTTP interface.

Address these concerns.
```
2:20 PM <kitcambridge> i've come to the conclusion i don't understand anything on http://autopush.readthedocs.io/en/latest/http.html#push-service-bridge-http-interface
2:22 PM <kitcambridge> specifically, i'm confused what "The bridge specific public exchange token" and "The bridge specific private identifier token" are, what token is in `POST /v1/{type}/{token}/registration`, and how that's different from {"token": "1ab2c3"}
2:24 PM <kitcambridge> and what's "Authorization: WebPush 0123abcdef", isn't that vapid? do we use vapid for the http bridge, too?
2:28 PM <•jrconlin> kitcambridge: yeah, the problem is that "token" is very generic.
2:30 PM <•jrconlin> In the URL "token" (as defined by third party systems that use it and call it "token") is the sender identifier. For GCM/FCM it's SenderID, for APNS it's irrelevant.
2:30 PM <•jrconlin> For the body "token" is (as defined by the same third party systems that use and call it "token" because they hate me) is the client ID token. For both of these, it's a long string identifier. 
2:31 PM — kitcambridge is so confused
2:32 PM <•jrconlin> I feel your pain.
2:32 PM <•jrconlin> O
2:32 PM <kitcambridge> so the senderID is a token, the client ID is a token, but the client ID is not the uaid, and there's an auth_token that's a bearer token
2:32 PM <•jrconlin> I will happily go and revamp the docs to be less confusing, provided I can figure out how to best specify exactly that.
2:33 PM <•jrconlin> pretty much.
2:33 PM <•jrconlin> And I can't use "senderID" because that's limited to a single bridge system.
2:33 PM <kitcambridge> ok
2:35 PM <kitcambridge> what's the apns version of senderID?
2:35 PM <•jrconlin> If it's any help, I have to verify that the "auth_token" mentioned is the secret handed back in the registration.
2:35 PM <•jrconlin> for APNS, the senderID is irrelevant. use any value, like "aaaa" if you want.
2:36 PM <kitcambridge> oh, so the auth_token isn't a vapid thing?
2:36 PM <•jrconlin> no, since it predates vapid.
2:36 PM <kitcambridge> what does it do?
2:37 PM <•jrconlin> bearer token for auth. 
2:37 PM <kitcambridge> looks like it's just the secret that gets passed back on registration?
2:37 PM <•jrconlin> That is what I'm looking to confirm.
2:37 PM <benbangert> kitcambridge: its to reduce the attack vector of griefers guessing UUID's
2:38 PM <benbangert> we won't honor requests to add/drop sub's if the auth token doesn't match the UAID
2:39 PM <kitcambridge> ah cool
2:40 PM <•jrconlin> confirmed, it's a simple bearer token provided by the registration call. Will update the docs to be clearer about that.
2:40 PM <•jrconlin> I may also change the type name from WebPush to "Bearer" since that's resolved now.
2:40 PM <•jrconlin> (Code actually ignores that bit)
2:41 PM <kitcambridge> ok, so...
2:41 PM <kitcambridge> let's say i want to register for gcm
2:41 PM <•jrconlin> ok.
2:42 PM <kitcambridge> i do POST /v1/gcm/mySenderID/registration with no Authorization header and { token: "myClientIDToken" } in the body
2:42 PM <•jrconlin> kitcambridge: correct.
2:43 PM <kitcambridge> that stores mySenderID and myClientID in dynamo, and creates an endpoint for me, too
2:43 PM <kitcambridge> and i get back a uaid, secret, channelID, endpoint
2:43 PM <•jrconlin> correct
2:43 PM <kitcambridge> awesome
2:43 PM <kitcambridge> now, i want to register another channel
2:44 PM <•jrconlin> ok.
2:44 PM <kitcambridge> so, i do POST /v1/gcm/mySenderID/registration/uaid/subscription
2:44 PM <•jrconlin> You PUT, but yes, to that URL.
2:44 PM <kitcambridge> oh, ok
2:45 PM <•jrconlin> Sorry, my fault, 
2:45 PM <kitcambridge> so i PUT to that
2:45 PM <kitcambridge> nothing in the body
2:45 PM <•jrconlin> You POST. Doc wasn't updated.
2:45 PM <•jrconlin> ARGH! 
2:45 PM <kitcambridge> oh, no worries. ok, so I POST, with nothing in the body, and a `Authorization: Bearer {secret}` header
2:45 PM <•jrconlin> it's put. (really need to fix that.)
2:47 PM <kitcambridge> ok, and to update, if the senderID or the myClientIDToken changes...
2:47 PM <kitcambridge> i do PUT /v1/gcm/senderID/registration/{uaid}
2:47 PM <kitcambridge> also with `Authorization: Bearer {secret}`
2:47 PM <•jrconlin> argh2: it's POST, Put is used to change the third party id used to identify the recipient.
2:47 PM <kitcambridge> and with { token: "myNewClientIDToken" } in the body
2:47 PM <•jrconlin> Correct.
2:49 PM <kitcambridge> ok, cool, thanks :-)
2:49 PM <kitcambridge> now, if i'm apns, then there's no senderID
2:50 PM <kitcambridge> so the url prefix is going to be something like /v1/apns/aaaa/registration
2:50 PM <•jrconlin> yep.
2:51 PM <•jrconlin> We need that "aaaa" bit just because it's expected by the server, but the code ignores it completely.
2:51 PM <•jrconlin> You can use any, rendering value, like "a" or "0" if you like.
2:52 PM <•jrconlin> And the code cleanup refactor just blew chunks all over my git repo. Yay!
2:53 PM <kitcambridge> gotcha
2:55 PM — kitcambridge saves this scrollback for future reference
2:56 PM <kitcambridge> walking through that was really helpful, jrconlin. thanks!
2:56 PM <•jrconlin> heh, yeah, well, now i get to translate that into the docs, too. 
2:56 PM <•jrconlin> And yeah, clearing up the syntax helps a lot.
2:57 PM <kitcambridge> it was useful to see the flow
```

jrconlin:
There was confusion about the various token references. These are
clarified with simpler references and more distinct labels.

closes #627

### DIFF
--- a/docs/http.rst
+++ b/docs/http.rst
@@ -34,10 +34,6 @@ Push Service HTTP API
 The following section describes how remote servers can send Push
 Notifications to apps running on remote User Agents.
 
-API methods requiring Authorization must provide the Authorization
-header containing the authrorization token. The Authorization token is returned
-as "secret" in the registration response.
-
 Lexicon
 -------
 
@@ -85,7 +81,7 @@ Error Codes
 Unless otherwise specified, all calls return the following error codes:
 
 -  20x - Success
--  301 - Moved + `Location:` if `{token}` is invalid (Bridge API Only)
+-  301 - Moved + `Location:` if `{client_token}` is invalid (Bridge API Only)
 -  400 - Bad Parameters
 
    - errno 101 - Missing neccessary crypto keys
@@ -133,11 +129,14 @@ Calls
 Send Notification
 ~~~~~~~~~~~~~~~~~
 
-Send a notification to the given endpoint identified by it's `token`.
+Send a notification to the given endpoint identified by its `push_endpoint`.
+Please note, the Push endpoint URL (which is what is used to send notifications)
+should be considered "opaque". We reserve the right to change any portion
+of the Push URL in future provisioned URLs.
 
 **Call:**
 
-.. http:put:: /push/{token}
+.. http:put:: {push_endpoint}
 
     If the client is using webpush style data delivery, then the body in its
     entirety will be regarded as the data payload for the message per
@@ -165,7 +164,7 @@ Send a notification to the given endpoint identified by it's `token`.
 
 **Return Codes:**
 
-    :statuscode 404: `token` is invalid.
+    :statuscode 404: Push subscription is invalid.
     :statuscode 202: Message stored for delivery to client at a later
                      time.
     :statuscode 200: Message delivered to node client is connected to.
@@ -248,6 +247,10 @@ external protocol like
 `GCM <https://developers.google.com/cloud-messaging/>`__/`FCM <https://firebase.google.com/docs/cloud-messaging/>`__ or
 `APNs <https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Introduction.html#//apple_ref/doc/uid/TP40008196-CH1-SW1>`__. All message bodies must be UTF-8 encoded.
 
+API methods requiring Authorization must provide the Authorization
+header containing the registration secret. The registration secret is
+returned as "secret" in the registration response.
+
 Lexicon
 -------
 
@@ -258,30 +261,38 @@ For the following call definitions:
 Allowed bridges are `gcm` (Google Cloud Messaging), `fcm` (Firebase Cloud
 Messaging), and `apns` (Apple Push Notification system)
 
-   :{token}: The bridge specific public exchange token
+   :{app_id}: The bridge specific application identifier
 
-Each protocol requires a unique token that addresses the remote application.
-For GCM/FCM, this is the `SenderID` and is pre-negotiated outside of the push
-service.
+Each bridge may require a unique token that addresses the remote application
+For GCM/FCM, this is the `SenderID` (or 'project number') and is pre-negotiated outside of the push
+service. You can find this number using the
+`Google developer console <https://console.developers.google.com/iam-admin/settings/project>`__.
+For APNS, there is no client token, so this value is arbitrary.
+Feel free to use "a" or "0" or any other valid URL path token. For our examples, we will use a client token of
+"33clienttoken33".
 
-   :{instanceid}: The bridge specific private identifier token
+   :{instance_id}: The bridge specific private identifier token
 
-Each protocol requires a unique token that addresses the
-application on a given user's device. This is usually the product of the
-application registering the {instanceid} with the native bridge protocol
-agent.
+Each bridge requires a unique token that addresses the
+application on a given user's device. This is the
+"`Registration Token <https://firebase.google.com/docs/cloud-messaging/android/client#sample-register>`__" for
+GCM/FCM or "`Device Token <https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/IPhoneOSClientImp.html#//apple_ref/doc/uid/TP40008194-CH103-SW2>`__"
+for APNS. This is usually the product of the
+application registering the {instance_id} with the native bridge via the user
+agent. For our examples, we will use an instance ID of "11-instance-id-11".
 
-   :{auth_token}: The Authorization token
+   :{secret}: The registration secret from the Registration call.
 
 Most calls to the HTTP interface require a Authorization header. The
-Authorization header is a bearer token, which has been provided by the
-**Registration** call and is preceded by the token type word "WebPush".
+Authorization header is a simple bearer token, which has been provided by the
+**Registration** call and is preceded by the scheme name "Bearer". For
+our examples, we will use a registration secret of "00secret00".
 
 An example of the Authorization header would be:
 
 ::
 
-    Authorization: WebPush 0123abcdef
+    Authorization: Bearer 00secret00
 
 Calls
 -----
@@ -290,19 +301,19 @@ Registration
 ~~~~~~~~~~~~
 
 Request a new UAID registration, Channel ID, and optionally set a bridge
-type and token for this connection.
+type and 3rd party bridge instance ID token for this connection.
 
 **Call:**
 
 
-.. http:post:: /v1/{type}/{token}/registration
+.. http:post:: /v1/{type}/{app_id}/registration
 
-This call requires no Authorization for first time use.
+This call requires no Authorization header for first time use.
 
 **Parameters:**
 
 
-    {"token":{instanceid}}
+    {"token":{instance_id}}
 
     .. note::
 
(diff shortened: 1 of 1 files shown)


jrconlin:
formatted version available at: http://10.252.26.247/docs/http.html#push-service-bridge-http-interface

docs/http.rst
```diff
@@ -35,8 +35,8 @@ The following section describes how remote servers can send Push
 Notifications to apps running on remote User Agents.
 
 API methods requiring Authorization must provide the Authorization
-header containing the authrorization token. The Authorization token is returned
-as "secret" in the registration response.
+header containing the authrorization secret. The Authorization secret is
```
pjenvey: typo here before: "authrorization"

docs/http.rst
```diff
@@ -133,11 +133,14 @@ Calls
 Send Notification
 ~~~~~~~~~~~~~~~~~
 
-Send a notification to the given endpoint identified by it's `token`.
+Send a notification to the given endpoint identified by it's `push_endpoint`.
```
pjenvey: bah, "its" while you're here, now I'm a grammar nazi

jrconlin: yeah, there's a bit of a difference between pointing out mistakes and, well, Nazism. So, yeah, feel free to point out my mistakes.

docs/http.rst
jrconlin: odd, wonder why the spill chalkier didn't catch that.

pjenvey:
r+

docs/http.rst
```diff
@@ -35,8 +35,8 @@ The following section describes how remote servers can send Push
 Notifications to apps running on remote User Agents.
 
 API methods requiring Authorization must provide the Authorization
```
kitcambridge: Maybe this would be better under the "Push Service Bridge HTTP Interface" section? Placing it here at the top is a little confusing because the "Push Service HTTP API" uses VAPID, not the secret.

docs/http.rst
```diff
@@ -35,8 +35,8 @@ The following section describes how remote servers can send Push
...
-header containing the authrorization token. The Authorization token is returned
-as "secret" in the registration response.
+header containing the authorization secret. The Authorization secret is
```
kitcambridge: What do you think of calling it the "registration secret", to make it clearer that it's only for the bridge API? "Authorization secret" feels a little generic.

docs/http.rst
```diff
@@ -165,7 +168,7 @@ Send a notification to the given endpoint identified by it's `token`.
 
 **Return Codes:**
 
-    :statuscode 404: `token` is invalid.
+    :statuscode 404: Push 'token' is invalid.
```
kitcambridge: How about "push subscription is invalid"?

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
 Allowed bridges are `gcm` (Google Cloud Messaging), `fcm` (Firebase Cloud
 Messaging), and `apns` (Apple Push Notification system)
 
-   :{token}: The bridge specific public exchange token
+   :{client_token}: The bridge specific public exchange token
```
kitcambridge: How about "the bridge specific application ID"?

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
 
-Each protocol requires a unique token that addresses the remote application.
+Each protocol may require a unique token that addresses the remote application
 For GCM/FCM, this is the `SenderID` and is pre-negotiated outside of the push
```
kitcambridge: How about: "this is the 'sender ID' or 'project number', created when you set up a project in the Google Dev Console."

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
-service.
+service. For APNS, there is no client token, so this value is arbitrary.
+Feel free to use "a" or "0" or any other valid URL path token. For our examples, we will use a client token of "33clienttoken33".
 
-   :{instanceid}: The bridge specific private identifier token
+   :{instance_id}: The bridge specific private identifier token
```
kitcambridge: How about describing this as the "API key (GCM) or device token (APNS)"?

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
 
 Each protocol requires a unique token that addresses the
```
kitcambridge: Nit: s/protocol/bridge

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
 application on a given user's device. This is usually the product of the
```
kitcambridge: "This is the FCM registration token or APNS device token, generated when the UA registers with the bridge."

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
 Allowed bridges are `gcm` (Google Cloud Messaging), `fcm` (Firebase Cloud
 Messaging), and `apns` (Apple Push Notification system)
 
-   :{token}: The bridge specific public exchange token
+   :{client_token}: The bridge specific public exchange token
 
-Each protocol requires a unique token that addresses the remote application.
+Each protocol may require a unique token that addresses the remote application
```
kitcambridge: I wonder if we should call this the UA instead of "remote application," here and everywhere. When we're talking about endpoints and the WebSocket protocol, we use "app" and "app server" to refer to the web app, but here, "application" refers to Fennec (the UA).

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
+agent. For our examples, we will use an instance ID of "11-instance-id-11".
 
-   :{auth_token}: The Authorization token
+   :{secret}: The shared secret token from the Registration call.
 
 Most calls to the HTTP interface require a Authorization header. The
-Authorization header is a bearer token, which has been provided by the
-**Registration** call and is preceded by the token type word "WebPush".
+Authorization header is a simple bearer token, which has been provided by the
+**Registration** call and is preceded by the token type word "Bearer". For
```
kitcambridge: s/token type word/scheme name/.

docs/http.rst
```diff
@@ -258,30 +261,32 @@ For the following call definitions:
...
 
 Each protocol requires a unique token that addresses the
 application on a given user's device. This is usually the product of the
-application registering the {instanceid} with the native bridge protocol
-agent.
+application registering the {instance_id} with the native bridge protocol
+agent. For our examples, we will use an instance ID of "11-instance-id-11".
 
-   :{auth_token}: The Authorization token
+   :{secret}: The shared secret token from the Registration call.
```
kitcambridge: Let's just say "shared secret" or "registration secret", token is implied here. (Can you tell I have a soft spot in my heart for the word "token"? 😉)

docs/http.rst
```diff
@@ -418,8 +424,8 @@ example:
 
 .. code-block:: http
 
-    > POST /v1/fcm/a1b2c3/registration/abcdef012345/subscription
-    > Authorization: WebPush 0123abcdef
+    > POST /v1/fcm/33clienttoken33/registration/abcdef012345/subscription
+    > Authorization: Bearer 00secret00
```
kitcambridge: Awesome, thanks for cleaning up the params!

bbangert:
I think @kitcambridge has said everything I could possibly think to comment on.

docs/http.rst
jrconlin: Yeah, going to patch this up with direct links to the bridge protocols.

bbangert:
r+, looks great, quite readable now!

PR #637, opened 2016-08-26
bug: provide better logging and fix UAID misreference

### LINKED ISSUE #629: Don't send empty strings for attributes to DynamoDB
We're sending drop_user commands with an empty string to DynamoDB, we shouldn't.

Per https://sentry.prod.mozaws.net/operations/autopush-prod/issues/346331/

Traceback:
```
ValidationException: ValidationException: 400 Bad Request
{u'message': u'One or more parameter values were invalid: An AttributeValue may not contain an empty string', u'__type': u'com.amazon.coral.validate#ValidationException'}
  File "twisted/python/threadpool.py", line 246, in inContext
    result = inContext.theWork()
  File "twisted/python/threadpool.py", line 262, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "autopush/db.py", line 214, in wrapper
    return func(self, *args, **kwargs)
  File "autopush/db.py", line 632, in drop_user
    expected={"uaid__eq": huaid})
  File "boto/dynamodb2/table.py", line 913, in delete_item
    conditional_operator=conditional_operator)
  File "boto/dynamodb2/layer1.py", line 925, in delete_item
    body=json.dumps(params))
  File "boto/dynamodb2/layer1.py", line 2842, in make_request
    retry_handler=self._retry_handler)
  File "boto/connection.py", line 954, in _mexe
    status = retry_handler(response, i, next_sleep)
  File "boto/dynamodb2/layer1.py", line 2882, in _retry_handler
    response.status, response.reason, data)
```

jrconlin:
* when calling drop_user and the response.router_data is empty, fetch
the current UAID from known uaid_data
* make the logging far more verbose by adding _client_info to raven
calls

closes #629
@bbangert @pjenvey r?

### DIFF
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -24,6 +24,7 @@ class BaseHandler(cyclone.web.RequestHandler):
                                                self.request.remote_ip),
             authorization=self.request.headers.get('authorization', ""),
             message_ttl=self.request.headers.get('ttl', ""),
+            uri=self.request.uri,
         )
 
     def write_error(self, code, **kwargs):
@@ -39,8 +40,8 @@ class BaseHandler(cyclone.web.RequestHandler):
             self.log.failure(
                 format=kwargs.get('format', "Exception"),
                 failure=failure.Failure(*kwargs['exc_info']),
-                **self._client_info)
+                client_info=self._client_info)
         else:
             self.log.failure("Error in handler: %s" % code,
-                             **self._client_info)
+                             client_info=self._client_info)
         self.finish()
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -226,7 +226,7 @@ class AutoendpointHandler(BaseHandler):
         """errBack for throughput provisioned exceptions"""
         fail.trap(ProvisionedThroughputExceededException)
         self.log.info(format="Throughput Exceeded", status_code=503,
-                      errno=201, **self._client_info)
+                      errno=201, client_info=self._client_info)
         self._write_response(503, errno=201,
                              message="Please slow message send rate")
 
@@ -234,7 +234,8 @@ class AutoendpointHandler(BaseHandler):
         """errBack for JWS/JWT exceptions"""
         fail.trap(JOSEError)
         self.log.info(format="Authorization Failure",
-                      status_code=401, errno=109, **self._client_info)
+                      status_code=401, errno=109,
+                      client_info=self._client_info)
         self._write_response(401, errno=109,
                              message="Invalid Authorization")
 
@@ -242,7 +243,8 @@ class AutoendpointHandler(BaseHandler):
         """errBack for random boto exceptions"""
         fail.trap(BotoServerError)
         self.log.info(format="BOTO Error: %s" % str(fail.value),
-                      status_code=503, errno=202, **self._client_info)
+                      status_code=503, errno=202,
+                      client_info=self._client_info)
         self._write_response(503, errno=202,
                              message="Communication error, please retry")
 
@@ -267,20 +269,20 @@ class AutoendpointHandler(BaseHandler):
         if isinstance(exc, RouterException):
             if exc.log_exception or exc.status_code >= 500:
                 fmt = fail.value.message or 'Exception'
-                self.log.failure(format=fmt,
+                self.log.failure(format=fmt,  # pragma nocover
                                  failure=fail, status_code=exc.status_code,
                                  errno=exc.errno or "",
-                                 **self._client_info)  # pragma nocover
+                                 client_info=self._client_info)
             if 200 <= exc.status_code < 300:
                 self.log.info(format="Success", status_code=exc.status_code,
                               logged_status=exc.logged_status or "",
-                              **self._client_info)
+                              client_info=self._client_info)
             elif 400 <= exc.status_code < 500:
                 self.log.info(format="Client error",
                               status_code=exc.status_code,
                               logged_status=exc.logged_status or "",
                               errno=exc.errno or "",
-                              **self._client_info)
+                              client_info=self._client_info)
             self._router_response(exc)
 
     def _uaid_not_found_err(self, fail):
@@ -288,7 +290,7 @@ class AutoendpointHandler(BaseHandler):
         fail.trap(ItemNotFound)
         self.log.info(format="UAID not found in AWS.",
                       status_code=410, errno=103,
-                      **self._client_info)
+                      client_info=self._client_info)
         self._write_response(410, errno=103,
                              message="Endpoint has expired. "
                                      "Do not send messages to this endpoint.")
@@ -298,7 +300,7 @@ class AutoendpointHandler(BaseHandler):
         fail.trap(InvalidToken, InvalidTokenException)
         self.log.info(format="Invalid token",
                       status_code=400, errno=102,
-                      **self._client_info)
+                      client_info=self._client_info)
         self._write_response(400, 102,
                              message="Invalid endpoint.")
 
@@ -308,7 +310,7 @@ class AutoendpointHandler(BaseHandler):
         self.log.info(format="Invalid Auth token",
                       status_code=401,
                       errno=109,
-                      **self._client_info)
+                      client_info=self._client_info)
         self._write_unauthorized_response(
             message=fail.value.message,
             url="https://datatracker.ietf.org/doc/"
@@ -503,7 +505,7 @@ class EndpointHandler(AutoendpointHandler):
         except KeyError:
             self.log.debug(
                 format="Invalid router requested", status_code=400,
-                errno=108, **self._client_info)
+                errno=108, client_info=self._client_info)
             self._write_response(400, 108, message="Invalid router")
             return
 
@@ -520,14 +522,14 @@ class EndpointHandler(AutoendpointHandler):
             if data and not all([x in self.request.headers
                                  for x in req_fields]):
                 self.log.debug(format="Client error", status_code=400,
-                               errno=101, **self._client_info)
+                               errno=101, client_info=self._client_info)
                 self._write_response(
                     400, errno=101, message="Missing necessary crypto keys.")
                 return
             if ("encryption-key" in self.request.headers and
                     "crypto-key" in self.request.headers):
                 self.log.debug(format="Client error", status_code=400,
-                               errno=110, **self._client_info)
+                               errno=110, client_info=self._client_info)
                 self._write_response(
                     400, 110, message="Invalid crypto headers")
                 return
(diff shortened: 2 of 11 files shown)


autopush/db.py
```diff
@@ -407,6 +421,12 @@ def store_message(self, uaid, channel_id, message_id, ttl, data=None,
                       headers=None, timestamp=None):
         """Stores a message in the message table for the given uaid/channel
         with the message id"""
+        if not uaid:
```
pjenvey: r+

I'd love to chain all these `if not (uaid and channel_id and message_id)` but not a show stopper on late fri.

pjenvey:
r+

autopush/db.py
jrconlin: yeah, the only problem is that we need to get the logic right, and I wanted to be very explicit about what's being checked. I've tried to do clever logic chains in the past and failed horribly. I just default to simpler and pay the minor opcode penalty.

autopush/db.py
```diff
@@ -184,6 +184,8 @@ def preflight_check(storage, router, uaid="deadbeef00000000deadbeef00000000"):
             time.sleep(1)
 
     # Use a distinct UAID so it doesn't interfere with metrics
+    if not uaid:
+        raise ItemNotFound("Missing or empty UAID")
```
bbangert: Somewhat terrified we get this far without a uaid....

bbangert:
This reminds me of how our woes in endpoint.py started.... we'll just add a few if guards here, one there.... If we want to validate every functions arguments, we should use a validation library. If we can't call our own functions without flubbing it, I wonder where we stop. The slope feels slippery.

autopush/db.py
jrconlin: Same here, it's precautionary.

autopush/db.py
```diff
@@ -555,6 +557,8 @@ def get_uaid(self, uaid):
 
         """
         try:
+            if not uaid:
+                raise ItemNotFound("Missing or empty UAID")
```
bbangert: Thought we didn't need this anymore?

autopush/endpoint.py
```diff
@@ -209,13 +209,16 @@ def _write_unauthorized_response(self, message="Invalid authentication",
...
-    def _response_err(self, fail):
+    def _response_err(self, fail, caller=None):
         """errBack for all exceptions that should be logged
 
         This traps all exceptions to prevent any further callbacks from
         running.
 
         """
+        if not caller:
+            caller = self.__module__
```
bbangert: Why not just use "endpoint.CLASSNAME._response_err"?

autopush/endpoint.py
```diff
@@ -420,7 +423,7 @@ def _delete_message(self, kind, uaid, chid):
         d = deferToThread(self.ap_settings.message.delete_message, uaid,
                           chid, self.version)
         d.addCallback(self._delete_completed)
-        self._db_error_handling(d)
+        self._db_error_handling(d, caller=self.__module__ + ".delete")
```
bbangert: we already know this modules name, why not just use it?

autopush/endpoint.py
```diff
@@ -473,7 +476,9 @@ def put(self, api_ver="v0", token=None):
                           api_ver,
                           crypto_key_header)
         d.addCallback(self._process_auth)
-        d.addCallback(self._token_valid)
+        d.addCallback(self._token_valid,
+                      caller="%s.%s" % (self.__module__,
+                                        self.request.method.lower()))
```
bbangert: Why the function calls and format subs when we already know exactly where we are and the method?

jrconlin: We don't. This method is used for both "put" and "post". Pulling the method at least gives us that.

autopush/web/base.py
```diff
@@ -136,13 +106,16 @@ def _validation_err(self, fail):
...
-    def _response_err(self, fail):
+    def _response_err(self, fail, caller=None):
         """errBack for all exceptions that should be logged
 
         This traps all exceptions to prevent any further callbacks from
         running.
 
         """
+        if not caller:
+            caller = fail.frames[0][0]
```
bbangert: We're positive this will never throw an index error?

autopush/endpoint.py
jrconlin: This is the only source of logged 500,999 errors for this class, so pretty easy to figure out the method. The trick is figuring out the module.  While I don't expect caller to be empty by this point, it may be, so this is a last ditch attempt to harvest the module name.

autopush/endpoint.py
jrconlin: isn't that pretty much what's going on here? Is calling .__method__ expensive?

autopush/endpoint.py
bbangert: Actually, don't we want the class name since this is a base class and we want to know the actual class?

autopush/db.py
bbangert: Thought we were removing this too?

autopush/web/webpush.py
```diff
@@ -62,7 +64,7 @@ def _router_completed(self, response, uaid_data, warning=""):
                 # be deleted. There is no longer valid route information for
                 # this record.
                 d = deferToThread(self.ap_settings.router.drop_user,
-                                  self.uaid)
+                                  uaid_data.get("uaid"))
```
bbangert: This still leaves open the possibility that the router might remove the uaid, why not just grab uaid directly and risk the key error? If we see a KeyError, then we can fix the router error.

autopush/web/webpush.py
```diff
@@ -43,12 +44,13 @@ def post(self, api_ver="v1", token=None):
             headers=self.valid_input["headers"],
             ttl=self.valid_input["headers"]["ttl"]
         )
-
+        self._client_info["uaid"] = hasher(user_data.get("uaid"))
```
bbangert: Wouldn't setting self.uaid set this and appply hasher?

autopush/web/validation.py
```diff
@@ -74,14 +74,19 @@ def _decorator(self, func):
...
+            )
             d = deferToThread(self._validate_request, request_handler)
             d.addCallback(self._call_func, func, request_handler, *args,
                           **kwargs)
             d.addErrback(request_handler._overload_err)
             d.addErrback(request_handler._boto_err)
             d.addErrback(request_handler._validation_err)
-            d.addErrback(request_handler._response_err)
+            d.addErrback(request_handler._response_err,
+                         http_context=http_context)
```
bbangert: Why not just set the http context on the request_handler during its initialization? That way we always have it handy to pass in.

autopush/web/webpush.py
jrconlin: nope, removed the uaid setter for web/*. Removed the chid setter too. We're not using it and the only thing it was doing that was constructive was rendering this hash. So moved it to the initialize instead. We normalize the chid in db.py before we write it.

autopush/logging.py
```diff
@@ -118,6 +118,10 @@ def raven_log(self, event):
             exc_info=(f.type, f.value, tb),
             stack=stack,
             extra=extra,
+            log_format=event.get('log_format'),
```
pjenvey: AFAICT raven's pretty much going to ignore these kwargs. You wanna put them into the extra dict, that's its really where it wants arbitrary kwargs that'll show on the UI

autopush/web/simplepush.py
```diff
@@ -20,7 +21,10 @@ def put(self, api_ver="v1", token=None):
         sub = self.valid_input["subscription"]
         user_data = sub["user_data"]
         router = self.ap_settings.routers[user_data["router_type"]]
+        self._client_info["uaid"] = hasher(user_data.get("uaid"))
+        self._client_info["channel_id"] = user_data.get("chid")
         self._client_info["message_id"] = self.valid_input["version"]
+        self._client_info["stack"] = "%s._put" % self.__module__
```
pjenvey: I think you want to kill this line now? and the other stack one in webbpush

pjenvey:
r+

autopush/logging.py
```diff
@@ -113,12 +113,17 @@ def __call__(self, event):
...
+        extra = dict()
         tb = f.getTracebackObject()
         if not tb:
             # include the current stack for at least some context
             stack = list(iter_stack_frames())[4:]  # approx.
             extra = dict(no_failure_tb=True)
+        extra.update(log_format=event.get('log_format'),
+                     log_namespace=event.get('log_namespace'),
+                     why=event.get('why'),
+                     client_info=event.get('client_info'))
```
bbangert: Is this supposed to be getattr(event...) ? Below its attached as fail._client_info, does twisted change attributes on the fail into the event dict?

autopush/web/base.py
```diff
@@ -144,6 +113,7 @@ def _response_err(self, fail):
 
         """
         fmt = fail.value.message or 'Exception'
+        fail._client_info = self._client_info
```
bbangert: Ok, so the _client_info is attached to the failure object itself.... but above we search the event for 'client_info'. This works?

autopush/logging.py
```diff
@@ -113,12 +113,17 @@ def __call__(self, event):
 
     def raven_log(self, event):
         f = event["log_failure"]
-        stack = extra = None
+        stack = None
```
bbangert: This is stack, we include it in the sentry report, but its not the same 'stack' as the 'stack' we put into the client_info dict later. That seems a little confusing.

jrconlin: Forgot to remove the other "stack". Client info should have more than enough info to help us nail down bugs.

autopush/base.py
```diff
@@ -39,8 +39,8 @@ def write_error(self, code, **kwargs):
             self.log.failure(
                 format=kwargs.get('format', "Exception"),
                 failure=failure.Failure(*kwargs['exc_info']),
-                **self._client_info)
+                client_info=self._client_info)
```
bbangert: This will mess up our Kibana logging, where we do want the client_info dict as additional keys.

autopush/web/webpush.py
```diff
@@ -43,7 +44,9 @@ def post(self, api_ver="v1", token=None):
...
+        self._client_info["channel_id"] = user_data.get("chid")
+        self._client_info["stack"] = ["%s.post" % self.__module__]
```
pjenvey: one last thing -- let's *always* put this in _client_info by making BaseHandler._init_info do it, which is called at the beginning of every request.  Make it something like '%s.%s (%s)' % (self.__module__, self.__name__, self.request.method) i.e. 'autopush.endpoint.EndpointHandler (POST)'

the name 'stack' is confusing w/ the raven stack arg Ben pointed out above, maybe make it "handler" or something

autopush/web/base.py
bbangert: Is there a reason for attaching client_info here when its passed in below?

autopush/web/webpush.py
bbangert: Until the request is validated, we don't have all the client info, such as uaid.

autopush/base.py
jrconlin: Addressed in later patch. Kibana uses the json_formatting() function, so the client_info data is collapsed up in that function. raven will still get the sub-dict that will contain the full client-info data.

autopush/logging.py
jrconlin: Addressed in later patch. The "stack" element that was causing issues was removed. "log_namespace" specifies the module where the exception occurred, and added a "uri" element to client_info to specify the request path. These should provide enough info to determine the source of the error.

bbangert:
r+

pjenvey:
r+

PR #638, opened 2016-08-26
cleanup ap_settings

pjenvey:
Some cleanup while I was in here -- more than I originally intended. I extracted into the new BaseHandler to make sure PyCharm was as happy as it can be. There's more that can go into it, but enough for now

r? @jrconlin @bbangert

### DIFF
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -1,0 +1,46 @@
+import uuid
+
+import cyclone.web
+from twisted.logger import Logger
+from twisted.python import failure
+
+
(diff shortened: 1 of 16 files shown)


jrconlin:
looks good, r+

Wondering if we should merge this first, then rebase #637

pjenvey:
this managed to avoid db/test_db so they should merge fine, I can rebase this easily enough anyway

pjenvey:
Added a couple minor updates: fully moved _init_info into base and updated a couple docstrings

bbangert:
r+

PR #648, opened 2016-09-05
bug: label arguments for router.register to prevent misassignment

### LINKED ISSUE #642: SenderID is being dropped before going to GCM
https://sentry.prod.mozaws.net/operations/autopush-stage/issues/347170/

SenderID specified on URL, but is None in router/gcm.py

jrconlin:
Serial argument assignment can lead to misassignment of arguments in
some cases. It's best to be explicit where one can be.

closes #642


autopush/router/apnsrouter.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, ap_settings, router_conf):
         self._connect()
         self.log.debug("Starting APNS router...")
 
-    def register(self, uaid, router_data, *args, **kwargs):
+    def register(self, uaid, router_data={}, *args, **kwargs):
```
pjenvey: we really don't want the mutable default (router_data={}) here. router_data seems to be always be required, so it should continue being a positional argument to enforce that

bbangert:
r+

PR #656, opened 2016-09-09
fix: log all drop_user calls with record details when possible

### LINKED ISSUE #650: Log drop_user calls and reason
We should know exactly when and why we dropped a user record, along with a snap shot of what the user-record looked like.

bbangert:
Closes #650

### DIFF
--- a/autopush/router/webpush.py
+++ b/autopush/router/webpush.py
@@ -19,7 +19,11 @@ from twisted.web.client import FileBodyProducer
 from autopush.protocol import IgnoreBody
 from autopush.router.interface import RouterException, RouterResponse
 from autopush.router.simple import SimpleRouter
-from autopush.db import normalize_id
+from autopush.db import (
+    dump_uaid,
+    hasher,
+    normalize_id,
+)
 
 TTL_URL = "https://webpush-wg.github.io/webpush-protocol/#rfc.section.6.2"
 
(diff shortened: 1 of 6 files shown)


autopush/router/webpush.py
```diff
@@ -64,16 +68,18 @@ def preflight_check(self, uaid_data, channel_id):
         """Verifies this routing call can be done successfully"""
         uaid = uaid_data["uaid"]
         if 'current_month' not in uaid_data:
-            self.log.info("Record missing 'current_month' {record}",
-                          record=repr(uaid_data))
+            self.log.info(format="Dropping User", code=102,
```
jrconlin: I like that we log the action, should we still log the cause? It's an info so it may help.

bbangert: I numbered every code uniquely so we could identify where in the code-base we executed it from. The reason in a few cases is the same or similar, so knowing the origin of the call seemed more important than potentially the same reason in multiple places.

jrconlin: Ah, missed that the errno is different for each call.

jrconlin:
r+

autopush/web/webpush.py
```diff
@@ -63,8 +66,11 @@ def _router_completed(self, response, uaid_data, warning=""):
                 # An empty router_data object indicates that the record should
                 # be deleted. There is no longer valid route information for
                 # this record.
+                self.log.info(format="Dropping User", code=100,
```
pjenvey: This code isn't unique. I guess it really doesn't matter because it's in a separate file, but FYI

pjenvey:
r+

kitcambridge:
LGTM!

PR #657, opened 2016-09-09
feat: WIP ssl client cert

pjenvey:
Here's an early WIP of this to look at while I'm out next week.

Obviously there's pending cleanup/additions, so I'd like some feedback on the config options:

* Do we really need a "force client certs" option separate from a "only allow these certs"?

* I feel like sha256's would be enough for configuring allowed certs (vs supporting cert files too), thoughts?


autopush/base.py
```diff
@@ -45,3 +45,30 @@ def write_error(self, code, **kwargs):
...
+
+    def authenticate_peer_cert(self):
+        """Authenticate the client per the configured peer certs.
+
+        Aborts the request w/ a 401 on failure.
+
+        """
+        cert = self.request.connection.transport.getPeerCertificate()
+        if cert:
+            sha256 = cert.digest('sha256')
```
jrconlin: nit: can we call this something more explicit? (e.g. `cert_signature`?) Just think there's a small risk of collision with the hash function.

autopush/tests/certs/makecerts.py
```diff
@@ -0,0 +1,70 @@
...
+    cert = X509()
+    subject = cert.get_subject()
+
+    subject.C = b"TR"
+    subject.ST = b"Çorum"
+    subject.L = b"Başmakçı"
+    subject.CN = b"localhost"
+    subject.O = b"Mozilla Test"
+    subject.OU = b"Autopush Test %s" % filename
+    subject.emailAddress = b"pjenvey@mozilla.com"
```
jrconlin: trusting soul. Might want to make this "test@example.org" or some such.

autopush/base.py
```diff
@@ -45,3 +45,30 @@ def write_error(self, code, **kwargs):
...
+            auth = self.ap_settings.client_certs.get(sha256)
+            if auth is not None:
+                # TLS authenticated
+                self._client_info['tls_auth'] = auth
+                self._client_info['tls_auth_cn'] = cert.get_subject().CN
```
jrconlin: I'm guessing you're going to do the client cert signature checks here, or is that already being done?

jrconlin:
Ok, so you're requesting the client cert (which should validate and match one of the known set of signatures). I'm guessing you don't need the full private key in those cases. For sanity sake, we should probably also provide the signature of our cert to the remote parties so that they can do their own checks. Not super familiar with python's TLS stuff, so going to take a bit to better understand the provided hooks.

I agree that keeping private key PEM files off servers is generally A Good Thing™, but we're kinda hamstrung because we need to process the TLS locally instead of having a proxy do it for us.

jrconlin:
also looks like you might be missing a file `ssl.py` ?

autopush/base.py
pjenvey: At this point a client cert was required and verified. The signature check is 5 lines above, it's incredibly simple: -- if we don't have your verified cert's sha256, you fall out of the if block and are immediately 401'd.

PR #659, opened 2016-09-13
bug: Require auth header for endpoints with v2

### LINKED ISSUE #646: POSTing to a restricted subscription without VAPID headers should return a 401
STR:

1. Open http://mozilla-services.github.io/WebPushDataTestPage/?restricted=true.
2. Register the service worker, click "get subscription info", and encrypt the data, but don't add VAPID.
3. Send the data.

ER: We should return a 401 (Unauthorized) response with the `WWW-Authenticate: WebPush` header, because the credentials are missing.

AR: We return a 404, implying the subscription doesn't exist. The response is `{ "errno": 102, "message": "Request did not validate invalid token" }`.

jrconlin:
closes #646 

Also took the opportunity to fix the `(tuple) = function()` returns. 

@bbangert, @pjenvey r?

### DIFF
--- a/autopush/crypto_key.py
+++ b/autopush/crypto_key.py
@@ -34,7 +34,7 @@ http://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding-00#section-4
             ck_hash = {}
             for bit in bits:
                 try:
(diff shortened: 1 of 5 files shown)


bbangert:
r+

PR #660, opened 2016-09-14
feat: Add multiple cert handlers for APNs

### LINKED ISSUE #655: APNs bridge needs to allow multiple key pairs
Our APNs bridge only allows one key pair to be specified, but we have a different pair for each release channel: "Fennec", "Firefox", and "FirefoxBeta". I think we'll want something similar to the GCM bridge, which allows us to specify multiple sender IDs.

We currently ignore the `{app_id}` for APNs, so we can use that to determine the right key pair (for example, the URL would be `POST /v1/apns/FirefoxBeta/registration`). We can store that app ID in the router record, change the bridge config to accept multiple pairs, and have the router create an APNs client for each pair.

jrconlin:
This patch updates APNs handlers to accept platform based cert
configurations. See `configs/autopush_shared.ini.sample`. In addition,
this patch clarifies some argument references for routers (e.g. less
than useful `result` is now slightly more descriptive `uaid_data`)

Custom item names have been normalized to match gcm/fcm.
Document errors cleaned up a bit as well.

BREAKING CHANGE: the APNS configuration options have been altered, see
`configs/autopush_shared.ini.sample` for new APNS configuration
settings.

Closes #655 

@bbangert @pjenvey r?
@kitcambridge (optional r?)


autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+        self._config = router_conf
+        for channel in self._config.keys():
+            self.apns[channel] = self._connect(self._config[channel])
+        self._connect(router_conf)
         self.log.debug("Starting APNS router...")
         self.ap_settings = ap_settings
 
-    def register(self, uaid, router_data, *args, **kwargs):
-        """Validate that an APNs instance token is in the ``router_data``"""
+    def register(self, uaid, router_data, router_token="firefox",
```
kitcambridge: What do you think about calling this `app_id` to match the docs, and throwing instead of having a default?

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+        """Blocking APNS call to route the notification
+
+        :param notification: Notification data to send
+        :type notification: dict
+        :param router_data: Pre-initialized data for this connection
+        :type router_data: dict
+
+        """
+        router_token = router_data["token"]
+        self._channel = router_data["channel"]
```
kitcambridge: ISTM there's only [one instance](https://github.com/mozilla-services/autopush/blob/master/autopush/settings.py#L200) of `APNSRouter` per app, so this will clobber `self._channel` for concurrent requests (if one request comes in, then another one comes in while we're still processing the first one).

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+            alert=router_data.get("title", config.get('default_title',
+                                                      'Mozilla Push')),
+            content_available=1,
+            custom=custom)
         now = int(time.time())
-        self.messages[now] = {"token": token, "payload": payload}
-        # TODO: Add listener for error handling.
-        self.apns.gateway_server.register_response_listener(self._error)
+        self.messages[now] = {"token": router_token, "payload": payload}
+        apns_client.gateway_server.register_response_listener(self._error)
```
kitcambridge: Same here, `self.messages` and `register_response_listener` don't seem safe for concurrent use if `APNSRouter` is a singleton. `time.time()` is in seconds, so we might also clobber the previous entry if we handle two requests close together.

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+            "updates.client.bridge.apns.%s.attempted" % router_data["channel"],
             self._base_tags)
 
-        self.apns.gateway_server.send_notification(token, payload, now)
+        apns_client.gateway_server.send_notification(router_token, payload,
+                                                     now)
 
         # cleanup sent messages
         if self.messages:
             for time_sent in self.messages.keys():
```
bbangert: It's possible that multiple threads could be iterating on this at once, which means that one of the del calls will throw a KeyError. Would probably be better to toss the message cleanup function back into the event loop so it could clean-up without worry about locking the object.

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+        payload = apns.Payload(
+            alert=router_data.get("title", config.get('default_title',
+                                                      'Mozilla Push')),
+            content_available=1,
+            custom=custom)
         now = int(time.time())
-        self.messages[now] = {"token": token, "payload": payload}
-        # TODO: Add listener for error handling.
-        self.apns.gateway_server.register_response_listener(self._error)
+        self.messages[now] = {"token": router_token, "payload": payload}
```
bbangert: Is this a router_token, a channel_identifier, or a token? It's not great that its name keeps changing as it moves around to be more and more obscure in meaning.

configs/autopush_shared.ini.sample
```diff
@@ -115,3 +115,17 @@ endpoint_port = 8082
...
+; as well as any other bits of information that may be necessary. The format
+; is {"platform":
+;       {"cert": "path_to_cert_file",
+;        "key": "path_to_key_file",
+;        "default_title": "(optional) default alert title 'Mozilla Push'",
+;        "sandbox": False},  // Use the APNs sandbox feature
+;     ... }
+; e.g {"firefox": {"cert": "certs/main.cert", "key": "certs/main.key"}, "beta": {"cert": "certs/beta.cert", "key": "certs/beta.key"}}
+#apns_creds =
+
```
bbangert: The connection node needs this? Or is there some other reason its now in shared.ini?

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
         """Create a new APNS router and connect to APNS"""
         self.ap_settings = ap_settings
         self._base_tags = []
-        self.config = router_conf
-        self.default_title = router_conf.get("default_title", "SimplePush")
-        self.default_body = router_conf.get("default_body", "New Alert")
-        self._connect()
+        self.apns = dict()
+        self._config = router_conf
+        for channel in self._config.keys():
```
jrconlin: the iOS team have certs assigned to the firefox channel release, so these would map to "firefox", "beta", "gecko" and whatever other channel definition they want to use (e.g. "nightly?")

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
+        """Register an endpoint for APNS, on the `router_token` channel.
+
+        This will validate that an APNs instance token is in the
+        ``router_data``,
+
+        :param uaid: User Agent Identifier
+        :type uaid: str
+        :param router_data: Dict containing router specific configuration info
+        :type router_data: dict
+        :param router_token: The channel identifier for cert info lookup
```
jrconlin: Yep, resolving that. There will still be one "router_token" because that's what APNs calls it.

configs/autopush_shared.ini.sample
jrconlin: Consistency, mostly. The GCM/FCM definitions are located in shared. In addition, the router args are currently pulled in by main via `add_shared_args()`.

autopush/router/apnsrouter.py
jrconlin: `router_token` is now unique and locally defined as the `router_data["token"]` from above. I'm fine making it more explicit but `router_data_token = router_data["token"]` feels verbose.

autopush/router/apnsrouter.py
bbangert: Could that be better expressed by a better variable name or more documentation around this code?

jrconlin: changing to `rel_channel` as an abbreviation for "release channel".

autopush/router/apnsrouter.py
```diff
@@ -27,78 +27,129 @@ class APNSRouter(object):
...
-        payload = apns.Payload(alert=router_data.get("title",
-                                                     self.default_title),
-                               content_available=1,
-                               custom=custom)
+        payload = apns.Payload(
+            alert=router_data.get("title", config.get('default_title',
+                                                      'Mozilla Push')),
+            content_available=1,
+            custom=custom)
         now = int(time.time())
```
jrconlin: since I'm collapsing key indexes with the expry, made this nanosecond based.

autopush/router/apnsrouter.py
jrconlin: changed `router_token` to `app_id` for registration, added `app_id` as a default arg for `register`

autopush/router/apnsrouter.py
jrconlin: moved `rel_channel` to `messages` hash.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +27,147 @@ class APNSRouter(object):
...
-            for time_sent in self.messages.keys():
-                if time_sent < now - self.config.get("expry", 10):
-                    del self.messages[time_sent]
+        payload = apns.Payload(
+            alert=router_data.get("title", config.get('default_title',
+                                                      'Mozilla Push')),
+            content_available=1,
+            custom=custom)
+        now = time.time()
+        self.messages[now] = {"rel_channel": router_data["rel_channel"],
```
kitcambridge: Is `rel_channel` the same as `app_id`, or is it the `instance_id` from `{“token”:{instance_id}}`? And is `router_token` also the `instance_id`?

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,162 @@ class APNSRouter(object):
...
+            "router_token": router_token,
+            "payload": payload}
         self.ap_settings.metrics.increment(
-            "updates.client.bridge.apns.succeed",
+            "updates.client.bridge.apns.%s.attempted" %
+            router_data["rel_channel"],
             self._base_tags)
+
+        apns_client.gateway_server.send_notification(router_token, payload,
+                                                     now)
```
kitcambridge: I think you meant to pass the ID here, too, instead of `now`?

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,162 @@ class APNSRouter(object):
...
+        for msg_id in keys:
+            message = self.messages[msg_id]
+            expry = self._config[message['rel_channel']].get("expry", 10)
+            if message["time_sent"] < now - expry:
+                del self.messages[msg_id]
+        return response
+
     def _error(self, err):
         """Error handler"""
         if err['status'] == 0:
```
kitcambridge: Nit: `self.messages.pop(err['identifier'], None)` might be better; `del` will throw a `KeyError` if the entry was already removed from the dict.

bbangert: That shouldn't now that cleanup is run off the event loop.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
-        self.default_title = router_conf.get("default_title", "SimplePush")
-        self.default_body = router_conf.get("default_body", "New Alert")
-        self._connect()
-        self.log.debug("Starting APNS router...")
+        self.apns = dict()
+        self.messages = dict()
+        self._config = router_conf
+        if self._config.get('max_messages'):
+            self._max_messages = self._config['max_messages']
+            del(self._config['max_messages'])
```
bbangert: del the function vs. del the keyword?

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
+        else:
+            self._max_messages = 100
+        for rel_channel in self._config.keys():
+            self.apns[rel_channel] = self._connect(self._config[rel_channel])
+            self.apns[rel_channel].gateway_server.register_response_listener(
+                self._error)
+        self._connect(router_conf)
```
kitcambridge: Is this still needed, now that the args are a dict?

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
+        """Connect to APNS
+
+        :param cert_info: APNS certificate configuration info
+        :type cert_info: dict
+
+        returns an instance of APNs that can be stored under the proper
+        release channel name.
+
+        """
+        # Do I still need to call this in _error?
```
kitcambridge: I guess not. 😄

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
         self._base_tags = []
-        self.config = router_conf
-        self.default_title = router_conf.get("default_title", "SimplePush")
-        self.default_body = router_conf.get("default_body", "New Alert")
-        self._connect()
-        self.log.debug("Starting APNS router...")
+        self.apns = dict()
+        self.messages = dict()
+        self._config = router_conf
+        if self._config.get('max_messages'):
```
bbangert:
```python
self._max_messages = self._config.pop('max_messages', 100)
```

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
+        for rel_channel in self._config.keys():
+            self.apns[rel_channel] = self._connect(self._config[rel_channel])
+            self.apns[rel_channel].gateway_server.register_response_listener(
+                self._error)
+        self._connect(router_conf)
         self.ap_settings = ap_settings
+        # "unclean" is used by testing to prevent unclean reactor errors.
+        # It also prevents the local cleanup loop from running without
+        # requiring heavy lifting.
+        if "unclean" not in router_conf:
```
kitcambridge: Ugh. Unclean reactor errors make me sad. 😭

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
-        self.log.debug("Starting APNS router...")
+        self.apns = dict()
+        self.messages = dict()
+        self._config = router_conf
+        if self._config.get('max_messages'):
+            self._max_messages = self._config['max_messages']
+            del(self._config['max_messages'])
+        else:
+            self._max_messages = 100
+        for rel_channel in self._config.keys():
```
bbangert: Default behavior on iteration of a dict is to iterate through its keys. Usually keys() is only needed if you might mutate the dict during iteration such that you need a list first.

kitcambridge:
Great! r+ with nits fixed.

autopush/router/apnsrouter.py
jrconlin: I'd like to keep the connection in a function, even if it adds some slight overhead. There's a bit of messy history about that with the apns lib.

autopush/router/apnsrouter.py
jrconlin: same here, thus why I added this to just not call the reactor when we're running tests.

autopush/router/apnsrouter.py
jrconlin: so, yeah, remember my cryptic comment about calling self._connect() before? ...

autopush/router/apnsrouter.py
kitcambridge: I meant you're calling `self._connect()` for each entry in `router_conf`, and then you're calling `self._connect(router_conf)`. ISTM that won't work? If it does, I'd like to understand why. 😄

bbangert: So, first router_conf is assigned to self._config, then we iterate through keys in self._config (which is now router_conf minux the max_messages key) calling self._connect. But then we also call self._connect with the whole thing again? What's going on here?

jrconlin: ah, quite right. context is a good thing for me to have when I reply. I should try  it more.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
+            self.apns[rel_channel] = self._connect(self._config[rel_channel])
+            self.apns[rel_channel].gateway_server.register_response_listener(
+                self._error)
+        self._connect(router_conf)
         self.ap_settings = ap_settings
+        # "unclean" is used by testing to prevent unclean reactor errors.
+        # It also prevents the local cleanup loop from running without
+        # requiring heavy lifting.
+        if "unclean" not in router_conf:
+            reactor.callLater(10, self._cleanup)
```
bbangert: There's a task looping call used in main.py for functions that should be called regularly.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +28,175 @@ class APNSRouter(object):
...
+        ``router_data``,
+
+        :param uaid: User Agent Identifier
+        :type uaid: str
+        :param router_data: Dict containing router specific configuration info
+        :type router_data: dict
+        :param app_id: The release channel identifier for cert info lookup
+        :type app_id: str
+
+        returns a modified router_data dict to be stored
```
bbangert: ``:returns:`` ?

autopush/router/apnsrouter.py
jrconlin: ah-HA! turns out this was a merge artifact, since it just happened again after a conflict from master. ._.

jrconlin: git <strike>blend</strike> merge failure.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +27,164 @@ class APNSRouter(object):
...
+
+        :param notification: Notification data to send
+        :type notification: dict
+        :param router_data: Pre-initialized data for this connection
+        :type router_data: dict
+
+        """
+        router_token = router_data["token"]
+        rel_channel = router_data["rel_channel"]
+        config = self._config[rel_channel]
```
bbangert: Sure hope we never, ever, change valid release channel names.

autopush/router/apnsrouter.py
```diff
@@ -27,86 +27,164 @@ class APNSRouter(object):
...
+        # "apns-id"
+        msg_id = str(uuid.uuid4())
+        self.messages[msg_id] = {
+            "time_sent": now,
+            "rel_channel": router_data["rel_channel"],
+            "router_token": router_token,
+            "payload": payload}
         self.ap_settings.metrics.increment(
-            "updates.client.bridge.apns.succeed",
+            "updates.client.bridge.apns.%s.attempted" %
```
pjenvey: FYI send_notification appears to never fail w/ enhanced=True (assuming the input is valid for a payload), which if I'm right makes the new separate metrics unnecessary

autopush/router/apnsrouter.py
```diff
@@ -27,86 +27,164 @@ class APNSRouter(object):
...
+        :param notification: Notification data to send
+        :type notification: dict
+        :param uaid_data: User Agent specific data
+        :type uaid_data: dict
+
+        """
         router_data = uaid_data["router_data"]
         # Kick the entire notification routing off to a thread
-        return deferToThread(self._route, notification, router_data)
+        d = deferToThread(self._route, notification, router_data)
```
pjenvey: (since I know you're touching this again) could you please restore: return deferToThread..

autopush/router/apnsrouter.py
jrconlin: If we do, thumbscrews. Fortunately ops has the ability to generate more certs. They'll just have to change the app_id they're sending in to use the new cert pair.

PR #666, opened 2016-09-22
fix: remove extra jws_err handler which caused extra write after finish

### LINKED ISSUE #665: Write after finish called because of extra jws error handler
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/351161/

jrconlin:
For bogus or invalid Authorization headers, an extra write could occur
after finish.

closes #665

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -569,9 +569,5 @@ def endpoint_main(sysargs=None, use_files=True):
     # Start the table rotation checker/updater
     l = task.LoopingCall(settings.update_rotating_tables)
     l.start(60)
-    if settings.routers.get('apns'):
-        l = task.LoopingCall(settings.routers['apns']._cleanup)
-        l.start(10)
-
     reactor.suggestThreadPoolSize(50)
     reactor.run()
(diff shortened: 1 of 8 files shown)


autopush/endpoint.py
```diff
@@ -342,9 +342,12 @@ def _store_auth(self, jwt, crypto_key, token, result):
     def _invalid_auth(self, fail):
         if isinstance(fail.value, VapidAuthException):
             raise fail.value
+        if isinstance(fail.value, JOSEError):
```
bbangert: This could be added to the isinstance check above, it can take a tuple of types to check.

autopush/endpoint.py
```diff
@@ -342,9 +342,12 @@ def _store_auth(self, jwt, crypto_key, token, result):
...
+            raise fail.value
         message = fail.value.message or repr(fail.value)
         if (isinstance(fail.value, AssertionError) or
```
pjenvey: let's collapse this into one isinstance call while you're in here

autopush/endpoint.py
jrconlin: Yeah, didn't know about the tuple thing. Fixing up the other isinstance set to do the same.

jrconlin:
The hell?! this wasn't merged?

Issue #662, opened 2016-09-17
Move to APNs http/2 API


pjenvey:
The old API is incredibly bad. Apple recommends "All developers should migrate their remote notification provider servers to the more capable and more efficient HTTP/2-based API".

How apns.py and our apnsrouter.py have to handle the retry logic is particularly bad and probably not all that scalable. We don't even utilize its silly feedback server for potential retries.

The new HTTP/2 API should improve this. However I'm not sure there's a reliable enough python lib out there yet to switch to..

jrconlin:
honestly? I'd prefer to use HTTP2. There was some discussion earlier about using SNS to effectively wrap the APNs calls, but that introduces its own level of overhead and complexity. (Same with FCM for what it's worth.) There's a weird cottage industry of having other things deal with APNs, mostly because all of the hassle around trying to send messages to them.

It may be time for me to renew my Apple Dev license and just write a new APNs library.

kitcambridge:
Connections to the old API can also [wedge](https://github.com/Voxer/node-apn/commit/527a68de33087ca548dbce20f7ccf0b7394ffb25#commitcomment-9610565).

jrconlin:
That's... Sigh...

PR #667, opened 2016-09-22
feat: Switch to new http2 based APNS protocol

jrconlin:
Existing libraries either lack tests or are not good matches for
twisted. Created a very light weight apns2 library that does connection
pooling and sends out the new notification. Connection pooling is
strongly suggested by APNs docs.

closes: #662

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -1,0 +1,147 @@
+import json
+from collections import deque
+
+import hyper.tls
+from hyper import HTTP20Connection
+from hyper.http20.exceptions import HTTP20Error
+
+from autopush.router.interface import RouterException
+
+SANDBOX = 'api.development.push.apple.com'
+SERVER = 'api.push.apple.com'
+
+APNS_PRIORITY_IMMEDIATE = 10
+APNS_PRIORITY_LOW = 5
+APNS_MAX_CONNECTIONS = 20
+
+
+class APNSException(Exception):
+    pass
+
+
+class APNSClient(object):
+    def __init__(self, cert_file, key_file, topic,
+                 alt=False, use_sandbox=False,
+                 max_connections=APNS_MAX_CONNECTIONS,
+                 logger=None, metrics=None):
+        """Create the APNS client connector.
+
+        The cert_file and key_file can be derived from the exported `.p12`
+        **Apple Push Services: *bundleID*** key contained in the **Keychain
+        Access** application. To extract the proper PEM formatted data, you
+        can use the following commands:
+
+        ```
+        openssl pkcs12 -in file.p12 -out apns_cert.pem -clcerts -nokeys
+        openssl pkcs12 -in file.p12 -out apns_key.pem -nocerts -nodes
+        ```
+
+        The *topic* is the Bundle ID of the bridge recipient iOS application.
+        Since the cert needs to be tied directly to an application, the topic
+        is usually similar to "com.example.MyApplication".
+
+        :param cert_file: Path to the PEM formatted APNs certification file.
+        :type cert_file: str
+        :param key_file: Path to the PEM formatted APNs key file.
+        :type key_file: str
+        :param topic: The *Bundle ID* that identifies the assoc. iOS app.
+        :type topic: str
+        :param alt: Use the alternate APNs publication port (if 443 is blocked)
+        :type alt: bool
+        :param use_sandbox: Use the development sandbox
+        :type use_sandbox: bool
+        :param max_connections: Max number of pooled connections to use
+        :type max_connections: int
+        :param logger: Status logger
+        :type logger: logger
+        :param metrics: Metric recorder
+        :type metrics: autopush.metrics.IMetric
+        """
+        self.server = SANDBOX if use_sandbox else SERVER
+        self.port = 2197 if alt else 443
+        self.ssl_context = hyper.tls.init_context(cert=(cert_file, key_file))
+        self.log = logger
+        self.metrics = metrics
+        self.topic = topic
+        self._max_connections = max_connections
+        self.connections = deque(maxlen=max_connections)
+        if self.log:
+            self.log.info("Starting APNS connection")
+
+        self.connections.extendleft((HTTP20Connection(
+            self.server,
+            self.port,
+            ssl_context=self.ssl_context,
+            force_proto='h2') for x in range(0, max_connections)))
+
+    def send(self, router_token, payload, apns_id,
+             priority=True, topic=None, exp=None):
+        """Send the dict of values to the remote bridge
+
+        This sends the raw data to the remote bridge application using the
+        APNS2 HTTP2 API.
+
+        :param router_token: APNs provided hex token identifying recipient
+        :type router_token: str
+        :param payload: Data to send to recipient
+        :type payload: dict
+        :param priority: True is high priority, false is low priority
+        :type priority: bool
+        :param topic: BundleID for the recipient application (overides default)
+        :type topic: str
+        :param exp: Message expiration timestamp
+        :type exp: timestamp
+
+        """
+        body = json.dumps(payload)
+        priority = APNS_PRIORITY_IMMEDIATE if priority else APNS_PRIORITY_LOW
+        headers = {
+            'apns-id': apns_id,
+            'apns-priority': priority,
+            'apns-topic': topic or self.topic,
+        }
+        if exp:
+            headers['apns-expiration'] = exp
+        url = '/3/device/' + router_token
+        connection = self._get_connection()
+        try:
+            # request auto-opens closed connections, so if a connection
+            # has timed out or failed for other reasons, it's automatically
+            # re-established.
+            stream_id = connection.request(
+                'POST', url=url, body=body, headers=headers)
+            response = connection.get_response(stream_id)
+            if response.status != 200:
+                reason = json.loads(response.read().decode('utf-8'))['reason']
+                raise RouterException(
+                    "APNS Transmit Error {}:{}".format(response.status,
+                                                       reason),
+                    status_code=500,
+                    response_body="APNS could not process "
+                                  "your message {}".format(reason))
+        except HTTP20Error as ex:
+            connection.close()
+            raise RouterException(
+                "APNS Processing error: {}".format(repr(ex)),
+                status_code=503,
+                response_body="APNS returned an error processing request",
+            )
+        finally:
+            # Returning a closed connection to the pool is ok.
+            # hyper will reconnect on .request()
+            self._return_connection(connection)
+
+    def _get_connection(self):
+        try:
+            connection = self.connections.pop()
+            return connection
+        except IndexError:
+            raise RouterException(
+                "Too many APNS requests, increase pool from {}".format(
+                    self._max_connections
+                ),
+                status_code=503,
+                response_body="APNS busy, please retry")
+
+    def _return_connection(self, connection):
+        self.connections.appendleft(connection)
--- a/autopush/tests/test_router.py
+++ b/autopush/tests/test_router.py
@@ -2,17 +2,20 @@
 from unittest import TestCase
 import uuid
 import time
+import json
 
 from mock import Mock, PropertyMock, patch
 from moto import mock_dynamodb2
-from nose.tools import eq_, ok_
+from nose.tools import eq_, ok_, assert_raises
 from twisted.trial import unittest
 from twisted.internet.error import ConnectError, ConnectionRefusedError
-from twisted.python.failure import Failure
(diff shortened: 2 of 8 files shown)


autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        :param cert_file: Path to the PEM formatted APNs certification file.
+        :type cert_file: str
+        :param key_file: Path to the PEM formatted APNs key file.
+        :type key_file: str
+        :param topic: The *Bundle ID* that identifies the assoc. iOS app.
+        :type topic: str
+        :param alt: Use the alternate APNs publication port (if 443 is blocked)
+        :type alt: bool
+        :param use_sandbox: Use the development sandbox
+        :type alt: bool
```
pjenvey: :type use_sandbox

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        :type payload: dict
+        :param priority: True is high priority, false is low priority
+        :type priority: bool
+        :param topic: BundleID for the recipient application (overides default)
+        :type topic: str
+        :param exp: Message expiration timestamp
+        :type exp: timestamp
+
+        """
+        if not apns_id:
```
pjenvey: this feels unnecessary

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+            apns_id = str(uuid.uuid4()).lower()
+        body = json.dumps(payload)
+        if priority:
```
pjenvey: priority = APNS_PRIORITY_IMMEDIATE if priority else APNS_PRIORITY_LOW

I think it fits!?

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        self.metrics = metrics
+        self.topic = topic
+        self._max_connections = max_connections
+        self._min_connections = min_connections
+        self.connections = deque(maxlen=max_connections)
+        if self.log:
+            self.log.info("Starting APNS connection")
+        # HTTP20Connection is not thread safe, so it's best to keep processing
+        # of the result in the same thread.
+        for i in xrange(0, min_connections):
```
pjenvey: this pool logic here isn't quite right. this would create them and immediately throw them away (they're not added to self.connections)

max_connections doesn't seem to be really utilized either -- (the appendleft is guarded by len check vs min_connections)..

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+                force_proto='h2')
+
+    def _return_connection(self, connection):
+        if len(self.connections) < self._min_connections:
+            self.connections.appendleft(connection)
+        else:
+            connection.close()
+
+    def _process_send(self, connection, url, body, headers):
+        # HTTP20Connection is not thread safe. It's best to keep this and
```
kitcambridge: This comment about thread safety confuses me a bit, since you're creating the connection on the main thread and using it from a background thread with `deferToThread`.

Do you mean that it can't be used concurrently from multiple threads, or that it must only be used on the thread it was created?

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        # related functions in the same thread instance.
+        # Should these be pooled?
+        stream_id = connection.request(
```
kitcambridge: Do you need to wrap this in a `try...finally` block that closes the connection if there's an error...or does the library handle that for you? The docs for `close` are a bit vague: "In some cases this method will automatically be called."

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        :param priority: True is high priority, false is low priority
+        :type priority: bool
+        :param topic: BundleID for the recipient application (overides default)
+        :type topic: str
+        :param exp: Message expiration timestamp
+        :type exp: timestamp
+
+        """
+        if not apns_id:
+            apns_id = str(uuid.uuid4()).lower()
```
bbangert: Isn't it already lower-cased?

autopush/router/apns2.py
jrconlin: Honestly, it's a confusion point for me too. Looking over the docs and examples, it appears that remotely closed connections automatically close. Wrapping it with a try handler is probably safer so I'll do that.

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        connection = self._get_connection()
+        d = deferToThread(
+            self._process_send,
+            connection=connection,
+            url=url,
+            body=body,
+            headers=headers,
+        )
+        # failed connections self-close
+        d.addCallback(self._return_connection)
```
bbangert: Would probably be best to give the connection to the callback here. Also, it should be addBoth to ensure the connection is always returned even if some other random error happens. Or is it intentional that the connection is only returned if it works out ok?

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            return self.connections.pop()
+        except IndexError:
+            return HTTP20Connection(
+                self.server, self.port,
+                ssl_context=self.ssl_context,
+                force_proto='h2')
+
+    def _return_connection(self, connection):
+        if len(self.connections) < self._min_connections:
```
bbangert: Shouldn't this be max_connections its checking against?

autopush/router/apns2.py
```diff
@@ -0,0 +1,152 @@
...
+        headers = {
+            'apns-id': apns_id,
+            'apns-priority': priority,
+            'apns-topic': topic or self.topic,
+        }
+        if exp:
+            headers['apns-expiration'] = exp
+        url = '/3/device/' + router_token
+        connection = self._get_connection()
+        d = deferToThread(
```
bbangert: The send function is called from a thread already, why is it being punted to yet another thread?

autopush/router/apns2.py
jrconlin: It can't be used concurrently. The docs note that if you're careful, you can use the stream_id identifier to allow it to be used, but still note caution should be applied. Using the same connection for the duration of a thread's requirement seems the safest method.

autopush/router/apns2.py
jrconlin: They are at the moment, yes. Since APNS requires this to be lower case, I'm making sure they stay that way in the future.

autopush/router/apns2.py
jrconlin: If there's a problem with the connection, it should be dropped. APNS would close it anyway. Creating a new connection is reasonably cheap.

autopush/router/apns2.py
jrconlin: relic from when I was considering making this a stand-alone. removing the function.

autopush/router/apns2.py
kitcambridge: OK, so it's all right to create on one thread, pass it off to another thread, and hand it back when it's done?

autopush/router/apns2.py
bbangert: I can't imagine Python would ever ever change this, as that would probably break the world.

autopush/router/apns2.py
jrconlin: So long as it's not being used by two separate threads at the same time, sure.

autopush/router/apns2.py
jrconlin: Not so sure about that. Python uses UUID as an object, which encodes a numeric to a formatted hex string. The case is irrelevant (and honestly, is for most applications that use a UUID) Changing between dashed and undashed could cause problems as we know, but changing case shouldn't make any difference. I was taken back a bit by the requirement that this be presented as lowercase.

I'll remove the `.lower()` for now.

autopush/router/apns2.py
pjenvey: I meant that send shouldn't handle a nonzero apns_id (act like it's a required arg). The router and tests always generate one and pass it along

autopush/router/apns2.py
```diff
@@ -0,0 +1,144 @@
...
+                    response_body="APNS could not process "
+                                  "your message {}".format(reason))
+            return
+        except HTTP20Error as ex:
+            connection.close()
+            raise RouterException(
+                "APNS Processing error: {}".format(repr(ex)),
+                status_code=503,
+                response_body="APNS returned an error processing request",
+            )
```
kitcambridge: Does it make sense to open another connection here, instead of returning the closed one to the pool?

autopush/router/apns2.py
```diff
@@ -0,0 +1,144 @@
...
+        :type use_sandbox: bool
+        :param max_connections: Max number of pooled connections to use
+        :type max_connections: int
+        :param logger: Status logger
+        :type logger: logger
+        :param metrics: Metric recorder
+        :type metrics: autopush.metrics.IMetric
+        """
+        self.server = SERVER if not use_sandbox else SANDBOX
+        self.port = 443 if not alt else 2197
```
kitcambridge: Nit: I think removing the negative would read a bit easier (`2197 if alt else 443`), but it's up to you.

kitcambridge:
So much nicer than the crufty artisanal API!

autopush/router/apns2.py
jrconlin: Meh, since these are altering i'd prefer the first value be the most likely value.

autopush/router/apns2.py
jrconlin: There's an exceptionally good chance that the connection would be auto-closed by whatever triggered the HTTP20Error. Closing the connection is still a good idea in any case since the connection is likely corrupted. If the connection is already closed, this is a no-op. Connections are automatically re-established by `request()`.

autopush/router/apns2.py
```diff
@@ -0,0 +1,144 @@
...
+                "APNS Processing error: {}".format(repr(ex)),
+                status_code=503,
+                response_body="APNS returned an error processing request",
+            )
+        finally:
+            self._return_connection(connection)
+
+    def _get_connection(self):
+        try:
+            connection = self.connections.pop()
```
pjenvey: let's return here

autopush/router/apns2.py
```diff
@@ -0,0 +1,144 @@
...
+            return
+        except HTTP20Error as ex:
+            connection.close()
+            raise RouterException(
+                "APNS Processing error: {}".format(repr(ex)),
+                status_code=503,
+                response_body="APNS returned an error processing request",
+            )
+        finally:
+            self._return_connection(connection)
```
bbangert: This will execute before the except clause re-raises the exception. Which means the closed connection will be returned to the pool.

bbangert: I see above the closed connection will be re-opened automagically, so that's fine. Maybe just a comment indicating that's ok so we don't forget in the future and think this is a problem?

autopush/tests/test_router.py
```diff
@@ -81,18 +84,46 @@ def init(self, settings, router_conf):
 
 class APNSRouterTestCase(unittest.TestCase):
 
-    @patch("twisted.internet.reactor.callLater")
-    def setUp(self, cl):
+    def _waitfor(self, func):
+            times = 0
+            while not func():  # pragma: nocover
+                time.sleep(1)
```
pjenvey: I'm guessing this is rarely hit but can we sleep less, anyway?

autopush/router/apns2.py
```diff
@@ -0,0 +1,148 @@
...
+        self.ssl_context = hyper.tls.init_context(cert=(cert_file, key_file))
+        self.log = logger
+        self.metrics = metrics
+        self.topic = topic
+        self._max_connections = max_connections
+        self.connections = deque(maxlen=max_connections)
+        if self.log:
+            self.log.info("Starting APNS connection")
+
+        for i in xrange(0, max_connections):
```
pjenvey: can we appendleft here just to be consistent w/ return?

in fact we can extendleft with a generator statement instead of a for loop (I loved extending).. and let's just use range() instead of xrange #barelypython3compat

autopush/router/apns2.py
```diff
@@ -0,0 +1,144 @@
...
+            headers['apns-expiration'] = exp
+        url = '/3/device/' + router_token
+        connection = self._get_connection()
+        try:
+            # request auto-opens closed connections.
+            stream_id = connection.request(
+                'POST', url=url, body=body, headers=headers)
+            response = connection.get_response(stream_id)
+            if response.status != 200:
+                reason = json.loads(response.read().decode('utf-8'))['reason']
```
jrconlin: If not, then APNS is horribly broken and the JSON is probably corrupted as well. Might as well blow up early with something that'll get tossed into sentry.

PR #670, opened 2016-09-26
feat: add webpush topics

### LINKED ISSUE #643: Add WebPush Topics
This will require a new keying scheme for the DynamoDB sort key. Right now the sort key is:

``chidmessageid`` - CHID : Encrypted(UAID : CHID)

This also means that our messages are in arbitrary orders on retrieval.

There are some constraints on keys to consider:

1. Keys shouldn't be too long, as they need to be sent to/from the client for every message
2. We must be able to locate a message of a given topic for delete/replacement. Ideally we can replace any existing message of the same topic to avoid 2 writes per Topic'd message (as a delete counts as a write).
3. Migration from the existing sort key scheme must be transparent.

A proposed scheme for sort keys originally failed to account for the need to possibly DELETE the message. Here's a new scheme that might work:

For topic messages only:

``VERSION : CHID : HASH(TOPIC)``

When a topic-bound message is created, we create a URL resource for a Location like so:

``Encrypted(VERSION : UAID: CHID : HASH(TOPIC))``

This is stored as the update_id for the message. Since we use a nonce-based crypto scheme, it changes every time we update it (as desired to reflect new resource creation). We can also locate the message, and ensure the update_id matches before deleting it.

For normal messages:

``VERSION: CHID: Encrypted(VERSION : UAID : CHID)``

Note that we must now include the VERSION so we know when we're using the newly versioned token scheme.

bbangert:
Add's webpush topics with versioned sort key.

Closes #643

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -1,4 +1,34 @@
-"""Database Interaction"""
+"""Database Interaction
+
+WebPush Sort Keys
+-----------------
+
+Messages for WebPush are stored using a partition key + sort key, originally
+the sort key was:
+
+    CHID : Encrypted(UAID: CHID)
+
+The encrypted portion was returned as the Location to the Application Server.
+Decrypting it resulted in enough information to create the sort key so that
+the message could be deleted and located again.
+
+For WebPush Topic messages, a new scheme was needed since the only way to
+locate the prior message is the UAID + CHID + Topic. Using Encryption in
+the sort key is therefore not useful since it would change every update.
+
+The sort key scheme for WebPush messages is:
+
+    VERSION : CHID : TOPIC
+
+To ensure updated messages are not deleted, each message will still have an
+update-id key/value in its item.
+
+Non-versioned messages are assumed to be original messages from before this
+scheme was adopted.
+
+``VERSION`` is a 2-digit 0-padded number, starting at 01 for Topic messages.
+
+"""
 from __future__ import absolute_import
 
 import datetime
@@ -19,7 +49,11 @@ from boto.dynamodb2.table import Table, Item
 from boto.dynamodb2.types import NUMBER
 
 from autopush.exceptions import AutopushException
-from autopush.utils import generate_hash
+from autopush.utils import (
+    generate_hash,
+    normalize_id,
+    WebPushNotification,
+)
 
 key_hash = ""
 TRACK_DB_CALLS = False
@@ -61,16 +95,6 @@ def dump_uaid(uaid_data):
         return repr(uaid_data.items())
     else:
         return repr(uaid_data)
-
-
-def normalize_id(ident):
-    if (len(ident) == 36 and
-            ident[8] == ident[13] == ident[18] == ident[23] == '-'):
-        return ident.lower()
-    raw = filter(lambda x: x in '0123456789abcdef', ident.lower())
-    if len(raw) != 32:
-        raise ValueError("Invalid UUID")
-    return '-'.join((raw[:8], raw[8:12], raw[12:16], raw[16:20], raw[20:]))
 
 
 def make_rotating_tablename(prefix, delta=0, date=None):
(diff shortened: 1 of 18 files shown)


autopush/endpoint.py
```diff
@@ -547,6 +554,20 @@ def _uaid_lookup_results(self, uaid_data):
...
+                        "characters"
+                    )
+                    return
+
+                if not VALID_BASE64_URL.match(topic):
+                    self._write_response(
+                        400, 113, message="Topic must be URL and Filename "
+                        "safe Base64 alphabet"
+                    )
+
```
pjenvey: forgot a return here

autopush/web/validation.py
```diff
@@ -35,6 +35,9 @@
 AUTH_SCHEMES = ["bearer", "webpush"]
 PREF_SCHEME = "webpush"
 
+# Base64 URL validation
+VALID_BASE64_URL = re.compile(r'^[0-9A-Za-z\-_=]*$')
```
jrconlin: is `foo=bar` a valid topic name? maybe `r'^[a-zA-Z0-9\-_]+=*$'` to more closely match Base64?

kitcambridge: We could simplify it to `r'^[a-zA-Z0-9\-_]+$'`; `=` isn't valid, anyway.

autopush/utils.py
```diff
@@ -2,15 +2,28 @@
...
+    attrs,
+    attrib
+)
 from jose import jwt
 from ua_parser import user_agent_parser
 
 
+# Remove trailing padding characters from complex header items like
+# Crypto-Key and Encryption
+strip_padding = re.compile('=+(?=[,;]|$)')
```
pjenvey: capitalize

autopush/endpoint.py
```diff
@@ -411,14 +415,25 @@ def delete(self, token):
...
+        d.addCallback(self._delete_message, token)
         d.addErrback(self._token_err)
         d.addErrback(self._response_err)
         return d
 
-    def _delete_message(self, kind, uaid, chid):
-        d = deferToThread(self.ap_settings.message.delete_message, uaid,
-                          chid, self.version)
+    def _delete_message(self, result, token):
+        key_info = WebPushNotification.parse_decrypted_message_id(result)
```
pjenvey: maybe this method could be notif = WebPushNotification.from_decrypted_...(token, result)

bbangert: I mainly wanted to have a single method that could invert what the generate_message_id method does that encrypts it. It doesn't make a complete proper WebPushNotification as it leaves out bits that aren't needed for deleting it.

autopush/utils.py
```diff
@@ -198,3 +223,272 @@ def parse_user_agent(agent_string):
...
+        )
+
+    def cleanup_headers(self):
+        """Sanitize the headers for this notification
+
+        This only needs to be run when creating a notification from passed
+        in application server headers.
+
+        """
+        self.headers = sanitize_webpush_headers(self.headers)
```
jrconlin: We seem to only be using this function to call `sanitize_webpush_headers`. Can we collapse a function?

autopush/tests/test_db.py
```diff
@@ -301,106 +313,45 @@ def test_message_storage(self):
...
-
-        # Delete them all
-        message.delete_messages_for_channel(self.uaid, chid)
-
-        # Verify they're gone
-        all_messages = list(message.fetch_messages(self.uaid, limit=100))
-        eq_(len(all_messages), 0)
-
     def test_message_delete_fail_condition(self):
+        notif = WebPushNotification(
```
jrconlin: Why not call `make_webpush_notification` and add the extra fields? (or since they're None, adding them to the function?)

autopush/endpoint.py
```diff
@@ -547,6 +554,20 @@ def _uaid_lookup_results(self, uaid_data):
...
+                    self._write_response(
+                        400, 113, message="Topic must be no greater than 32 "
+                        "characters"
+                    )
+                    return
+
+                if not VALID_BASE64_URL.match(topic):
+                    self._write_response(
+                        400, 113, message="Topic must be URL and Filename "
+                        "safe Base64 alphabet"
```
bbangert: This is what the WebPush spec calls it, so I retained their naming exactly.

jrconlin: Ah, ok.

autopush/db.py
```diff
@@ -263,6 +287,13 @@ def generate_last_connect():
     return int(val)
 
 
+def parse_webpush_sort_key_for_delivery(sort_key):
```
pjenvey: kill?

PR #671, opened 2016-09-26
fix: remove extra jws_err handler which caused extra write after finish

### LINKED ISSUE #665: Write after finish called because of extra jws error handler

jrconlin:
For bogus or invalid Authorization headers, an extra write could occur
after finish.

closes #665


jrconlin:
So, yeah, #666 got closed, but not merged.

Issue #465, opened 2016-05-09
Refactor main.py into well structured server setup and separate CLI parser


bbangert:
The CLI and config reading is entangled with the server object setup, this makes it harder to test and identify the components for modification.

Example code that is laid out nicer:
https://github.com/andrewgodwin/daphne
https://github.com/donnemartin/haxor-news/tree/master/haxor_news

PR #673, opened 2016-09-26
feat: add a new client_certs endpoint config option

### LINKED ISSUE #498: Add SSL client cert option to endpoint
Our endpoint should have a separate config option to require SSL client certs, along with a config option to indicate acceptable SSL client certs.

The endpoint should include the domain for SSL client certs used in the push message logging.

pjenvey:
requires whitelisted TLS client certificates for connections to the
endpoint except to health/log_check handlers

closes #498


autopush/tests/test_main.py
```diff
@@ -284,6 +287,40 @@ def test_bad_apnsconf(self):
...
+            "--ssl_key=keys/server.key",
+            """--client_certs={"foo": "bar"}"""
+        ], False)
+
+    @patch('hyper.tls', spec=hyper.tls)
+    def test_client_certs_parse(self, mock):
+        ap = make_settings(self.TestArg)
+        eq_(ap.client_certs, dict(ABCD="partner1",
+                                  EFGH="partner2",
+                                  IJKL="partner2"))
```
jrconlin: Include a nameless cert to match above condition?

autopush/main.py
```diff
@@ -342,6 +345,33 @@ def make_settings(args, **kwargs):
...
+        except (ValueError, TypeError):
+            log.critical(format="Invalid JSON specified for client_certs")
+            return
+        if client_certs_arg:
+            if not args.ssl_key:
+                log.critical(format="client_certs specified without SSL "
+                                    "enabled (no ssl_key specified)")
+                return
+            client_certs = {}
+            for name, sigs in client_certs_arg.iteritems():
```
jrconlin: So from the config file, each cert is identified as `{'name':['sig'] ...}`, where 'sig' has a definite format. Do you want to verify that the sigs provided match that format at least?

autopush/tests/certs/makecerts.py
```diff
@@ -0,0 +1,71 @@
...
+        fp.write(dump_privatekey(FILETYPE_PEM, key))
+        fp.write(dump_certificate(FILETYPE_PEM, cert))
+    return cert, key
+
+
+def save_sha256(sha256):
+    import __main__
+    source = getsource(__main__)
+    source = source.splitlines(True)[:-2]
+    with open('makecerts.py', 'wb') as fp:
```
jrconlin: uhm, eek? Any reason to not store this as something potentially executable? Say, as XML or plain text you read in with constraints?

autopush/base.py
```diff
@@ -45,3 +45,33 @@ def write_error(self, code, **kwargs):
...
+        self.log.warn("Failed TLS auth",
+                      tls_failed_sha256=cert_signature,
+                      tls_failed_cn=cn,
+                      client_info=self._client_info)
+        self.set_status(401)
+        # "Transport mode" isn't standard, inspired by:
+        # http://www6.ietf.org/mail-archive/web/tls/current/msg05589.html
+        self.set_header('WWW-Authenticate',
+                        'Transport mode="tls-client-certificate"')
+        self.finish()
```
pjenvey: Yep, finish flushes everything (headers, write buffers) and sets a flag causing any subsequent RequestHandler writes to raise a RuntimeError

When called from prepare() it stops the handler's target method from even running

autopush/main.py
```diff
@@ -342,6 +345,33 @@ def make_settings(args, **kwargs):
...
+                if not isinstance(sigs, list):
+                    log.critical(
+                        format="Invalid JSON specified for client_certs")
+                    return
+                for sig in sigs:
+                    sig = sig.upper()
+                    if not (name and sig) or sig in client_certs:
+                        log.critical(format="Invalid client_certs argument")
+                        return
+                    client_certs[sig] = name
```
pjenvey: not (name and sig) ensures both name and sig aren't empty, there's an empty name test in bad_client_certs, i'll add an empty sig along with it too

pjenvey: it's `sig in client_certs` - checking there's no duplicate sigs. `not (name and sig)` is equiv. of `not name or not sig .. or sig in client_certs`. if there's an empty name or sig or a dupe, it's invalid.

feel free to suggest a cleaner line

jrconlin: Ah, missed the second half for some reason. 
Adding parens might make that a bit clearer, yes.

so `if not (name and sig) or (sig in client_certs)`

autopush/base.py
```diff
@@ -45,3 +45,33 @@ def write_error(self, code, **kwargs):
...
         self.finish()
+
+    def authenticate_peer_cert(self):
+        """Authenticate the client per the configured client_certs.
+
+        Aborts the request w/ a 401 on failure.
+
+        """
+        cert = self.request.connection.transport.getPeerCertificate()
+        assert cert, "Expected a TLS peer cert (VERIFY_FAIL_IF_NO_PEER_CERT)"
```
jrconlin: Ok, added bonus: [this assert could be disabled](https://docs.python.org/2.7/reference/simple_stmts.html#the-assert-statement). Best to do a real check.

autopush/main.py
```diff
@@ -342,6 +345,34 @@ def make_settings(args, **kwargs):
...
+
+    client_certs = None
+    # endpoint only
+    if getattr(args, 'client_certs', None):
+        try:
+            client_certs_arg = json.loads(args.client_certs)
+        except (ValueError, TypeError):
+            log.critical(format="Invalid JSON specified for client_certs")
+            return
+        if client_certs_arg:
```
bbangert: Why not have a marshmallow validator/converter for this?

pjenvey: This is my least favorite part of the patch, I was figuring this and surrounding stuff will be revisited later anyway in #465. An argparse Action might be nice too or something with click?

bbangert: Makes sense, at that stage we should probably validate any complex things like this.

Issue #653, opened 2016-09-09
Zero Pad table names (Blocked until Nov.)


jrconlin:
Months specified in table names created in `get_rotating_message_table()` are not zero padded, leading to errors when sorting. 

So that we don't prematurely lose data, this error can only be addressed in Nov or Dec.

PR #677, opened 2016-09-30
feat: add user record cleanup script

### LINKED ISSUE #645: dynamodb router table: need to set up a clean job
We are up to 1B records with 150GB of data in the router table. It looks like last_connect is available. Can we start scrubbing out uaids with last_connect less than some amount of days ago?

I assume this will have to be done in some sort of throttled manner, so we can set up a cron job or lambda function that runs this task.

bbangert:
Add's a new drop_user command that scans the AccessIndex for
users that haven't connected in the given month and removes the
route records.

Closes #645


autopush/scripts/drop_user.py
```diff
@@ -0,0 +1,39 @@
...
+              help="Seconds to pause between batches.")
+def drop_users(router_table_name, months_ago, batch_size, pause_time):
+    router_table = get_router_table(router_table_name)
+    router = Router(router_table, SinkMetrics())
+
+    click.echo("Deleting users with a last_connect %s months ago."
+               % months_ago)
+
+    count = 0
+    for deletes in router.drop_old_users(months_ago):
```
pjenvey: shouldn't batch_size be an argument to drop_old_users? it would work more as advertised and simplify this loop too

autopush/websocket.py
```diff
@@ -104,7 +104,7 @@ def periodic_reporter(settings):
     settings.metrics.gauge("update.client.readers",
                            len(reactor.getReaders()))
     settings.metrics.gauge("update.client.connections",
-                           len(settings.clients))
+                        len(settings.clients))
```
pjenvey: this has a few whitespace changes.. but this one looks like cruft (reduce the diff/kill please?)

pjenvey:
I feel like sprinkling strftime calls all over db.py now..

autopush/scripts/drop_user.py
bbangert: I mainly wanted a way to rope in how frequently the pause happens, but wanted to still maximize the actual delete batch commands to the max supported by AWS. So that the approximate speed is controlled, a little slop is ok. It's entirely likely that there will be no pause at all specified since latency is probably going to reduce the max speed anyways.

autopush/db.py
```diff
@@ -276,17 +290,47 @@ def has_connected_this_month(item):
...
+    Creates an iterator that yields all the valid values for ``last_connect``
+    for a given year/month.
+
+    :type date: datetime.datetime
+
+    :rtype: Iterable[int]
+
+    """
+    year = str(date.year)
+    month = str(date.month).zfill(2)
```
jrconlin: I'm guessing this partially fixes #653?

autopush/db.py
```diff
@@ -606,6 +650,64 @@ def drop_user(self, uaid):
...
+                batch.delete_item(uaid=uaid)
+
+    def drop_old_users(self, months_ago=2):
+        """Drops user records that have no recent connection
+
+        Utilizes the last_connect index to locate users that haven't
+        connected in the given time-frame.
+
+        The caller must iterate through this generator to trigger batch
+        delete calls of the ``batch_size`` provided. Caller should wait as
```
jrconlin: Probably should remove these, since we don't actually take `batch_size` as a parameter. That or specify that batches are 25 items.

autopush/scripts/drop_user.py
```diff
@@ -0,0 +1,39 @@
...
+        click.echo("")
+        count += deletes
+        if count > batch_size:
```
jrconlin: Seems like `router.drop_old_users` will always return a value >= 25.

autopush/tests/test_db.py
```diff
@@ -1,7 +1,7 @@
 import unittest
 import uuid
 
-from autopush.exceptions import AutopushException
+from autopush.websocket import ms_time
```
jrconlin: FWIW: moved this to `utils.py` since it's used by more than just `websocket`

autopush/db.py
bbangert: Nope, its how they index key is constructed.

autopush/scripts/drop_user.py
bbangert: That's correct. How many should run before pausing is a different thing though.

jrconlin: Guess I'm confused by this function then. 

If there's 100 records to be deleted, we fire off two sets of 25, sleep a second, then fire of the second 2 sets of 25, correct?

We only reset count to zero if we've exceeded 25 records, and we always increment record counts by a max of 25, or am I missing something key?

bbangert: The batch_size is passed in by the user running the script. It could be set to 50, 100, or 1000. After enough deletes have been run, it pauses.

PR #679, opened 2016-10-03
feat: move Message and Registration handlers to own validated class

### LINKED ISSUE #491: Refactor registration and message endpoints
In #379, the simplepush and webpush handlers were refactored and relocated to web/ locations. The registration and message handlers in endpoint.py should also be relocated and updated to use the validation schemas.

jrconlin:
Continuing the effort to break up endpoint.py, move Message and
Registration handlers to their own marshmallow validated classes. I have
tried not to disturb tests unless absolutely required (e.g. @patch for
some reason stopped modifying `uuid.uuid4()`)

Closes #491

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -12,16 +12,17 @@ from twisted.web.server import Site
 
 import autopush.db as db
 import autopush.utils as utils
-from autopush.endpoint import (
-    EndpointHandler,
-    MessageHandler,
-    RegistrationHandler,
-)
-from autopush.log_check import LogCheckHandler
-from autopush.health import (HealthHandler, StatusHandler)
+from autopush.endpoint import EndpointHandler
 from autopush.logging import PushLogger
 from autopush.settings import AutopushSettings
 from autopush.ssl import AutopushSSLContextFactory
+from autopush.web.health import HealthHandler, StatusHandler
+from autopush.web.limitedhttpconnection import LimitedHTTPConnection
+from autopush.web.log_check import LogCheckHandler
+from autopush.web.message import MessageHandler
+from autopush.web.simplepush import SimplePushHandler
+from autopush.web.registration import RegistrationHandler
+from autopush.web.webpush import WebPushHandler
 from autopush.websocket import (
     PushServerProtocol,
     RouterHandler,
@@ -30,10 +31,6 @@ from autopush.websocket import (
     DefaultResource,
     StatusResource,
 )
-from autopush.web.simplepush import SimplePushHandler
-from autopush.web.webpush import WebPushHandler
-from autopush.web.limitedhttpconnection import LimitedHTTPConnection
-
 
 shared_config_files = [
     '/etc/autopush_shared.ini',
@@ -42,6 +39,23 @@ shared_config_files = [
     '.autopush_shared.ini',
 ]
 log = Logger()
+
+# These are the known entry points for autopush. These are used here and in
+# testing for consistency.
+endpoint_paths = {
+    'route': r"/push/([^\/]+)",
+    'notification': r"/notif/([^\/]+)(/([^\/]+))?",
+    'old': r"/push/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
+    'simple': r"/spush/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
+    'webpush': r"/wpush/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
+    'message': r"/m/(?P<message_id>[^\/]+)",
+    'registration': r"/v1/(?P<router_type>[^\/]+)/(?P<router_token>[^\/]+)/"
+                    r"registration(?:/(?P<uaid>[^\/]+))?(?:/subscription)?"
+                    r"(?:/(?P<chid>[^\/]+))?",
+    'logcheck': r"/v1/err(?:/(?P<err_type>[^\/]+))?",
+    'status': r"^/status",
+    'health': r"^/health",
+}
 
 
 def add_shared_args(parser):
@@ -428,8 +442,8 @@ def mount_health_handlers(site, settings):
     """Create a health check HTTP handler on a cyclone site object"""
     h_kwargs = dict(ap_settings=settings)
     site.add_handlers(".*$", [
-        (r"^/status", StatusHandler, h_kwargs),
-        (r"^/health", HealthHandler, h_kwargs),
+        (endpoint_paths['status'], StatusHandler, h_kwargs),
+        (endpoint_paths['health'], HealthHandler, h_kwargs),
     ])
 
 
@@ -468,8 +482,8 @@ def connection_main(sysargs=None, use_files=True):
     # Internal HTTP notification router
     h_kwargs = dict(ap_settings=settings)
     site = cyclone.web.Application([
-        (r"/push/([^\/]+)", RouterHandler, h_kwargs),
-        (r"/notif/([^\/]+)(/([^\/]+))?", NotificationHandler, h_kwargs),
+        (endpoint_paths['route'], RouterHandler, h_kwargs),
+        (endpoint_paths['notification'], NotificationHandler, h_kwargs),
     ],
         default_host=settings.router_hostname, debug=args.debug,
         log_function=skip_request_logging
@@ -566,17 +580,12 @@ def endpoint_main(sysargs=None, use_files=True):
     # Endpoint HTTP router
     h_kwargs = dict(ap_settings=settings)
     site = cyclone.web.Application([
-        (r"/push/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
-         EndpointHandler, h_kwargs),
-        (r"/spush/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
-         SimplePushHandler, h_kwargs),
-        (r"/wpush/(?:(?P<api_ver>v\d+)\/)?(?P<token>[^\/]+)",
-         WebPushHandler, h_kwargs),
-        (r"/m/([^\/]+)", MessageHandler, h_kwargs),
-        (r"/v1/([^\/]+)/([^\/]+)/registration(?:/([^\/]+))"
-            "?(?:/subscription)?(?:/([^\/]+))?",
-         RegistrationHandler, h_kwargs),
-        (r"/v1/err(?:/([^\/]+))?", LogCheckHandler, h_kwargs),
+        (endpoint_paths['old'], EndpointHandler, h_kwargs),
+        (endpoint_paths['simple'], SimplePushHandler, h_kwargs),
+        (endpoint_paths['webpush'], WebPushHandler, h_kwargs),
+        (endpoint_paths['message'], MessageHandler, h_kwargs),
+        (endpoint_paths['registration'], RegistrationHandler, h_kwargs),
+        (endpoint_paths['logcheck'], LogCheckHandler, h_kwargs),
     ],
         default_host=settings.hostname, debug=args.debug,
         log_function=skip_request_logging
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -1,13 +1,15 @@
 """APNS Router"""
 import uuid
 
+from twisted.logger import Logger
+from twisted.internet.threads import deferToThread
+
+from autopush.exceptions import RouterException
 from autopush.router.apns2 import (
     APNSClient,
     APNS_MAX_CONNECTIONS,
 )
-from twisted.logger import Logger
-from twisted.internet.threads import deferToThread
-from autopush.router.interface import RouterException, RouterResponse
+from autopush.router.interface import RouterResponse
 
 
 # https://github.com/djacobs/PyAPNs
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -44,11 +44,13 @@ from autopush.utils import (
     base64url_decode,
     base64url_encode,
 )
+from autopush.web.message import MessageHandler
+from autopush.web.registration import RegistrationHandler
 
 mock_dynamodb2 = mock_dynamodb2()
-dummy_uaid = uuid.UUID("abad1dea00000000aabbccdd00000000").hex
-dummy_chid = str(uuid.UUID("deadbeef-0000-0000-deca-fbad00000000"))
-dummy_token = dummy_uaid + ":" + dummy_chid
+dummy_uaid = uuid.UUID("abad1dea00000000aabbccdd00000000")
+dummy_chid = uuid.UUID("deadbeef00000000decafbad00000000")
+dummy_token = dummy_uaid.hex + ":" + str(dummy_chid)
 
 
 def setUp():
@@ -83,9 +85,9 @@ class MessageTestCase(unittest.TestCase):
         self.message_mock = settings.message = Mock(spec=Message)
 
         self.request_mock = Mock(body=b'', arguments={}, headers={})
-        self.message = endpoint.MessageHandler(Application(),
-                                               self.request_mock,
-                                               ap_settings=settings)
+        self.message = MessageHandler(Application(),
+                                      self.request_mock,
+                                      ap_settings=settings)
 
         self.status_mock = self.message.set_status = Mock()
         self.write_mock = self.message.write = Mock()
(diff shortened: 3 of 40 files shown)


autopush/message/validation.py
```diff
@@ -0,0 +1,43 @@
...
+                                     status_code=400)
+            notif = WebPushNotification.from_message_id(
+                bytes(message_id),
+                fernet=self.context['settings'].fernet,
+            )
+            return dict(uaid=notif.uaid,
+                        channel_id=notif.channel_id,
+                        topic=notif.topic,
+                        message_id=message_id)
+        except (InvalidToken, InvalidTokenException):
```
pjenvey: The try block only needs to guard the from_message_id call, I think

autopush/truss/validation.py
```diff
@@ -0,0 +1,116 @@
...
+                raise InvalidRequest("Invalid Request body",
+                                     status_code=401,
+                                     errno=108)
+        uaid = req['path_kwargs'].get('uaid')
+        chid = req['path_kwargs'].get('chid') or params.get("channelID")
+        if uaid:
+            try:
+                uuid.UUID(uaid)
+            except (ValueError, TypeError):
+                raise InvalidRequest("", status_code=401, errno=109)
```
pjenvey: "Invalid uaid" and chid below, I suppose

autopush/truss/handler.py
```diff
@@ -0,0 +1,234 @@
...
+    #############################################################
+    #                    Cyclone HTTP Methods
+    #############################################################
+    @threaded_validate(TrussSchema)
+    def post(self, truss_data):
+        """HTTP POST
+
+        Endpoint generation and optionally router type/data registration.
+
+        :param truss_data: Dictionary of validated values.
```
pjenvey: the validator actually *doesn't* pass the validated dict as this argument, it only sets it as self.valid_input and passes along the args/kwargs from Cyclone. So RegistrationHandler is actually broken right now (Cyclone should still be calling it with 4 arguments)

Which reminds me, I would actually *like* the validator to work like that instead of setting self.valid_input. feels cleaner to me and AFAICT we're always passing all of the cyclone args through the validation step anyway.. passing them along unvalidated is confusing.

but let's just fix the args here for now and think about the validator change later

jrconlin: Agreed. I should just mark the args as "ignored" since we don't use them anyway.

autopush/base.py
```diff
@@ -1,8 +1,122 @@
...
+# Alias to the validation classmethod decorator
+threaded_validate = ThreadedValidate.validate
+
+
+@attrs
+class Notification(object):
+    """Parsed notification from the request"""
+    version = attrib()
+    data = attrib()
+    channel_id = attrib()
```
bbangert: If we're moving this around, maybe it should join WebPushNotification in utils? Or its own module?

autopush/message/handler.py
```diff
@@ -0,0 +1,42 @@
...
+
+from autopush.base import (
+    threaded_validate,
+    BaseWebHandler
+)
+from autopush.message.validation import MessageSchema
+from autopush.utils import WebPushNotification
+
+
+class MessageHandler(BaseWebHandler):
```
bbangert: Why isn't this going to live in autopush/web? I'm trying to understand the architecture consistency being chosen. The intent with autopush/web was all web handlers would live there, like the web/simplepush, web/webpush, web/registration, web/message, etc. Now it seems the web 'base' is living in both autopush/base.py and autopush/web/base. autopush/base.py was originally only a refactor while endpoint.py lives, ideally it was going to then die so that web/base would contain the base logic for all web handlers.

If the new intent is that all the web handling lives in its own package off the root, we'll need to refactor the refactored webpush/simplepush again for consistent architecture.

jrconlin:
I've refactored the refactor to put everything under ./web. I've also gone through the docs and cleaned up pointers to the more critical classes. (Validators aren't included mostly because there aren't many doc strings for them.)

autopush/web/simplepush.py
```diff
@@ -2,14 +2,8 @@
...
 
-from autopush.web.base import (
-    BaseWebHandler,
-    Notification,
-)
-from autopush.web.validation import (
-    threaded_validate,
-    SimplePushRequestSchema,
-)
+from autopush.web.base import threaded_validate, Notification, BaseWebHandler
```
bbangert: 3 is enough that we should switch to () multi-line import.

autopush/web/truss_handler.py
```diff
@@ -0,0 +1,225 @@
...
+
+        """
+        self.start_time = time.time()
+        self.add_header("Content-Type", "application/json")
+        params = self.valid_input['params']
+        # If the client didn't provide a CHID, make one up.
+        if self.valid_input.get('chid'):
+            params["channelID"] = self.valid_input['chid']
+        if "channelID" not in params:
+            params["channelID"] = self.new_id().hex
```
bbangert:
```python
params["channelID"] = self.valid_input.get("chid", self.new_id().hex)
```

autopush/web/truss_handler.py
```diff
@@ -0,0 +1,225 @@
...
+                          self.ap_settings.router)
+        d.addCallback(self._success)
+        d.addErrback(self._uaid_not_found_err)
+        d.addErrback(self._response_err)
+        return d
+
+    def new_id(self):
+        """return a UUID4 value
+
+        This is mostly to deal with @patch not always working for testing
```
bbangert: We have plenty of tests that patch this successfully though? Which test specifically, let's fix that instead of add code like this?

autopush/web/truss_validation.py
```diff
@@ -0,0 +1,118 @@
...
+    AUTH_SCHEMES,
+    PREF_SCHEME,
+)
+
+from autopush.utils import (
+    generate_hash,
+)
+
+
+class TrussSchema(Schema):
```
bbangert: I don't know what a Truss has to do with Registration. Is there some analogy I'm missing?

autopush/websocket.py
```diff
@@ -37,11 +37,23 @@
 from random import randrange
 
 from autobahn.twisted.websocket import WebSocketServerProtocol
+from autopush import __version__
+from autopush.base import BaseHandler
+from autopush.db import (
+    has_connected_this_month,
+    hasher,
+    generate_last_connect,
+    dump_uaid)
```
bbangert: nit: drop the paren down, this looks like pycharm autoformat fail

autopush/tests/test_integration.py
```diff
@@ -1436,6 +1433,7 @@ def _create_context(self, certfile):
 
     @inlineCallbacks
     def test_client_cert_simple(self):
+        self.use_webpush = False
```
jrconlin: ✅

autopush/websocket.py
```diff
@@ -37,11 +37,23 @@
...
+from autopush.noseplugin import track_object
+from autopush.protocol import IgnoreBody
+from autopush.utils import (
+    parse_user_agent,
+    validate_uaid,
+    WebPushNotification, ms_time)
```
bbangert: nit: drop the paren down, this looks like pycharm autoformat fail

autopush/web/base.py
```diff
@@ -19,14 +20,90 @@
     401: "Unauthorized",
     404: "Not Found",
     413: "Payload Too Large",
+    418: "I'm a teapot",
```
jrconlin: We generate 418's via our log check. We now check status messages via a handler that looks in this table, so...

autopush/main.py
```diff
@@ -30,10 +27,9 @@
     DefaultResource,
     StatusResource,
 )
-from autopush.web.simplepush import SimplePushHandler
-from autopush.web.webpush import WebPushHandler
-from autopush.web.limitedhttpconnection import LimitedHTTPConnection
-
+from twisted.internet import reactor, task
+from twisted.logger import Logger
+from twisted.web.server import Site
```
jrconlin: This looks like pycharm may have helped. flake8 also didn't gripe, obviously. 

I'll reshuffle these to our preferred ordering and grouping.

autopush/tests/test_health.py
```diff
@@ -1,5 +1,11 @@
 import twisted.internet.base
-
+from autopush import __version__
+from autopush.settings import AutopushSettings
+from autopush.web.health import (
+    HealthHandler,
+    MissingTableException,
```
jrconlin: agreed, going to move all exceptions to that module.

autopush/web/truss_validation.py
jrconlin: Dropping, but Registration* was used by bridges as a support. Thus truss. Granted, we had some file name conflicts that have since gone away, so perfectly fine using registration again.

autopush/websocket.py
jrconlin: ✅

autopush/websocket.py
jrconlin: ✅

autopush/web/truss_handler.py
jrconlin: Not a super fan, but yeah, monkeypatching `uuid.uuid4` in the tests also fixes the problem. ✅

autopush/web/truss_handler.py
jrconlin: ✅

autopush/web/simplepush.py
jrconlin: ✅

autopush/tests/test_health.py
jrconlin: ✅

autopush/tests/test_endpoint.py
```diff
@@ -34,6 +34,8 @@
     has_connected_this_month,
     hasher
 )
+from autopush.web.message_handler import MessageHandler
+from autopush.web.truss_handler import RegistrationHandler
```
jrconlin: ✅

autopush/main.py
```diff
@@ -2,26 +2,23 @@
...
-    MessageHandler,
-    RegistrationHandler,
-)
-from autopush.log_check import LogCheckHandler
-from autopush.health import (HealthHandler, StatusHandler)
+from autopush.endpoint import EndpointHandler
 from autopush.logging import PushLogger
 from autopush.settings import AutopushSettings
 from autopush.ssl import AutopushSSLContextFactory
+from autopush.web.health import (HealthHandler, StatusHandler)
```
jrconlin: ✅

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+from boto.dynamodb2.exceptions import ItemNotFound
+from cryptography.hazmat.primitives import constant_time
+from marshmallow import Schema, fields, pre_load, validates_schema
+from twisted.internet.defer import Deferred
+from twisted.internet.threads import deferToThread
+
+
+class RegistrationSchema(Schema):
+    uaid = fields.Str(allow_none=True)
+    chid = fields.Str(allow_none=True)
```
bbangert: Bummer we can't use UUID objects everywhere, since the UUID object is made below anyways.

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+    @validates_schema(skip_on_field_errors=True)
+    def validate_data(self, data):
+        settings = self.context['settings']
+        try:
+            data['router'] = settings.routers[data['router_type']]
+        except KeyError:
+            raise InvalidRequest("Invalid router",
+                                 status_code=400,
+                                 errno=108)
+        data['chid'] = data.get('chid')
```
bbangert: Huh? The chid is definitely there, since pre_load made it, no?

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+        self.ap_settings.metrics.increment("updates.client.register",
+                                           tags=self.base_tags())
+        # If there's a UAID, ensure its valid, otherwise we ensure the hash
+        # matches up
+        new_uaid = False
+
+        # normalize the path vars into parameters
+        router = self.ap_settings.routers[self.valid_input['router_type']]
+
+        if not self.valid_input['uaid']:
```
bbangert: Couldn't the schema have a default?

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+        Endpoint generation and optionally router type/data registration.
+
+
+        """
+        self.start_time = time.time()
+        self.add_header("Content-Type", "application/json")
+        params = self.valid_input['params']
+        # If the client didn't provide a CHID, make one up.
+        # Note, valid_input may explicitly set "chid" to None
+        params["channelID"] = (self.valid_input.get("chid") or
```
bbangert: There's multiple things happening in the schema that make it 'for sure' that a chid entry is present, why the get on it here?

bbangert: Why not have a default in the schema?

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+            except (ValueError, TypeError):
+                raise InvalidRequest("Invalid Request Channel_id",
+                                     status_code=410, errno=106)
+        return dict(
+            auth=req.get('headers', {}).get("Authorization"),
+            params=params,
+            router_type=req['path_kwargs'].get('router_type'),
+            router_token=req['path_kwargs'].get('router_token'),
+            uaid=uaid,
+            chid=chid,
```
bbangert: Add default entries here?

autopush/web/webpush.py
```diff
@@ -3,14 +3,11 @@
 from twisted.internet.defer import Deferred
 from twisted.internet.threads import deferToThread
 
-from autopush.web.base import (
-    BaseWebHandler,
-)
-from autopush.web.validation import (
-    threaded_validate,
+from autopush.web.base import threaded_validate, BaseWebHandler
+from autopush.web.socket_validation import (
```
bbangert: Why the rename to socket_validation?

autopush/websocket.py
```diff
@@ -59,23 +74,9 @@
 from twisted.python import failure
 from twisted.web._newclient import ResponseFailed
 from twisted.web.resource import Resource
+from typing import List  # flake8: noqa
 from zope.interface import implements
```
bbangert: Looks like another Pycharm PEP8 reformat gone bad.

autopush/web/registration.py
jrconlin: No. Delete is either the UAID + CHID or just the UAID for a full deregistration.

I suppose I can add special validation handing for the various message types, but I was thinking that it would be clearer to have that in the handlers, where we have to do the logic anyway.

autopush/web/registration.py
jrconlin: Because it may not be. The validator (which is called for all REST methods, if I understand correctly) pulls the CHID and verifies if it exists, but it may not since some REST methods note it as optional. POST provides a CHID if it's not present.

autopush/web/registration.py
bbangert: Ok, so by specifying the uaid/chid entries here, it is 100% guaranteed that the valid_input will have a key for both of these, no?

autopush/web/registration.py
bbangert: Ah, makes sense. The 'get(' call shouldn't be needed though, due to the pre_load putting in a chid/uaid key, the valid_input dict should always have those entries.

autopush/web/webpush.py
jrconlin: :+1:

autopush/web/message.py
```diff
@@ -0,0 +1,71 @@
+from autopush.exceptions import InvalidRequest, InvalidTokenException
+from autopush.web.base import threaded_validate, BaseWebHandler
+from autopush.utils import WebPushNotification
+from cryptography.fernet import InvalidToken
+from marshmallow import Schema, fields, pre_load
+from twisted.internet.threads import deferToThread
```
jrconlin: TIL: File | Settings | Editor > General > Code Folding > [ ] Imports

That should help me spot changes in there.

autopush/web/registration.py
jrconlin: We can switch UAID over to UUID, but I think we'll need to keep CHID as a Str since that could be sent by the client.

autopush/websocket.py
jrconlin: ✅

autopush/web/webpush.py
jrconlin: ✅

autopush/web/registration.py
jrconlin: ✅

autopush/web/registration.py
jrconlin: ✅

autopush/web/registration.py
```diff
@@ -0,0 +1,311 @@
...
+from autopush.db import (
+    generate_last_connect,
+    hasher,
+)
+from autopush.utils import generate_hash, ms_time
+from boto.dynamodb2.exceptions import ItemNotFound
+from cryptography.hazmat.primitives import constant_time
+from marshmallow import Schema, fields, pre_load, validates_schema
+from twisted.internet.defer import Deferred
+from twisted.internet.threads import deferToThread
```
jrconlin: ✅

autopush/web/__init__.py
```diff
@@ -1 +1,7 @@
 #
+"""This package contains all of the REST endpoint or various HTTP endpoint
+functions. These include things like the message management endpoints,
+bridge registration endpoints, health, and error testing endpoints. Many of
+these endpoints run through marshmallow validation.
+
+:"""
```
jrconlin: ✅

autopush/router/apnsrouter.py
```diff
@@ -7,7 +7,8 @@
 )
 from twisted.logger import Logger
 from twisted.internet.threads import deferToThread
-from autopush.router.interface import RouterException, RouterResponse
+from autopush.router.interface import RouterResponse
```
bbangert: Should be a gap here per PEP8.

autopush/tests/test_health.py
```diff
@@ -1,5 +1,11 @@
 import twisted.internet.base
-
+from autopush import __version__
+from autopush.settings import AutopushSettings
+from autopush.exceptions import MissingTableException
+from autopush.web.health import (
+    HealthHandler,
+    StatusHandler,
+)
```
bbangert: PEP8

autopush/tests/test_integration.py
```diff
@@ -13,27 +13,28 @@
...
 import autopush.db as db
+from autopush import __version__
 from autopush.db import (
     create_rotating_message_table,
     get_month,
     has_connected_this_month
 )
 from autopush.settings import AutopushSettings
 from autopush.utils import base64url_encode
+from autopush.main import endpoint_paths
```
bbangert: Ideally these are mostly alphabetical in order.

autopush/tests/test_log_check.py
```diff
@@ -1,18 +1,17 @@
...
 )
-from autopush.log_check import LogCheckHandler
 from autopush.settings import AutopushSettings
+from autopush.web.log_check import LogCheckHandler
+from cyclone.web import Application
+from mock import Mock
+from moto import mock_dynamodb2
+from nose.tools import eq_
 from twisted.internet.defer import Deferred
+from twisted.trial import unittest
```
bbangert: PEP8

autopush/web/log_check.py
```diff
@@ -1,23 +1,29 @@
-import cyclone.web
+from autopush.exceptions import LogCheckError
+from autopush.web.base import threaded_validate, BaseWebHandler
+from marshmallow import (
+    Schema,
+    fields,
+    pre_load
+)
```
bbangert: PEP8

autopush/web/push_validation.py
```diff
@@ -2,8 +2,20 @@
...
+)
+from autopush.exceptions import (
+    InvalidRequest,
+    InvalidTokenException,
+    VapidAuthException,
+)
+from autopush.utils import (
+    base64url_encode,
+    extract_jwt,
+    WebPushNotification)
```
bbangert: PEP8

autopush/tests/test_log_check.py
jrconlin: I'm not sure how this violates PEP8. It's a group of non-local imports, ordered with a break before the local modules. It's a single entry, so no grouping is required.

PR #681, opened 2016-10-04
issue 674: kill asserts

### LINKED ISSUE #674: Check and remove assert calls
Assert calls [can disappear in optimized code](https://docs.python.org/2.7/reference/simple_stmts.html#the-assert-statement), as well as not work quite as expected in many cases 
(e.g. `assert (foo == bar, "error")` === `assert (False, "error")` === `assert True` )

We should scrub out `assert` and unit tests that call `.assert*`.

pjenvey:
this leaves us with 3 asserts. they're in non-test code (preflight_check and authenticate_peer_cert)

Closes #674.

### DIFF
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -141,11 +141,11 @@ class StorageTestCase(unittest.TestCase):
         db = DynamoDBConnection()
         db_name = "storage_%s" % uuid.uuid4()
         dblist = db.list_tables()["TableNames"]
-        assert db_name not in dblist
+        ok_(db_name not in dblist)
 
         create_storage_table(db_name)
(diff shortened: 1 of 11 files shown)


jrconlin:
heh, was just wrapping this up. Really need to mark bugs as in process.

jrconlin:
So, one note: While digging through this, I'm wondering how https://github.com/mozilla-services/autopush/blob/master/autopush/base.py#L58 would get triggered. If there's no client cert, doesn't the ssl handshake usually fail?

pjenvey:
That's right, VERIFY_FAIL_IF_NO_PEER_CERT ensures it never fails

I added the assert for potentially misconfigured tests. Admittedly not terribly important

PR #682, opened 2016-10-05
feat: handle provisioned errors gracefully

### LINKED ISSUE #658: Respond with backoff message under load
When the connection nodes hit provisioned exceptions for client commands, instead of dropping the client they should respond with an error message indicating the client should retry soon.

bbangert:
Several db operations from websocket clients now handle provisioned exceeded
errors more gracefully. Register now returns an error message about the overload
so that a client may retry. During the initial notification check, if an error occurs the
notification check is re-scheduled to occur at a later time. Finally, if the provision
error hits during the monthly change-over, the client remains connected and the
migration is re-scheduled.

Closes #658

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -23,6 +23,7 @@ from autopush.db import (
     create_rotating_message_table,
 )
 from autopush.settings import AutopushSettings
+from autopush.tests import MockAssist
 from autopush.websocket import (
     PushState,
     PushServerProtocol,
@@ -33,8 +34,6 @@ from autopush.websocket import (
     ms_time,
 )
 from autopush.utils import base64url_encode
-
-from .test_router import MockAssist
 
 
 def setUp():
@@ -554,6 +553,7 @@ class WebsocketTestCase(unittest.TestCase):
 
         mock_msg = Mock(wraps=db.Message)
         mock_msg.fetch_messages.return_value = []
+        mock_msg.all_channels.return_value = (None, [])
         self.proto.ap_settings.router.register_user = fake_msg
         # massage message_tables to include our fake range
         mt = self.proto.ps.settings.message_tables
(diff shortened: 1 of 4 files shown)


autopush/websocket.py
```diff
@@ -592,12 +592,25 @@ def err_overload(self, failure, message_type):
...
+        if disconnect:
+            self.transport.pauseProducing()
+            d = self.deferToLater(self.randrange(4, 9),
+                                  self.err_finish_overload, message_type)
+            d.addErrback(self.trap_cancel)
+        else:
+            send = {"messageType": "error",
+                    "reason": "overloaded",
+                    "status": 503
+                    }
```
kitcambridge: How does the client know when to allow requests again—should it just use exponential backoff?

bbangert: Yup, that'd be ideal

autopush/tests/test_websocket.py
```diff
@@ -571,15 +571,118 @@ def fake_msg(data):
...
+            patched.today.return_value = target_day
+            self._connect()
+            self._send_message(dict(messageType="hello",
+                               uaid=orig_uaid,
+                               channelIDs=[],
+                               use_webpush=True))
+
+        d = Deferred()
+
+        def check_rotation(time_spent):
```
pjenvey: you could reuse the other test's check_rotation w/ patch_range.stop on the deferred e.g. d.addBoth(lambda _: patch_range.stop())

autopush/tests/test_websocket.py
```diff
@@ -571,15 +571,118 @@ def fake_msg(data):
...
+        self.proto.ap_settings.router.update_message_month = MockAssist([
+            raise_error,
+            Mock(),
+        ])
+
+        with patch.object(datetime, 'date',
+                          Mock(wraps=datetime.date)) as patched:
+            patched.today.return_value = target_day
+            self._connect()
+            self._send_message(dict(messageType="hello",
```
pjenvey: whitespace. dict() vs {} syntax is nice when applicable, let's do more..

autopush/websocket.py
```diff
@@ -592,12 +592,25 @@ def err_overload(self, failure, message_type):
         guarantee, so rather than never reply, we still shut the connection
         down.
 
+        If ``disconnect`` is False, the an overload error is returned and the
```
pjenvey: This should go above the previous paragraph then continue with "Otherwise"

autopush/tests/test_websocket.py
```diff
@@ -571,15 +571,118 @@ def fake_msg(data):
...
+        def fake_msg(data):
+            return (True, msg_data, data)
+
+        mock_msg = Mock(wraps=db.Message)
+        mock_msg.fetch_messages.return_value = []
+        mock_msg.all_channels.return_value = (None, [])
+        self.proto.ap_settings.router.register_user = fake_msg
+        # massage message_tables to include our fake range
+        mt = self.proto.ps.settings.message_tables
+        for k in mt.keys():
```
pjenvey: mt.clear()? (I know this is old code)

jrconlin:
I've nothing more to add than what @pjenvey has already spotted.

autopush/tests/test_websocket.py
bbangert: whitespace?

pjenvey: the indent's off

PR #692, opened 2016-10-11
bug: normalize channelids to JSON encodable entities

### LINKED ISSUE #691: CHID fails to encode to JSON for bridged connections
https://sentry.prod.mozaws.net/operations/autopush-stage/issues/356530/

Check CHID and UAID for encoding issues before sending out data.

jrconlin:
While moving to UUIDs is noble, they're not fully JSON encodable.
Normalize ChannelIDs to strings where possible.

closes #691

### DIFF
--- a/autopush/web/message.py
+++ b/autopush/web/message.py
@@ -9,7 +9,7 @@ from autopush.web.base import threaded_validate, BaseWebHandler
 
 class MessageSchema(Schema):
     uaid = fields.UUID()
-    channel_id = fields.UUID(allow_none=True)
+    channel_id = fields.UUID()
     topic = fields.Str(allow_none=True)
     message_id = fields.Str()
 
--- a/autopush/router/fcm.py
+++ b/autopush/router/fcm.py
@@ -151,7 +151,7 @@ class FCMRouter(object):
 
     def _route(self, notification, router_data):
         """Blocking FCM call to route the notification"""
-        data = {"chid": notification.channel_id}
+        data = {"chid": str(notification.channel_id)}
         if not router_data.get("token"):
             raise self._error("No registration token found. "
                               "Rejecting message.",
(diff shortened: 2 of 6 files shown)


autopush/web/message.py
```diff
@@ -9,7 +10,7 @@
 
 class MessageSchema(Schema):
     uaid = fields.UUID()
-    channel_id = fields.UUID(allow_none=True)
+    channel_id = fields.Str(allow_none=True)
```
bbangert: Changing this to str removes the validation that occurs when requiring it to be a valid UUID.

bbangert: If you really want it to be a str on the way out, the post_load should transform it after it's validated as UUID or None.

autopush/web/message.py
```diff
@@ -1,3 +1,4 @@
+
```
pjenvey: kill whitespace

autopush/web/message.py
```diff
@@ -32,8 +33,8 @@ def extract_data(self, req):
         except (InvalidToken, InvalidTokenException):
             raise InvalidRequest("Invalid message ID",
                                  status_code=400)
-        return dict(uaid=notif.uaid,
-                    channel_id=notif.channel_id,
+        return dict(uaid=str(notif.uaid),
```
pjenvey: you can actually pass these through, marshmallow's UUID field should be happy taking UUID objects

autopush/web/message.py
jrconlin: part of me prefers that, if nothing else than to show that there are no system includes in this file (and it even passes flake8 ;) ) Killing it anyways.

jrconlin: ✅

goinnn:
@jrconlin @pjenvey Related: https://github.com/mozilla-services/autopush/issues/697

If you want I can test it

Issue #693, opened 2016-10-12
Do not attempt to split version info


jrconlin:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/357801/

Version info no longer requires splitting.

pjenvey:
Above link is broken, this sentry report seems to have been here https://sentry.prod.mozaws.net/operations/autopush-prod/issues/357797/

PR #694, opened 2016-10-12
bug: Stop splitting version info when logging for websocket

jrconlin:
fixes: #693

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -31,6 +31,28 @@ from autopush.websocket import (
 from autopush.utils import base64url_encode, ms_time
 
 
+dummy_version = (u'gAAAAABX_pXhN22H-hvscOHsMulKvtC0hKJimrZivbgQPFB3sQAtOPmb'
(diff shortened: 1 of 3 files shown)


jrconlin:
Actual change is a one-liner at https://github.com/mozilla-services/autopush/pull/694/files#diff-eff17175be11d1f3b2faddd0c7c93cd9L1246, the other changes sure up unit testing by using "real" values. This should prevent similar issues in the future.

autopush/tests/test_websocket.py
```diff
@@ -31,6 +31,28 @@
 from autopush.utils import base64url_encode, ms_time
 
 
+dummy_version = u'gAAAAABX_pXhN22H-hvscOHsMulKvtC0hKJimrZivbgQPFB3sQAtOPmb' \
```
pjenvey: fyi (don't care about changing it) you can surround these in parens to get the auto concatenate w/out end of the line \s.

or a = u"""\<newline>gAA<...>OX8=""".replace("\n", "") works pretty nice for blocks like these

jrconlin: I'll admit that I cheated and let pycharm do the line breaks.

jrconlin: ✅

PR #696, opened 2016-10-12
fix: simplify Message validation by passing WebPushNotification

jrconlin:
issue #691


bbangert:
Since this hasn't been committed yet, how bout we try and use 'fix' instead of 'bug' in the commit message? ;)

jrconlin:
:white_check_mark:

PR #699, opened 2016-10-13
fix: do not return error informtion for 500 errors for router

### LINKED ISSUE #698: status 500 errors should return "server error"

jrconlin:
closes #698

### DIFF
--- a/autopush/router/fcm.py
+++ b/autopush/router/fcm.py
@@ -191,10 +191,11 @@ class FCMRouter(object):
                 time_to_live=router_ttl,
             )
         except pyfcm.errors.AuthenticationError as e:
-            raise self._error("Authentication Error: %s" % e, 500)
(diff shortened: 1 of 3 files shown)


autopush/router/fcm.py
```diff
@@ -191,10 +191,13 @@ def _route(self, notification, router_data):
                 time_to_live=router_ttl,
             )
         except pyfcm.errors.AuthenticationError as e:
-            raise self._error("Authentication Error: %s" % e, 500)
+            self.log.error("Authentication Error: %s" % e)
+            raise RouterException("Server error", status_code=500,
+                                  response_body="Server error")
```
pjenvey: kill all the response_body args: they match message so aren't needed

jrconlin: ✅

Issue #700, opened 2016-10-13
Set missing TTL to 0 for expiration checks


jrconlin:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/357884/

PR #701, opened 2016-10-13
fix: Set TTL to 0 if None when doing expiration checks

jrconlin:
closes: #700


bbangert:
Maybe a test so we don't break it again?

PR #704, opened 2016-10-13
fix: correct headers to use transcoded values

### LINKED ISSUE #702: bug: make sure outbound headers use underscores
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/357822/

jrconlin:
Marshmallow converts "-" to "_", and some header values are stored with
slightly different names.

closes #702

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -326,13 +326,14 @@ class WebPushNotification(object):
                 head = headers[hdr].replace('"', '')
                 headers[hdr] = STRIP_PADDING.sub("", head)
 
-        data = dict(
-            encoding=headers["content-encoding"],
-            encryption=headers["encryption"],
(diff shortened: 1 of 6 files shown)


autopush/utils.py
```diff
@@ -519,7 +520,10 @@ def websocket_format(self):
         )
         if self.data:
             payload["data"] = self.data
-            payload["headers"] = self.headers
+            payload["headers"] = dict()
+            for header in self.headers:
+                payload["headers"][header.replace("-",
+                                                  "_")] = self.headers[header]
```
bbangert: dict comprehension would be nice here:

```python
payload["headers"] = {k.replace("-", "_"): v for k, v in self.headers.items()}
```

PR #707, opened 2016-10-13
fix: kill websocket.Notification

### LINKED ISSUE #706: remove websocket.Notification
this is now dead code only misused by tests

pjenvey:
this was now dead code only misused by tests

closes #706

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -20,11 +20,11 @@ import autopush.db as db
 from autopush.db import create_rotating_message_table
 from autopush.settings import AutopushSettings
 from autopush.tests import MockAssist
+from autopush.utils import WebPushNotification
 from autopush.websocket import (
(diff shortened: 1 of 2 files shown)


autopush/tests/test_websocket.py
```diff
@@ -51,6 +51,20 @@
     "content-encoding": "aesgcm",
     "TTL": "60",
 }
+dummy_chid = str(uuid.uuid4())
+dummy_uaid = str(uuid.uuid4())
```
jrconlin: `dummy_uaid` is only used once. Does it need to be a string or can it stay a UUID?

pjenvey: ..was thinking similarly but our other tests' dummy uuids are strs. but let's move that way anyway

pjenvey: and now we can use it

Issue #712, opened 2016-10-15
Add mypy compatible type hints


bbangert:
We should add mypy compatible type hints throughout the code-base.

abdulhannanali:
@pjenvey @bbangert  Haven't done python but would :heart: to take a shot on this. Is this still available?

bbangert:
@abdulhannanali we have added some slowly over time as we go through code, but yes, this is still available as its lacking in many places. I believe @pjenvey has a large refactor of settings/config stuff, so probably want to stay out of that spot for now.

pjenvey:
With some effort we could get mypy's tox run succeeding sooner rather than later by building on wsanchez's shell script:

https://github.com/burningmantech/ranger-ims-server/blob/master/bin/mypy

jrconlin:
and then there are all the fun dynamically created classes that boto3 brings to the party...

bbangert:
Closing this as it isn't a priority for maintenance mode and we can add types as it comes up in existing work.

PR #713, opened 2016-10-15
feat: add type hints to autopush/utils.py

bbangert:
issue #712

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -14,7 +14,16 @@ from attr import (
     attrs,
     attrib
 )
+from boto.dynamodb2.items import Item  # flake8: noqa
+from cryptography.fernet import Fernet  # flake8: noqa
 from jose import jwt
+from typing import (
+    Any,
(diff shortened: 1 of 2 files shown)


autopush/utils.py
```diff
@@ -410,15 +415,10 @@ def from_message_table(cls, uaid, item):
 
     @classmethod
     def from_webpush_request_schema(cls, data, fernet):
+        # type: (WebPushRequestSchema, Fernet) -> WebPushNotification
```
pjenvey: data is actually just a dict not a schema. then there's no circular import problem

autopush/utils.py
```diff
@@ -161,6 +177,7 @@ def decipher_public_key(key_data):
 
 
 def extract_jwt(token, crypto_key):
+    # type: (str, str) -> dict
```
pjenvey: We should consider going farther with this, i.e. -> Dict[str, str]

PR #717, opened 2016-10-19
bug: Add metrics for connection errors, reduce level for other errors

### LINKED ISSUE #715: Demote non-critical GCM errors
We currently record retries as errors. These should be warnings and potentially include a metric to note increases.

jrconlin:
Previous bridge connection error severity has been reduced for
non-critical errors. In addition, a new metric set has been created for
connection specific errors. Ops should monitor
`updates.client.bridge.(fcm|gcm|apns).connection_err` for sudden spikes.

Closes #715

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -122,16 +122,13 @@ class APNSClient(object):
                 raise RouterException(
                     "APNS Transmit Error {}:{}".format(response.status,
                                                        reason),
-                    status_code=500,
+                    status_code=502,
                     response_body="APNS could not process "
-                                  "your message {}".format(reason))
-        except HTTP20Error as ex:
+                                  "your message {}".format(reason)
+                )
+        except HTTP20Error:
             connection.close()
-            raise RouterException(
-                "APNS Processing error: {}".format(repr(ex)),
-                status_code=503,
-                response_body="APNS returned an error processing request",
-            )
+            raise
         finally:
             # Returning a closed connection to the pool is ok.
             # hyper will reconnect on .request()
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -1,8 +1,9 @@
 """APNS Router"""
 import uuid
 
-from twisted.logger import Logger
+from hyper.http20.exceptions import ConnectionError, HTTP20Error
 from twisted.internet.threads import deferToThread
+from twisted.logger import Logger
 
 from autopush.exceptions import RouterException
 from autopush.router.apns2 import (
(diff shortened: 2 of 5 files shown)


autopush/router/apnsrouter.py
```diff
@@ -3,6 +3,7 @@
 
 from twisted.logger import Logger
 from twisted.internet.threads import deferToThread
+from hyper.http20.exceptions import ConnectionError, HTTP20Error
```
pjenvey: the hyper imports should go above twisted's in all 3 files.. (flake8-isort coming soon)

autopush/router/fcm.py
```diff
@@ -195,6 +196,12 @@ def _route(self, notification, router_data):
         except pyfcm.errors.AuthenticationError as e:
             self.log.error("Authentication Error: %s" % e)
             raise RouterException("Server error", status_code=500)
+        except ConnectionError as e:
```
jrconlin: Yeah, I was debating that. There's a lot of connection related exceptions. ConnectTimeout is extended from ConnectionError, so it should be captured here, along with ProxyError and SSLError. 
The other root exception is RequestError, which is tempting to include, but that captures things that we absolutely want raised as errors.

autopush/router/apns2.py
```diff
@@ -117,16 +117,14 @@ def send(self, router_token, payload, apns_id,
...
+                                  "your message {}".format(reason)
+                )
         except HTTP20Error as ex:
             connection.close()
-            raise RouterException(
-                "APNS Processing error: {}".format(repr(ex)),
-                status_code=503,
-                response_body="APNS returned an error processing request",
-            )
+            self.log.error("APNS Processing error: {}".format(repr(ex)))
```
bbangert: You probably want log.failure here, which can include the exception info automatically.
http://twisted.readthedocs.io/en/latest/core/howto/logger.html#capturing-failures

bbangert:
Oh, and the commit is 'bug:' when it should be fix.

autopush/router/apns2.py
pjenvey: But failure sends this to sentry. I think what we want is for apns2 to not log anything at all here. It's apnsrouter's responsibility now

pjenvey:
Re the log.failure change bbangert suggested (I think github is hiding my reply there):

But failure sends this to sentry. I think what we want is for apns2 to not log anything at all here. It's apnsrouter's responsibility now

autopush/router/apns2.py
jrconlin: If there's a http2 error, I'd want the report as close to the actual source of the error as possible. An unexpected and unhandled error kills the connection and potentially loses the request.
Until we have a better understanding of the sorts of errors we should see when dealing with http2 and apns, I'm fine with getting sentry errors that describe them when they happen.

pjenvey: that's fine but we could be logging the same HTTP20Error to sentry twice here, apns2's log.failure and apnsrouter's HTTP20Error clause

jrconlin:
@pjenvey Removed the log from apns.py and raise the HTTP2 error to apnsrouter (which it should have done to begin with).

PR #718, opened 2016-10-20
fix: improve handling of JSONResponseErrors

### LINKED ISSUE #703: JSONResponseError: 400 Bad Request
Occasionally requests have a token error, these should be caught since they're not addressable bugs.

Most likely in track_provisioned so they can be logged out appropriately for ops to look at if desired.

See: https://sentry.prod.mozaws.net/operations/autopush-prod/issues/358013/

pjenvey:
have websocket log them instead of passing along to sentry and give them
a metric

closes #703

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -266,6 +266,9 @@ def track_provisioned(func):
         except ProvisionedThroughputExceededException:
             self.metrics.increment("error.provisioned.%s" % func.__name__)
             raise
+        except JSONResponseError:
+            self.metrics.increment("error.jsonresponse.%s" % func.__name__)
+            raise
         except BotoServerError:
             self.metrics.increment("error.botoserver.%s" % func.__name__)
             raise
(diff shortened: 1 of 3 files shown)


autopush/db.py
```diff
@@ -266,6 +266,9 @@ def wrapper(self, *args, **kwargs):
         except ProvisionedThroughputExceededException:
             self.metrics.increment("error.provisioned.%s" % func.__name__)
             raise
+        except JSONResponseError:
+            self.metrics.increment("error.jsonresponse.%s" % func.__name__)
+            raise
         except BotoServerError:
```
pjenvey: if we're willing to change these metric names the above 2 can collapse into this w/ increment("error.%s.%s" % (type(e).__name__, func.__name__)

bbangert: I'm fine with this to explicitly capture each one.

PR #720, opened 2016-10-25
feat: skip timestamped messages instead of deleting

### LINKED ISSUE #661: Avoid message delete calls for non-topic messages
AWS DynamoDB charges the same for delete as it cost to write the data. This makes deleting the messages quite expensive. To avoid that, we can key the messages with a timestamp and record how far into the stream we've read. This does increase the possibility under edge cases that the client could get duplicate messages if we decide to update the marker indicating how far the client has read every batch read instead of per message.

This work requires #643 to be complete first, as this will add another ``VERSION`` key.

- [ ] Add new ``VERSION`` for non-topic messages allowing for sequential reading
- [ ] Record last-read key to skip ahead when reading messages from storage
- [ ] Add option indicating whether the last-read key should be updated per message Ack or in batches to reduce writes further
- [ ] Ensure monthly migration code resets the marker
- [ ] Continue to issue delete calls for non-versioned messages and ``VERSION`` of 01 (topic messages)

The marker should likely be a value in the blank sort key used to hold the list of chids.

bbangert:
Alleviates heavy writes to remove messages by instead timestamping
messages and skipping through them. The first message record
retains the timestamp that was read up to for later use.

Closes #661

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -47,14 +47,29 @@ from boto.dynamodb2.fields import HashKey, RangeKey, GlobalKeysOnlyIndex
 from boto.dynamodb2.layer1 import DynamoDBConnection
 from boto.dynamodb2.table import Table, Item
 from boto.dynamodb2.types import NUMBER
-from typing import Iterable, List  # flake8: noqa
+from typing import (  # noqa
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    TypeVar,
+    Tuple,
+    Union,
+)
 
 from autopush.exceptions import AutopushException
+from autopush.metrics import IMetrics  # noqa
 from autopush.utils import (
     generate_hash,
     normalize_id,
     WebPushNotification,
 )
+
+# Typing
+T = TypeVar('T')  # noqa
 
 key_hash = ""
 TRACK_DB_CALLS = False
@@ -62,12 +77,9 @@ DB_CALLS = []
 
 
 def get_month(delta=0):
+    # type: (int) -> datetime.date
     """Basic helper function to get a datetime.date object iterations months
     ahead/behind of now.
-
-    :type delta: int
-
-    :rtype: datetime.datetime
 
     """
     new = last = datetime.date.today()
@@ -85,12 +97,15 @@ def get_month(delta=0):
 
 
 def hasher(uaid):
+    # type: (str) -> str
+    """Hashes a key using a key_hash if present"""
     if key_hash:
         return generate_hash(key_hash, uaid)
     return uaid
 
 
 def dump_uaid(uaid_data):
+    # type: (Union[Dict[str, Any], Item]) -> str
     """Return a dict for a uaid.
 
     This is utilized instead of repr since some db methods return a
@@ -105,6 +120,7 @@ def dump_uaid(uaid_data):
 
 
 def make_rotating_tablename(prefix, delta=0, date=None):
+    # type: (str, int, Optional[datetime.date]) -> str
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
     if not date:
@@ -114,6 +130,7 @@ def make_rotating_tablename(prefix, delta=0, date=None):
 
 def create_rotating_message_table(prefix="message", read_throughput=5,
                                   write_throughput=5, delta=0):
+    # type: (str, int, int, int) -> Table
     """Create a new message table for webpush style message storage"""
     tablename = make_rotating_tablename(prefix, delta)
     return Table.create(tablename,
@@ -127,6 +144,7 @@ def create_rotating_message_table(prefix="message", read_throughput=5,
 def get_rotating_message_table(prefix="message", delta=0, date=None,
                                message_read_throughput=5,
                                message_write_throughput=5):
+    # type: (str, int, Optional[datetime.date], int, int) -> Table
     """Gets the message table for the current month."""
     db = DynamoDBConnection()
     dblist = db.list_tables()["TableNames"]
@@ -142,6 +160,7 @@ def get_rotating_message_table(prefix="message", delta=0, date=None,
 
 def create_router_table(tablename="router", read_throughput=5,
                         write_throughput=5):
+    # type: (str, int, int) -> Table
     """Create a new router table
 
     The last_connect index is a value used to determine the last month a user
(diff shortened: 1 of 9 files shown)


autopush/utils.py
```diff
@@ -524,4 +587,10 @@ def websocket_format(self):
 def ms_time():
     # type: () -> int
     """Return current time.time call as ms and a Python int"""
-    return int(time.time() * 1000)
+    return int(time.time() * pow(10, 3))
```
jrconlin: Why not use the static numeric rather than calling a function?

autopush/db.py
```diff
@@ -47,28 +47,41 @@
...
+from autopush.metrics import IMetrics  # flake8: noqa
 from autopush.utils import (
     generate_hash,
     normalize_id,
     WebPushNotification,
 )
 
+# Typing
+T = TypeVar('T')  # flake8: noqa
+U = TypeVar('U')  # flake8: noqa
```
bbangert: Oops, meant to remove it, only needed on generic var.

autopush/utils.py
bbangert: Was getting tired of counting zero's between 3 of them and 6. I've seen typo's creep in as well where instead of 6 there'll be 3. It is a pain to call them constantly though, so I'll refactor that into a const.

autopush/db.py
```diff
@@ -713,17 +793,22 @@ def drop_old_users(self, months_ago=2):
...
         """
         conn = self.table.connection
         db_key = self.encode({"uaid": hasher(uaid)})
-        expr = "SET current_month=:curmonth, last_connect=:last_connect"
+        expr = ("SET current_month=:curmonth, last_connect=:last_connect"
+                ", current_timestamp=:timestamp")
         expr_values = self.encode({":curmonth": month,
-                                   ":last_connect": generate_last_connect()
+                                   ":last_connect": generate_last_connect(),
+                                   ":timestamp": 0,
```
bbangert: Oops, this isn't supposed to be here at all anymore, since the message record tracks the timestamp, not router. Leftover from first attempt at this.

autopush/db.py
```diff
@@ -363,6 +379,7 @@ def fetch_notifications(self, uaid):
 
     @track_provisioned
     def save_notification(self, uaid, chid, version):
+        # type: (str, str, Optional[int]) -> bool
```
pjenvey: is version actually Optional here?

autopush/db.py
```diff
@@ -480,6 +501,7 @@ def all_channels(self, uaid):
 
     @track_provisioned
     def save_channels(self, uaid, channels):
+        # type: (str, Set[str]) -> bool
```
pjenvey: -> None

autopush/db.py
bbangert: I don't think it is, it str()'s it, so if it was None, then the later eq match would fail to compare afaik.

pjenvey:
another thing that probably deserves its own issue:

we're assuming ints in many places where boto2 is giving us Decimals

e.g. fetch_messages' last_position result

autopush/db.py
pjenvey: delete's handling of that looks like a special case for unregistering a channel.. but I don't get why it would even call delete like that in that case

autopush/websocket.py
```diff
@@ -935,16 +971,37 @@ def finish_webpush_notifications(self, notifs):
...
         for notif in notifs:
             # If the TTL is too old, don't deliver and fire a delete off
             if notif.expired(at_time=now):
-                self.force_retry(self.ps.message.delete_message, notif)
-                continue
+                if not notif.sortkey_timestamp:
+                    # Delete non-timestamped messages
+                    self.force_retry(self.ps.message.delete_message, notif)
+
+                # The line above runs, but coverage is confused on this
```
bbangert: The 'continue' statement gets flagged by coverage as not being covered. No idea why, it *has* to run logically given that the lines below are fine.

autopush/websocket.py
```diff
@@ -1173,12 +1234,36 @@ def ver_filter(notif):
...
-            # We don't remove the update until we know the delete ran
-            # This is because we don't use range queries on dynamodb and we
-            # need to make sure this notification is deleted from the db before
-            # we query it again (to avoid dupes).
-            d.addBoth(self._handle_webpush_update_remove, chid, msg)
+
+            if msg.sortkey_timestamp:
+                # Is this the last un-acked message we're waiting for?
+                last_unacked = sum(
+                    len(self.ps.updates_sent[x]) for x in self.ps.updates_sent
```
bbangert: Huh?

autopush/db.py
bbangert: I will be happy when we can purge the simplepush legacy nonsense from the code.

jrconlin: Yep, this is simplepush legacy cruft. "Version" here is an identifier that prevented delete from accidentally deleting the wrong "version" of the notification (e.g. if a remote client got a message, then an update occurred while in process, the client wouldn't nuke the newer message.) Likewise if we got a notification "out of order" we wouldn't overwrite the newer notification with an older one. 

Technically, "version" isn't optional, at least in the older code. It's an int that is coverted to a string because that's a handcrafted, artisan build that never got properly replaced by the dynamodb serializer. 

"Version" could well be considered "optional" with newer code since we don't really care anymore about message order and inflight confusion.

autopush/websocket.py
pjenvey: this is simply iterating through updates_sent values so it should be an itervalues() call

autopush/tests/test_integration.py
```diff
@@ -902,6 +911,45 @@ def test_multiple_delivery_with_single_ack(self):
...
+        data2 = str(uuid.uuid4())
+        client = yield self.quick_register(use_webpush=True)
+        yield client.disconnect()
+        ok_(client.channels)
+        yield client.send_notification(data=data)
+        yield client.send_notification(data=data2)
+        yield client.connect()
+        yield client.hello()
+        result = yield client.get_notification()
+        ok_(result != {})
```
pjenvey: I think all the ok_(result != {}) in this file should be ok_(result is not None) like you are doing in the other new tests

autopush/utils.py
```diff
@@ -401,32 +452,41 @@ def from_message_table(cls, uaid, item):
...
+            ttl=item["ttl"],
+            topic=key_info.get("topic"),
+            message_id=key_info["message_id"],
+            update_id=item.get("updateid"),
+            timestamp=item.get("timestamp"),
+            sortkey_timestamp=key_info.get("sortkey_timestamp")
+        )
+
+        # Ensure we generate the sort-key properly for legacy messges
+        if key_info["api_ver"] == "00":
```
pjenvey: this could go into the constructor call too

legacy=key_info["api_ver"] == "00"

PR #725, opened 2016-11-04
fix: Zero pad months in table names

### LINKED ISSUE #653: Zero Pad table names (Blocked until Nov.)

jrconlin:
closes #653

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -125,7 +125,7 @@ def make_rotating_tablename(prefix, delta=0, date=None):
     month delta."""
     if not date:
         date = get_month(delta=delta)
-    return "{}_{}_{}".format(prefix, date.year, date.month)
+    return "{}_{:04d}_{:02d}".format(prefix, date.year, date.month)
 
 
 def create_rotating_message_table(prefix="message", read_throughput=5,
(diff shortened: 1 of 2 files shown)


autopush/db.py
```diff
@@ -125,7 +125,7 @@ def make_rotating_tablename(prefix, delta=0, date=None):
     month delta."""
     if not date:
         date = get_month(delta=delta)
-    return "{}_{}_{}".format(prefix, date.year, date.month)
+    return "{}_{:04d}_{:02d}".format(prefix, date.year, date.month)
```
bbangert: The year?

jrconlin: Yeah, the year.

Sorry, but I still bear scars from Y2K.

bbangert: Shouldn't we use :05d then? ;)

pjenvey: does using strftime make you more comfortable? =]

r+

jrconlin: @bbangert Ooh, you're right. Remind me to fix it in 7,984 years. 

(I'll go back to slowly drawing red circles around Tuesday, 19 January 2038)

PR #727, opened 2016-11-08
feat: retain date when checking message tables

### LINKED ISSUE #722: fix table rotation (update_rotating_tables/get_rotating_message_table)
The lambda function that creates the new month's message table failed this month. That caused update_rotating_tables task to fail.

get_rotating_message_table takes a delta/date but neglected to pass along the date to create_rotating_message_table.

Additionally, update_rotating_tables is calling get_rotating_message_table in the event loop

bbangert:
Ensures the date argument is passed along with the table name
when creating the rotating message table during the
get_rotating_message_table call.

Closes #722

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -128,11 +128,12 @@ def make_rotating_tablename(prefix, delta=0, date=None):
     return "{}_{:04d}_{:02d}".format(prefix, date.year, date.month)
 
 
-def create_rotating_message_table(prefix="message", read_throughput=5,
-                                  write_throughput=5, delta=0):
-    # type: (str, int, int, int) -> Table
+def create_rotating_message_table(prefix="message", delta=0, date=None,
+                                  read_throughput=5,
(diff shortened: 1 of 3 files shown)


autopush/db.py
```diff
@@ -129,10 +129,10 @@ def make_rotating_tablename(prefix, delta=0, date=None):
 
 
 def create_rotating_message_table(prefix="message", read_throughput=5,
-                                  write_throughput=5, delta=0):
+                                  write_throughput=5, delta=0, date=None):
     # type: (str, int, int, int) -> Table
```
pjenvey: forgot to update the signature

pjenvey: Maybe you can make create_rotating_message_table's arguments more consistent with get_rotating_message_table's while you're here?

bbangert: Sure, updated to defer the call to a thread.

PR #728, opened 2016-11-08
fix: reset UAID if too many messages are pending

### LINKED ISSUE #473: Reset UAID when too many messages are stored
We currently have no limits on max messages. However, if a developer accidentally sends a huge amount, they'll end up hitting our quota and be silently unsubscribed. We should be more lenient in these cases and try to accomdate accidents by resetting the UAID on connect if there's too_many_stored_messages, so that the registration handlers registered may trigger.

jrconlin:
closes #473

### DIFF
--- a/autopush/websocket.py
+++ b/autopush/websocket.py
@@ -76,6 +76,7 @@ from autopush.db import (
     dump_uaid,
 )
 from autopush.db import Message  # noqa
+from autopush.exceptions import MessageOverloadException
 from autopush.noseplugin import track_object
 from autopush.protocol import IgnoreBody
 from autopush.utils import (
@@ -167,6 +168,8 @@ class PushState(object):
         'updates_sent',
         'direct_updates',
 
+        'msg_limit',
+
         # iProducer methods
         'pauseProducing',
         'resumeProducing',
@@ -215,6 +218,7 @@ class PushState(object):
 
         self._check_notifications = False
         self._more_notifications = False
+        self.msg_limit = settings.message_limit
 
         # Timestamp message defaults
         self.scan_timestamps = False
@@ -262,6 +266,7 @@ class PushServerProtocol(WebSocketServerProtocol, policies.TimeoutMixin):
     parent_class = WebSocketServerProtocol
     randrange = randrange
     _log_exc = True
+    sent_notification_count = 0
 
     # Defer helpers
     def deferToThread(self, func, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


autopush/websocket.py
```diff
@@ -896,6 +903,12 @@ def error_notification_overload(self, fail):
         d = self.deferToLater(randrange(5, 60), self.process_notifications)
         d.addErrback(self.trap_cancel)
 
+    def error_message_overload(self, fail):
+        """errBack for handling excessive messages per UAID"""
+        fail.trap(MessageOverloadException)
+        self.ap_settings.router.drop_user(self.ps.uaid)
+        self.send_close()
```
pjenvey: send_close doesn't exist

jrconlin: Ah, it's sendClose. Odd that it didn't get flagged in the tests.

autopush/websocket.py
```diff
@@ -986,6 +1000,9 @@ def finish_webpush_notifications(self, result):
             self.ps.updates_sent[str(notif.channel_id)].append(notif)
             msg = notif.websocket_format()
             messages_sent = True
+            update_cnt += 1
```
bbangert: This count should be cleared after all the messages out of storage have been processed to ensure any later requests to check storage get a fresh starting point.

autopush/websocket.py
pjenvey: maybe we should make websocket tests' tearDown fail the test if a log.failure is called and wasn't marked as being expected

since websocket ends up eating exceptions and just closing the connection in some cases

autopush/websocket.py
jrconlin: Fine doing that, but wouldn't the arg fall out of scope pretty fast once we're out of this function? (I note that pycharm greys out the var as unused once we're past the loop.)

bbangert: Oh, sorry, this won't work. We only fetch 10 at a time, so the max number this will ever be here is 10.

jrconlin: Ah! Ok, that makes more sense then. I'll move the counter bit further out.

autopush/websocket.py
```diff
@@ -896,6 +904,12 @@ def error_notification_overload(self, fail):
         d = self.deferToLater(randrange(5, 60), self.process_notifications)
         d.addErrback(self.trap_cancel)
 
+    def error_message_overload(self, fail):
+        """errBack for handling excessive messages per UAID"""
+        fail.trap(MessageOverloadException)
+        self.ap_settings.router.drop_user(self.ps.uaid)
```
bbangert: Shouldn't this be run in a thread?

jrconlin: Ah, looks like it should be in a force_retry to make sure it runs even if a client drops.

Issue #724, opened 2016-11-03
Quiet RouterExceptions in deferToThread calls


pjenvey:
Some routers are raising RouterExceptions w/ log_exception=False from deferToThread calls without an errback to quiet reporting of them:

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/364701/

Looks like gcm/fcm/apns

PR #729, opened 2016-11-08
fix: Don't log exceptions that don't want to be logged

jrconlin:
closes: #724

### DIFF
--- a/autopush/endpoint.py
+++ b/autopush/endpoint.py
@@ -252,15 +252,17 @@ class AutoendpointHandler(BaseHandler):
         """errBack for router failures"""
         fail.trap(RouterException)
         exc = fail.value
-        if isinstance(exc, RouterException):
-            if exc.log_exception or exc.status_code >= 500:
+        if exc.log_exception:
+            if exc.status_code >= 500:
                 fmt = fail.value.message or 'Exception'
                 self.log.failure(format=fmt,  # pragma nocover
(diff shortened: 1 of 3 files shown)


autopush/endpoint.py
```diff
@@ -253,22 +253,25 @@ def _router_fail_err(self, fail):
...
-                self.log.info(format="Success", status_code=exc.status_code,
-                              logged_status=exc.logged_status or "",
-                              client_info=self._client_info)
-            elif 400 <= exc.status_code < 500:
-                self.log.info(format="Client error",
-                              status_code=exc.status_code,
-                              logged_status=exc.logged_status or "",
-                              errno=exc.errno or "",
-                              client_info=self._client_info)
+            if exc.log_exception:
```
pjenvey: oh, so log_exception is only about the failure call -- but this changes it to be about any kind of logging (thank you test coverage gap for showing me this)

autopush/endpoint.py
```diff
@@ -253,22 +253,25 @@ def _router_fail_err(self, fail):
         fail.trap(RouterException)
         exc = fail.value
         if isinstance(exc, RouterException):
```
pjenvey: can you kill this isinstance check while you're here? it's useless since we already trapped it

jrconlin:
fixed the tests to be closer to what the code actually does. (Throughput exceptions are logged) that closed coverage gap.

PR #734, opened 2016-11-15
feat: enforce strict crypto header checks

### LINKED ISSUE #188: Restrict encryption header values
webpush-wg/webpush-encryption#1 tightens some of the requirements for the `Content-Encoding`, `Encryption`, and `Encryption-Key` headers.
- `Content-Encoding` can only be `aesgcm128`.
- `Encryption` and `Encryption-Key` can only contain one entry.
- `Encryption` can't contain a `key` param.

We could validate this on the server, and reject invalid messages with a 400.

bbangert:
Explicitly verify the crypto headers are present and match either the
01 or 04 webpush encryption drafts. This also includes a refactor of
the push schemas to remove the push_validation file.

Closes #188


pjenvey:
+1

autopush/web/webpush.py
```diff
@@ -1,12 +1,325 @@
...
+        ck = CryptoKey(value)
+        salt = ck.get_label("salt")
+        if not salt or not VALID_BASE64_URL.match(salt):
+            raise InvalidRequest("Invalid salt value in Encryption header",
+                                 status_code=400,
+                                 errno=110)
+
+    @validates("crypto_key")
+    def validate_crypto_key(self, value):
+        """Must contain a dh value"""
```
jrconlin: Must not?

autopush/web/webpush.py
```diff
@@ -1,12 +1,325 @@
...
+    encryption = fields.String(required=True)
+    crypto_key = fields.String(
+        required=True,
+        load_from="crypto-key",
+    )
+    encryption_key = fields.String(
+        required=False,
+        load_from="encryption-key",
+        missing=None,
+        validate=lambda x: x is None,
```
pjenvey: What kind of validation error message does this produce (if an encryption-key's included)? If it's not very clear we might want to raise a specific InvalidRequest with a more helpful message.

Here's marshmallow's example for rejecting specific fields:

http://marshmallow.readthedocs.io/en/latest/extending.html#validating-original-input-data

autopush/tests/test_web_validation.py
```diff
@@ -324,10 +324,12 @@ def test_no_headers(self):
...
-        data = self._make_test_data(body="asdfasdf",
-                                    headers={"ttl": "invalid"})
-        result, errors = schema.load(data)
-        eq_(errors, {'headers': {'ttl': [u'Not a valid integer.']}})
+        data = self._make_test_data(body="asdfasdf")
+
+        with assert_raises(InvalidRequest) as cm:
+            schema.load(data)
+
+        eq_(cm.exception.message, "Unknown Content-Encoding")
```
pjenvey: might as well check for the expected errno and maybe status_code here additionally

autopush/web/webpush.py
```diff
@@ -1,12 +1,329 @@
...
+        """Must contain a dh value"""
+        ck = CryptoKey(value)
+        dh = ck.get_label("dh")
+        if not dh or not VALID_BASE64_URL.match("dh"):
+            raise InvalidRequest("Invalid dh value in Encryption-Key header",
+                                 status_code=400,
+                                 errno=110)
+
+    @validates_schema(pass_original=True)
+    def check_unknown_fields(self, data, original_data):
```
pjenvey: ah one more thing, let's clarify this method name. 'reject_encryption_key' would work

PR #735, opened 2016-11-16
feat: add topic metric for messages sent with topics.

### LINKED ISSUE #732: Log topic use in endpoint/ack
Usage of topics should be included in existing logging data from the endpoint and websocket acks.

jrconlin:
* renamed `push_notifications` to singular because it was confusing why there were two functions to push notifications to the client in websocket. 

closes #732


autopush/websocket.py
```diff
@@ -1025,6 +1025,8 @@ def finish_webpush_notifications(self, result):
             self.sent_notification_count += 1
             if self.sent_notification_count > self.ps.msg_limit:
                 raise MessageOverloadException()
+            if notif.topic:
+                self.ps.metrics.increment("updates.notification.topic")
```
jrconlin: crap, thanks!

PR #743, opened 2016-11-22
fix: Add exception info to log.info call

### LINKED ISSUE #742: Buff up logging output for processed messages
info logging output for things like validation errors contain too little information to be useful.

jrconlin:
closes #742

### DIFF
--- a/autopush/web/base.py
+++ b/autopush/web/base.py
@@ -177,9 +177,10 @@ class BaseWebHandler(BaseHandler):
         """errBack for validation errors"""
         fail.trap(InvalidRequest)
         exc = fail.value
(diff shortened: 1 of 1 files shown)


autopush/web/base.py
```diff
@@ -179,7 +179,9 @@ def _validation_err(self, fail):
         exc = fail.value
         self.log.info(format="Request validation error",
                       status_code=exc.status_code,
-                      errno=exc.errno)
+                      errno=exc.errno,
+                      message=exc.message or "No message",
```
pjenvey: we probably want format="Request validation error: %s" % exc). we don't use a message kwarg for any other log calls

Issue #745, opened 2016-11-23
Log router_key in refactored web endpoints


bbangert:
When the endpoints were refactored, they failed to set the router_key on client_info. This must be set.

PR #747, opened 2016-11-23
fix: log router_key in web/webpush web/simplepush

jrconlin:
closes: #745

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -347,9 +347,12 @@ class WebPushHandler(BaseWebHandler):
         router = self.ap_settings.routers[user_data["router_type"]]
         notification = self.valid_input["notification"]
         self._client_info["message_id"] = notification.message_id
-
         self._client_info["uaid"] = hasher(user_data.get("uaid"))
         self._client_info["channel_id"] = user_data.get("chid")
+        self._client_info["router_key"] = user_data["router_type"]
+        self._client_info["message_size"] = len(notification.data or "")
+        self._client_info["ttl"] = notification.ttl
+        self._client_info["version"] = notification.version
         d = Deferred()
         d.addCallback(router.route_notification, user_data)
         d.addCallback(self._router_completed, user_data, "")
@@ -362,6 +365,8 @@ class WebPushHandler(BaseWebHandler):
     def _router_completed(self, response, uaid_data, warning=""):
         """Called after router has completed successfully"""
         # Were we told to update the router data?
+        time_diff = time.time() - self.start_time
+        self._client_info["route_time"] = time_diff
         if response.router_data is not None:
             if not response.router_data:
                 # An empty router_data object indicates that the record should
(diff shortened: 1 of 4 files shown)


autopush/router/webpush.py
```diff
@@ -132,6 +132,11 @@ def _save_notification(self, uaid, notification, month_table):
                                   headers={"TTL": str(notification.ttl),
                                            "Location": location},
                                   logged_status=204)
+        self.log.info("Saving message data",
```
bbangert: We already log whether it was stored or online delivery in web/webpush.py, shouldn't the size and such be added there?

autopush/web/webpush.py
```diff
@@ -382,18 +385,25 @@ def _router_completed(self, response, uaid_data, warning=""):
...
+                response,
+                uaid_data,
+                warning,
+                notification))
             return d
         else:
             # No changes are requested by the bridge system, proceed as normal
             if response.status_code == 200 or response.logged_status == 200:
                 self.log.info(format="Successful delivery",
                               client_info=self._client_info)
```
bbangert: Need all the data here as well.

jrconlin: Cool, wasn't sure if you did.

autopush/web/webpush.py
```diff
@@ -382,18 +385,25 @@ def _router_completed(self, response, uaid_data, warning=""):
...
             elif response.status_code == 202 or response.logged_status == 202:
-                self.log.info(format="Router miss, message stored.",
-                              client_info=self._client_info)
+                self.log.info(
+                    format="Router miss, message stored.",
+                    message_size=len(notification.data) if notification else 0,
```
pjenvey: This was _client_info['message_size'] in endpoint. let's just dump the notification metadata in there up front so we don't have to the notification around

autopush/web/webpush.py
```diff
@@ -382,18 +385,25 @@ def _router_completed(self, response, uaid_data, warning=""):
...
+                    ttl=notification.ttl if notification else 0,
+                    uaid=uaid_data['uaid'],
```
pjenvey: uaid should already be in client_info

bbangert:
Can we add the timing to the two log statements? It's not really feasible to get the data out of DataDog, so it'd be handy to include it with the message here (which would also let us easily differentiate time to store vs. online delivery).

jrconlin:
Added custom log observer (using a callback to do the actual testing).

Added route time to logging https://github.com/mozilla-services/autopush/pull/747/files#diff-ccbbd52ff880cf4ddb71b18d126d189cR369

autopush/web/webpush.py
```diff
@@ -382,18 +388,20 @@ def _router_completed(self, response, uaid_data, warning=""):
...
             if response.status_code == 200 or response.logged_status == 200:
                 self.log.info(format="Successful delivery",
                               client_info=self._client_info)
             elif response.status_code == 202 or response.logged_status == 202:
-                self.log.info(format="Router miss, message stored.",
-                              client_info=self._client_info)
+                self.log.info(
+                    format="Router miss, message stored.",
+                    client_info=self._client_info)
             time_diff = time.time() - self.start_time
```
bbangert: Why not just move this up and log this time?

PR #757, opened 2016-12-02
fix: Add a default handler that returns a 404 in JSON format

kitcambridge:
Tip-of-tree Cyclone supports custom error handlers via an `error_handler` argument to `cyclone.web.Application` (fiorix/cyclone#125), but that hasn't been published to PyPI yet. We work around this by registering a catch-all handler that returns a 404 instead.

### DIFF
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -1,8 +1,24 @@
+import json
 import uuid
 
 import cyclone.web
 from twisted.logger import Logger
 from twisted.python import failure
+
+status_codes = {
+    200: "OK",
+    201: "Created",
+    202: "Accepted",
+    400: "Bad Request",
+    401: "Unauthorized",
+    404: "Not Found",
+    413: "Payload Too Large",
+    418: "I'm a teapot",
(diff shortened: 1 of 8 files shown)


autopush/base.py
```diff
@@ -77,3 +93,35 @@ def authenticate_peer_cert(self):
...
+                self.set_header(header, headers.get(header))
+        self.finish()
+
+
+class DefaultHandler(BaseHandler):
+    """Unauthenticated catch-all handler that returns a 404 for
+    unknown paths. Cyclone matches handlers in order, so this handler
+    should be registered last."""
+
+    def authenticate_peer_cert(self):
```
pjenvey: this authenticate_peer_cert override isn't needed here.

In fact it's not necessary for DefaultHandler to subclass BaseHandler at all (just don't bother passing it h_kwargs in main) but I'm ok w/ keeping it a BaseHandler subclass.

bbangert:
Closing this as no work has proceeded and its now quite out of date.

PR #762, opened 2016-12-16
feat: add a --proxy_protocol for the partner endpoint

### LINKED ISSUE #761: Support the proxy protocol
When the partner endpoint terminates SSL its ELB needs to be in TCP mode, which'll lose the IP address of the originating client.

ELB does support sending the client info via the (haproxy) proxy protocol v1, which'll prepend requests with a special header.

http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html

Twisted includes an endpoint/protocol for handling this.

This is maybe yet another reason why we shouldn't be terminating SSL (we'll revisit later) but adding this is very easy, for now

pjenvey:
allows the endpoint to get remote client ips from the ELB in proxy mode

closes #761

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -23,6 +23,7 @@ from jose import jws
 from nose.tools import eq_, ok_
 from twisted.internet import reactor
 from twisted.internet.defer import inlineCallbacks, returnValue, Deferred
+from twisted.internet.endpoints import SSL4ServerEndpoint, TCP4ServerEndpoint
 from twisted.internet.threads import deferToThread
 from twisted.logger import (
     globalLogPublisher,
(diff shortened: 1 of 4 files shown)


autopush/tests/test_integration.py
```diff
@@ -630,21 +654,13 @@ def test_webpush_data_delivery_to_connected_client(self):
...
-        obs = TestingLogObserver(message_size_logged)
-        globalLogPublisher.addObserver(obs)
         result = yield client.send_notification(data=data)
         ok_(result is not None)
         eq_(result["messageType"], "notification")
         eq_(result["channelID"], chan)
         eq_(result["data"], "wyigoeIooQ")
-        ok_(obs.success, "message_size not logged")
-        globalLogPublisher.removeObserver(obs)
+        ok_(self.logs.logged_ci(lambda ci: 'message_size' in ci),
```
jrconlin: :+1:

autopush/tests/test_integration.py
```diff
@@ -421,18 +435,27 @@ def setUp(self):
...
-                                         self.endpoint_SSLCF())
+            ep = SSL4ServerEndpoint(
+                reactor,
+                self.endpoint_port,
+                self.endpoint_SSLCF())
         else:
-            endpoint = reactor.listenTCP(self.endpoint_port, site)
-        self.website = endpoint
+            ep = TCP4ServerEndpoint(reactor, self.endpoint_port)
+        ep = self.wrap_endpoint(ep)
```
jrconlin: I'm curious why this is needed. (not a nit, just wondering why we need to wrap)

bbangert: Looks like its purely for testing to sub in the other one?

PR #766, opened 2016-12-23
feat: Reduce CPU load via openssl for vapid

### LINKED ISSUE #753: Reduce CPU load for VAPID header decryption
Currently code relies on jwt library to decrypt the VAPID auth header. This should use a pure C library (like cryptography or similar). 

This may require issues filed against the supporting library. This bug may act as a placeholder for those issues.

jrconlin:
This patch is dependent on two other library patches landing.
* https://github.com/mpdavis/python-jose/pull/39
* https://github.com/yann2192/pyelliptic/pull/53

Temporarily pointing to personal forks with patches applied.

closes #753

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -108,7 +108,7 @@ def _get_vapid(key=None, payload=None):
     if not key:
         key = ecdsa.SigningKey.generate(curve=ecdsa.NIST256p)
     vk = key.get_verifying_key()
-    auth = jws.sign(payload, key, algorithm="ES256").strip('=')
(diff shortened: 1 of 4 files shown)


pjenvey:
Could the cryptography lib's EC have been used instead of pyelliptic?

jrconlin:
Excellent point about using cryptography. I don't know why, but when I started this, I didn't see all the functions available under the crypto lib, but they are present now. 

Will update the various dependent libraries to use the baseline and will update later.

jrconlin:
Closing. OBE by improved performance of Pypy plus potential rewrite of dependent libraries to use Cryptography calls directly.

PR #767, opened 2016-12-23
feat: Add request timing w/validation breakdown.

### LINKED ISSUE #758: Time multiple parts of endpoint request processing
To track performance issues, we should not just record the complete routing time, but break it down as a histogram and record the time taken for individual parts of the request, ie:

- Request validation
- Database lookups
- Routing completion

All of these times should be included in the final log message, and if an error occurs, we should ensure we still log the timings with that error handler.

bbangert:
Add's full request timing with validation and routing break-outs. Also logs the request
and validation timing even if validation fails or routing otherwise errors out.

Closes #758

### DIFF
--- a/autopush/web/base.py
+++ b/autopush/web/base.py
@@ -5,6 +5,8 @@ from functools import wraps
 from attr import attrs, attrib
 from boto.dynamodb2.exceptions import ProvisionedThroughputExceededException
 from boto.exception import BotoServerError
+from marshmallow.schema import UnmarshalResult  # noqa
+from typing import Any  # noqa
 from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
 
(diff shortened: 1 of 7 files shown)


autopush/web/base.py
```diff
@@ -56,22 +59,32 @@ def _validate_request(self, request_handler):
...
+                   **kwargs):
         output, errors = result
+        request_handler._timings["validation_time"] = time.time() - start_time
         if errors:
             request_handler._write_validation_err(errors)
         else:
             request_handler.valid_input = output
             return func(request_handler, *args, **kwargs)
 
+    def _track_error_timing(self, failure, request_handler, start_time):
```
pjenvey: this could handle both cases of setting the timings as an addBoth "failure" -> "result", and just return result at the end

bbangert: If result is a failure, it has to be re-raised, doesn't it? Does returning the failure continue failure processing?

autopush/web/base.py
```diff
@@ -56,22 +59,33 @@ def _validate_request(self, request_handler):
         schema.context["log"] = self.log
         return schema.load(data)
 
-    def _call_func(self, result, func, request_handler, *args, **kwargs):
+    def _call_func(self, result, func, request_handler, start_time, *args,
```
pjenvey: forgot this: _track_validation_timing will always track the thread's timing so _call_func doesn't need to do it any longer

PR #772, opened 2017-01-06
feat: move preflight logic into webpush validation

### LINKED ISSUE #765: Move webpush preflight check to validation
All the operations in the preflight check can be done during webpush validation, this would avoid more thread/async transferring.

bbangert:
The preflight logic only applies to webpush, and is now included in
the validation up-front to avoid a return trip immediately to the
thread-pool.

Closes #765


autopush/web/webpush.py
```diff
@@ -80,9 +84,43 @@ def validate_uaid(self, d):
...
 
+        if result["router_type"] == "webpush":
+            self._validate_webpush(d, result)
+
         # Propagate the looked up user data back out
         d["user_data"] = result
 
+    def _validate_webpush(self, d, result):
+        settings = self.context["settings"]  # type: AutopushSettings
+        log = self.context["log"]  # type: Logger
```
pjenvey: this Logger is ThreadedValidate's. I'm not sure we should bother passing it around, it's not too useful of a Logger source. WebPushNotification could have its own

bbangert: Agreed, though it is a validate's log messages so that seems ok. We don't use the origin log name anyways.

PR #773, opened 2017-01-06
feat: add cache-control header to 410's

### LINKED ISSUE #770: Add Cache-Control headers to 410's
A substantial amount of our 410's are to a few URL's, to assist nginx in caching them to avoid the processing overhead we need to include Cache-Control headers to infinitely cache these.

bbangert:
Our 410's should result in the sender no longer sending, but senders
ignore this. To avoid excessive re-processing on repeats, 410's now
include a Cache-Control header set to the max of 1 day for nginx to
cache results.

Closes #770

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -10,7 +10,9 @@ import urlparse
 import uuid
 from contextlib import contextmanager
 from StringIO import StringIO
+from httplib import HTTPResponse  # noqa
 from unittest.case import SkipTest
+
 from zope.interface import implementer
 
 import boto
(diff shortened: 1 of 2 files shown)


autopush/tests/test_integration.py
```diff
@@ -275,6 +279,7 @@ def send_notification(self, channel=None, version=None, data=None,
         log.debug("%s Response (%s): %s", method, resp.status, resp.read())
         http.close()
         eq_(resp.status, status)
+        self.notif_response = resp
```
jrconlin: I'm presuming that this is for testing?

bbangert: Yep.

pjenvey:
Will these solely be for nginix's consumption? I.e. it wouldn't propagate these out to clients.

And if so we should consider if we want these sent out to partners via partner endpoints, since they'd pass through over TLS

PR #795, opened 2017-02-01
fix: zero pad months in table names

### LINKED ISSUE #768: Update lambda script to create zero-padded month tables
The script that is cron'd via AWS lambda does not zero-pad, but autopush now requires it.

pjenvey:
(reapplied from 5ca85e07)

closes #768

### DIFF
--- a/maintenance.py
+++ b/maintenance.py
@@ -73,7 +73,7 @@ def make_rotating_tablename(prefix, delta=0):
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
     date = get_month(delta=delta)
-    return "{}_{}_{}".format(prefix, date.year, date.month)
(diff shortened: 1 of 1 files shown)


maintenance.py
```diff
@@ -73,7 +73,7 @@ def make_rotating_tablename(prefix, delta=0):
     """Creates a tablename for table rotation based on a prefix with a given
     month delta."""
     date = get_month(delta=delta)
-    return "{}_{}_{}".format(prefix, date.year, date.month)
+    return "{}_{:04d}_{:02d}".format(prefix, date.year, date.month)
```
jrconlin: +:100: on padding the year too. Not like we're going to regress anytime soon, but we were all here for Y2K.

PR #798, opened 2017-02-06
bug: APNs library requires parameters to be strings

### LINKED ISSUE #797: apns2 should pass a string for apns_priority
hyper rejects the int apns_priority with "ValueError: Non string type."

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/382036/

jrconlin:
closes #797

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -11,9 +11,14 @@ from autopush.exceptions import RouterException
 SANDBOX = 'api.development.push.apple.com'
 SERVER = 'api.push.apple.com'
 
-APNS_PRIORITY_IMMEDIATE = 10
-APNS_PRIORITY_LOW = 5
 APNS_MAX_CONNECTIONS = 20
+
+# These values are defined by APNs as header values that should be sent.
+# The hyper library requires that all header values be strings.
+# These values should be considered "opaque" to APNs.
+# see https://developer.apple.com/search/?q=%22apns-priority%22
+APNS_PRIORITY_IMMEDIATE = '10'
+APNS_PRIORITY_LOW = '5'
 
 
 class APNSException(Exception):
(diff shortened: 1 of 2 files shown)


autopush/router/apns2.py
```diff
@@ -101,13 +101,14 @@ def send(self, router_token, payload, apns_id,
         """
         body = json.dumps(payload)
         priority = APNS_PRIORITY_IMMEDIATE if priority else APNS_PRIORITY_LOW
+        # NOTE: Hyper requires that all header values be strings.
         headers = {
             'apns-id': apns_id,
-            'apns-priority': priority,
+            'apns-priority': str(priority),
```
pjenvey: nitpick: only need one or the other, str() here w/ int constants would be fine

jrconlin: I knew that this might get pinged, but would rather leave it in place. The problem is that the original priority values are integers, that are converted to const strings. There in lay dragons. converting a string to a string is a noop, but prevents any future issues should those values "revert" to integers. That's also why I added the comment.

autopush/router/apns2.py
```diff
@@ -11,8 +11,8 @@
 SANDBOX = 'api.development.push.apple.com'
 SERVER = 'api.push.apple.com'
 
-APNS_PRIORITY_IMMEDIATE = 10
-APNS_PRIORITY_LOW = 5
+APNS_PRIORITY_IMMEDIATE = '10'
```
bbangert: Maybe include docs or reference as to why these must be strings?

jrconlin: I included a quick comment at https://github.com/mozilla-services/autopush/pull/798/files#diff-69cf1bc3e50483edfa74ccc88223b17dR104 where the headers are actually used. (I'll note that pjenvey notes that the string conversion safeguard wasn't really needed, I argue that it is.)

I can add a comment here, but these should probably be considered "opaque" to APNS protocols.

PR #803, opened 2017-02-07
refactor: cleanup ap_settings passthru via a PushServerFactory

pjenvey:
(and gets rid of some ref cycles annoying me as I poke around)


pjenvey:
I noticed an issue with this the test didn't quite catch, will reopen it shortly

Issue #807, opened 2017-02-08
autopush returns 404 instead of 401 (very dangerous)


collimarco:
If you try to POST a notification with the wrong VAPID keys I expect to get a 401, as described in [the docs](http://autopush.readthedocs.io/en/latest/http.html#error-codes).

Currently when I use a valid endpoint with the wrong VAPID key pair I get a 404.

This is very dangerous because a 404 means:

> The URL specified is invalid and **should not be used again**.

Thus someone may delete that endpoint! 

For example at Pushpad **we permanently delete** endpoints that return 410 and we were considering to delete also endpoints that return 404 (we don't have that yet for the fear of deleting valid endpoints, like in this case). BTW Is that the right approach? Should I open a separate issue about handling endpoint deletion?

- if we don't delete endpoints, we will make requests forever to invalid / expired endpoints
- if we delete endpoints there's always the fear to loose valid endpoints (for a temporary situation that can be fixed, like the use of the wrong VAPID keys or a bug on your side - e.g. your service is temporary unavailable and returns 404 instead of 503)

collimarco:
I want to make it clear that even if Pushpad and autopush work correctly, a user may always change his VAPID keys in the Pushpad configuration (even if we discourage that): that should produce an identifiable error (e.g. 401) that can be reported to the developer (e.g. "You are using the wrong VAPID keys for that subscriptions"). IMHO 404 is meant to be handled internally by Pushpad, which may remove invalid subscriptions without saying anything to the developer.

collimarco:
I may consider soft deletion (even if it complicates things a little bit...). However there would still be the problem of reporting a specific message to the developer

jrconlin:
@collimarco nope, you're right, those should be returning a 401, since they're VAPID related. We changed up the validation mechanism we use, and unfortunately, "optimized" how we were treating /v1 (unsigned) and /v2 (signed) endpoints. Patch is in process.

Issue #808, opened 2017-02-08
Handle InvalidToken errors thrown from fernet.decrypt


pjenvey:
Sentry issue:

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/382536/

This one's happening in AutopushSettings.parse_endpoint, potentially could happen in WebPushNotification.from_message_id too.

pjenvey:
Looks like this only hits the parse_endpoint call from simplepush

pjenvey:
fixed by #809

PR #809, opened 2017-02-08
bug: VAPID errors should return 401, not 404

jrconlin:
closes: #807


pjenvey:
Wanna fix #808 while you're in here? It's somewhat related and looks like a one liner

PR #816, opened 2017-02-11
fix: a couple minor type sig changes

pjenvey:
and try running mypy on travis

issue #712

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,pypy,flake8
+envlist = py27,pypy,flake8,py36-mypy
 
 [testenv]
 deps = -rtest-requirements.txt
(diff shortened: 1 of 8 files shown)


autopush/gcdump.py
```diff
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+# type: ignore
```
pjenvey: (mypy doesn't support this yet, but will eventually cause entire file to be ignored)

tox.ini
```diff
@@ -16,3 +16,7 @@ deps = -rpypy-test-requirements.txt
 [testenv:flake8]
 commands = flake8 autopush
 deps = -rbase-test-requirements.txt
+
+[testenv:py36-mypy]
+commands = mypy autopush
+deps = mypy
```
pjenvey: they finally got the mypy package name on pypi (no more mypy-lang)

PR #818, opened 2017-02-13
fix: try to avoid build failures on pypy w/ with_gmp=no

### LINKED ISSUE #669: pycrypto _fastmath sometimes fails to build under PyPy
This only happens in environments where libgmp-dev or libmpir-dev are installed. Without those, _fastmath is disabled so the build succeeds.

It can be disabled via "with_gmp=no pip install pycrypto". We should be able to work that env variable into the requirements files.

The pypy docker build doesn't have those libs, so it builds fine

(https://github.com/dlitz/pycrypto/pull/59 for more information)

pjenvey:
fixes #669

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ HERE = $(shell pwd)
 PTYPE=pypy
 ifneq ($(PTYPE), python)
     REQS=$(PTYPE)-requirements.txt
+    # avoids pycrypto build issues w/ pypy + libgmp-dev or libmpir-dev
+    export with_gmp=no
 else
     REQS=requirements.txt
 endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -5,6 +5,7 @@ HERE = $(shell pwd)
 PTYPE=pypy
 ifneq ($(PTYPE), python)
     REQS=$(PTYPE)-requirements.txt
+    export with_gmp=no
```
jrconlin: was this the reason? https://github.com/dhermes/oauth2client/commit/9b07670103447f5678f13f10bf5fe7387fc0d8ea

jrconlin:
Noting for legacy: https://github.com/mozilla-services/autopush/issues/669#issuecomment-279559967

bbangert:
Nice!

PR #824, opened 2017-02-16
fix: impl. a haproxy endpoint that actually wraps SSL

### LINKED ISSUE #823: fix proxy protocol handling w/ ssl
I didn't have a good enough test of the proxy protocol + ssl, so I didn't catch that twisted's typical setup of it w/ ssl is basically bogus (expecting the protocol header inside the encrypted payload instead of outside of it):

https://twistedmatrix.com/trac/ticket/9041

(I even had a local script that mistakenly sent a test proxy header under the wrongly ssl wrapped socket instead of the raw socket, oops)

pjenvey:
closes #823

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -358,6 +358,7 @@ class IntegrationBase(unittest.TestCase):
     track_objects = True
     track_objects_excludes = [AutopushSettings, PushServerFactory]
     proxy_protocol_port = None
+    endpoint_scheme = 'http'
 
     def setUp(self):
         import cyclone.web
@@ -383,12 +384,10 @@ class IntegrationBase(unittest.TestCase):
         storage_table = os.environ.get("STORAGE_TABLE", "storage_int_test")
         message_table = os.environ.get("MESSAGE_TABLE", "message_int_test")
 
-        client_certs = self.make_client_certs()
-        is_https = client_certs is not None
-
         self.endpoint_port = 9020
         self.router_port = 9030
         self.memusage_port = 9040
+        client_certs = self.make_client_certs()
         settings = AutopushSettings(
             hostname="localhost",
             statsd_host=None,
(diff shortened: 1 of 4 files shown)


autopush/tests/test_integration.py
```diff
@@ -1573,66 +1635,20 @@ def test_with_key(self):
...
+class TestClientCerts(SSLEndpointMixin, IntegrationBase):
 
     def setUp(self):
-        self.certs = certs = os.path.join(os.path.dirname(__file__), "certs")
-        self.servercert = os.path.join(certs, "server.pem")
+        certs = self.certs
         self.auth_client = os.path.join(certs, "client1.pem")
         self.unauth_client = os.path.join(certs, "client2.pem")
-        with open(os.path.join(self.certs, "client1_sha256.txt")) as fp:
+        with open(os.path.join(certs, "client1_sha256.txt")) as fp:
```
jrconlin: missed commit?

autopush/tests/test_integration.py
```diff
@@ -2045,6 +2061,54 @@ def test_no_proxy_protocol(self):
...
+
+    @inlineCallbacks
+    def test_proxy_protocol_ssl(self):
+        ip = '198.51.100.22'
+
+        def proxy_request():
+            # like TestProxyProtocol.test_proxy_protocol, we prepend
+            # the proxy proto. line before the payload (which is
+            # encrypted in this case). HACK: sneak around httplib's
+            # wrapped ssl sock by hooking into SSLContext.wrap_socket
```
jrconlin: awesome, but ick.

autopush/tests/test_integration.py
pjenvey: just making it consistent with the last couple lines

autopush/tests/test_integration.py
```diff
@@ -508,6 +506,70 @@ def legacy_endpoint(self):
         self._settings._notification_legacy = False
 
 
+class SSLEndpointMixin:
```
bbangert: We're not doing (object) anymore?

pjenvey: was actually avoiding it because I assumed twisted's TestCase was an old style class (not sure it would have even have mattered), now I see it's not, good catch

PR #827, opened 2017-02-17
feat: Add support for VAPID DRAFT 02 

### LINKED ISSUE #826: Update VAPID Support to DRAFT-02
https://tools.ietf.org/html/draft-ietf-webpush-vapid-02

### LINKED ISSUE #862: Sudden socket closure may result in AttributeError
https://sentry.prod.mozaws.net/operations/autopush-dev/issues/390836/

A socket may be suddenly closed while a pending rec'v is in progress, leading to a socket being set to None by the re-initialization, resulting in an AttributeError.

Treat such as a Connection Error.

jrconlin:
VAPID Draft-02 <https://tools.ietf.org/html/draft-ietf-webpush-vapid-02>
alters how VAPID information can be provided to the server.

closes #826

### DIFF
--- a/autopush/settings.py
+++ b/autopush/settings.py
@@ -39,7 +39,8 @@ from autopush.utils import (
     canonical_url,
     resolve_ip,
     repad,
-    base64url_decode
+    base64url_decode,
+    parse_auth_header,
 )
 from autopush.crypto_key import (CryptoKey, CryptoKeyException)
 
@@ -365,7 +366,19 @@ class AutopushSettings(object):
             except CryptoKeyException:
                 raise InvalidTokenException("Invalid key data")
             public_key = crypto_key.get_label('p256ecdsa')
-
+        if auth_header:
+            vapid_auth = parse_auth_header(auth_header)
+            if not vapid_auth:
+                raise VapidAuthException("Invalid Auth token")
+            self.metrics.increment("updates.notification.auth.{}".format(
+                vapid_auth['scheme']
+            ))
+            # pull the public key from the VAPID auth header if needed
+            try:
+                if vapid_auth['version'] != 1:
+                    public_key = vapid_auth['k']
+            except KeyError:
+                raise VapidAuthException("Missing Public Key")
         if version == 'v1' and len(token) != 32:
             raise InvalidTokenException("Corrupted push token")
         if version == 'v2':
--- a/autopush/tests/test_web_validation.py
+++ b/autopush/tests/test_web_validation.py
@@ -624,7 +624,6 @@ class TestWebPushRequestSchema(unittest.TestCase):
         info = self._make_test_data(
             body="asdfasdfasdfasdf",
             headers={
-                "authorization": "not vapid",
                 "content-encoding": "aesgcm128",
                 "encryption": "salt=" + padded_value,
                 "encryption-key": "dh=asdfasdfasdf",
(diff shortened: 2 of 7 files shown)


autopush/settings.py
```diff
@@ -354,7 +354,16 @@ def parse_endpoint(self, token, version="v1", ckey_header=None,
             except CryptoKeyException:
                 raise InvalidTokenException("Invalid key data")
             public_key = crypto_key.get_label('p256ecdsa')
-
+        if auth_header and auth_header[:6] == 'vapid ':
+            try:
+                items = {}
+                bits = auth_header.replace(' ', '').split(',')
```
pjenvey: XXX:

autopush/tests/test_web_validation.py
```diff
@@ -865,6 +865,99 @@ def test_valid_vapid_crypto_header_webpush(self):
...
+    def test_valid_vapid_02_crypto_header_webpush(self):
+        schema = self._make_fut()
+
+        header = {"typ": "JWT", "alg": "ES256"}
+        payload = {"aud": "https://pusher_origin.example.com",
+                   "exp": int(time.time()) + 86400,
+                   "sub": "mailto:admin@example.com"}
+
+        token, crypto_key = self._gen_jwt(header, payload)
+        auth = "vapid t={token},k={key}".format(token=token,
```
pjenvey: parse_endpoint collapses the whitespace of the header before splitting on =, so a valid case of "vapid k=..." will break

autopush/web/webpush.py
```diff
@@ -298,13 +298,35 @@ def token_prep(self, d):
...
+                                 status_code=401, errno=109,
+                                 headers={"www-authenticate": PREF_SCHEME})
+        return jwt
+
+    def vapid_draft_02(self, rawtoken):
+        try:
+            tokens = rawtoken.split(',')
+            elements = dict()
+            for token in tokens:
+                kv = token.replace(' ', '').split('=')
```
pjenvey: probably want maxsplit of 1 to be safe, like parse_endpoint does. kv[1] can potentially raise an IndexError too (even with a maxsplit)

autopush/tests/test_web_validation.py
```diff
@@ -865,6 +865,99 @@ def test_valid_vapid_crypto_header_webpush(self):
...
+                "authorization": auth,
+            }
+        )
+
+        with assert_raises(InvalidRequest) as cm:
+            schema.load(info)
+        eq_(cm.exception.status_code, 401)
+        eq_(cm.exception.errno, 109)
+
+    def aaa_test_invalid_vapid_draft2_crypto_header(self):
```
pjenvey: heeey aaa fonz

bbangert:
Is Draft 02 complete? Any further changes are made only to a new draft now?

autopush/tests/test_web_validation.py
jrconlin: huh, thought I nuked that in a follow up. sorry.

jrconlin:
Draft02 has been recorded. No additional notice has been made so it's "stable enough". It also meets an objective with ECE in removing the need for additional headers. 

This patch is not required immediately, but prepares for the inevitable change.

autopush/tests/test_web_validation.py
jrconlin: I'm not quite sure I'm seeing the problem you've spotted. Draft-02 has two parameters, `t` & `k`. I am missing a test that includes an extra, ignored parameter (and I'll switch parameter order as well), but there's already a test that checks for missing parameter. I switch `k` for a nonsense `n` in a later test that should fail. I'll add comments to the tests to clarify what's being tested.

pjenvey: the case I meant: auth = "vapid k={key},t={token}".format(token=token

jrconlin:
refactored for the refactor: 
@pjenvey @bbangert r?

autopush/utils.py
```diff
@@ -265,6 +265,9 @@ class WebPushNotification(object):
     # Whether this notification should follow legacy non-topic rules
     legacy = attrib(default=False)  # type: bool
 
+    # the version of the VAPID header
+    vapid_version = attrib(default='draft01')  # type: str
```
pjenvey: This field is never filled in (will always be draft01) -- arguably it's not the best place for it either.

Since it's only needed by the webpush handler method, I'd just have that method take it as an explicit kwarg, it's already in its **kwargs  as part of the WebPushRequestSchema

autopush/settings.py
```diff
@@ -341,6 +342,29 @@ def make_endpoint(self, uaid, chid, key=None):
         ep = self.fernet.encrypt(base + sha256(raw_key).digest()).strip('=')
         return root + 'v2/' + ep
 
+    def extract_auth(self, auth_header):
+        vapid_auth = {}
```
bbangert: Maybe this should be an attrs style object (like WebPushNotification), would help ensure proper dict access to the few fixed keys here, and the logic could be better separated from settings.

autopush/settings.py
```diff
@@ -359,13 +383,23 @@ def parse_endpoint(self, token, version="v1", ckey_header=None,
...
+        self.vapid_auth = None
         if ckey_header:
             try:
                 crypto_key = CryptoKey(ckey_header)
             except CryptoKeyException:
                 raise InvalidTokenException("Invalid key data")
             public_key = crypto_key.get_label('p256ecdsa')
-
+        if auth_header:
+            self.vapid_auth = self.extract_auth(auth_header)
```
bbangert: Um, so we're setting a global settings object's vapid_auth (used by multiple threads) to a request specific object?

autopush/web/webpush.py
```diff
@@ -315,27 +314,18 @@ def token_prep(self, d):
...
-
-        # If its not a bearer token containing what may be JWT, stop
-        if auth_type.lower() not in AUTH_SCHEMES or '.' not in token:
-            if needs_auth:
-                raise InvalidRequest("Missing Authorization Header",
-                                     status_code=401, errno=109)
-            return
-
         try:
+            token = self.context['settings'].vapid_auth['t']
```
bbangert: I don't understand how this got here, it's not being carried around on the global object, is it?

autopush/utils.py
```diff
@@ -179,7 +180,7 @@ def extract_jwt(token, crypto_key):
     # first split and convert the jwt.
     if not token or not crypto_key:
         return {}
-    return jwt.decode(token, decipher_public_key(crypto_key))
+    return jwt.decode(token, decipher_public_key(crypto_key.encode('utf8')))
```
bbangert: We're changing the crypto_key's encoding format?

autopush/tests/test_web_validation.py
```diff
@@ -288,6 +288,7 @@ def _make_fut(self):
         from autopush.web.webpush import WebPushRequestSchema
         schema = WebPushRequestSchema()
         schema.context["settings"] = Mock()
+        schema.context["settings"].vapid_auth = {}
```
jrconlin: crap, detrius.

autopush/utils.py
jrconlin: No, the problem is that the base64 decoder which gets called later rejects strings, so we recode it for an accepted value.

PR #829, opened 2017-02-21
bug: Do not attempt to register failed GCM registrations

### LINKED ISSUE #828: filter router data
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/383827/

need to filter against router type bozo attacks

jrconlin:
GCM messages that trigger a "fail" were attempting to update
registration with insufficient data. It's better to treat them like
other GCM errors and report back, but take no additional actions.
Also increased testing around these sorts of errors.

closes #828

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -7,7 +7,6 @@ from twisted.logger import Logger
 
 from autopush.exceptions import RouterException
 from autopush.router.interface import RouterResponse
-from autopush.utils import ms_time
 
 
 class GCMRouter(object):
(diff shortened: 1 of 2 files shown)


autopush/router/gcm.py
```diff
@@ -173,25 +172,17 @@ def _process_reply(self, reply, uaid_data, ttl, notification):
...
                                   log_exception=False,
                                   )
 
         # retries:
         if reply.needs_retry():
-            self.log.warn("GCM retry requested: {failed()}",
-                          failed=lambda: repr(reply.failed.items()))
             self.metrics.increment("updates.client.bridge.gcm.failed.retry",
                                    self._base_tags)
+            self.log.warn("GCM retry requested: {failed()}",
```
jrconlin: Moved this to make more consistent with other error cases.

bbangert:
Our contributing doc says it should be 'fix: '..... but ah well.

Issue #511, opened 2016-07-11
Audit and fix gcmclient library


jrconlin:
Looks like gcmclient library may be eating or returning invalid errors from the gcm servers. 
Need to either find a better library or fork/review and fix the issues with this one.

jrconlin:
Reviewed the gcmclient library. GCM only has a limited set of error codes. These codes are captured properly [by the library](https://bitbucket.org/sardarnl/gcm-client/src/c1ae4a4c9804b623456ffdd3a4531aa554816087/gcmclient/gcm.py?at=default&fileviewer=file-view-default#gcm.py-509:518). 

[Errors are passed back](https://bitbucket.org/sardarnl/gcm-client/src/c1ae4a4c9804b623456ffdd3a4531aa554816087/gcmclient/gcm.py?at=default&fileviewer=file-view-default#gcm.py-150) to the GCM handler, albeit extracted from the original response. Since we're only sending one message at a time, this is not a major concern. 

It might be worthwhile to do the same extended error information we use for the new FCM library, but [we are logging](https://github.com/mozilla-services/autopush/blob/master/autopush/router/gcm.py#L156) the abbreviated error code we get back from GCM. 

I do not believe any additional action is required.

Closing INVALID

PR #831, opened 2017-02-24
bug: Don't send non-priority messages to sentry

### LINKED ISSUE #830: Don't send bad APNS token exceptions to Sentry
Per Sentry issue:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/384162/

These should be logged, but don't need to be sent into Sentry.

jrconlin:
closes #830

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -494,6 +494,7 @@ def connection_main(sysargs=None, use_files=True):
         env=args.env,
         hello_timeout=args.hello_timeout,
         preflight_uaid="deadbeef000000000deadbeef" + postfix,
+        debug=args.debug,
     )
     if not settings:
         return 1  # pragma: nocover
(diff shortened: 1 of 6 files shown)


autopush/main.py
```diff
@@ -532,6 +533,9 @@ def connection_main(sysargs=None, use_files=True):
     resource = DefaultResource(WebSocketResource(factory))
     resource.putChild("status", StatusResource())
     site_factory = Site(resource)
+    # Silence starting/stopping messages
+    site_factory.noisy = args.debug
+    site.noisy = args.debug
```
pjenvey: Looks good but let's also set noisy on the other 2 cyclone.web.Applications in this file as well

jrconlin: fair enough. I thought I was being overly assertive with the base class, but 100% OK with being even more assertive.

pjenvey: oops, this assignment's now a dupe of the one on line #511

pjenvey: oops, this assignment's now a dupe of the one on line #511

PR #834, opened 2017-03-01
feat: capture and metric item not found instead of log

### LINKED ISSUE #811: Capture uaid Item error in webpush
When calling lookup_node in websocket.py, an invalid uaid goes to log_failure and is not caught or handled appropriately. This is logged hundreds of thousands of times in sentry and should be handled.

bbangert:
There's no need to log exceptions failing to save a message for a
user that has been deleted. Now a metric will be emitted instead.

This also tweaks the tox runner to no longer include object memory
tracking by default.

Closes #811

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -7,7 +7,10 @@ from collections import defaultdict
 
 import twisted.internet.base
 from autopush.tests.test_db import make_webpush_notification
-from boto.dynamodb2.exceptions import ProvisionedThroughputExceededException
+from boto.dynamodb2.exceptions import (
+    ProvisionedThroughputExceededException,
+    ItemNotFound
+)
 from boto.exception import JSONResponseError
 from cyclone.web import Application
 from mock import Mock, patch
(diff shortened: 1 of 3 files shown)


autopush/tests/test_websocket.py
```diff
@@ -459,7 +460,41 @@ def wait_for_agent_call():  # pragma: nocover
                 reactor.callLater(0.1, wait_for_agent_call)
 
             mock_metrics.increment.assert_called_with(
-                "error.notify_uaid_failure", tags=None)
+                "client.notify_uaid_failure", tags=None)
+            d.callback(True)
+        reactor.callLater(0.1, wait_for_agent_call)
+        return d
+
+    def test_close_with_delivery_cleanup_and_no_get_uaid_error(self):
```
jrconlin: nit: not?

bbangert: Doh!

autopush/tests/test_websocket.py
```diff
@@ -7,7 +7,8 @@
 
 import twisted.internet.base
 from autopush.tests.test_db import make_webpush_notification
-from boto.dynamodb2.exceptions import ProvisionedThroughputExceededException
+from boto.dynamodb2.exceptions import ProvisionedThroughputExceededException, \
```
pjenvey: prefer parens instead of \  :)

bbangert: I blame pycharm....

PR #839, opened 2017-03-07
feat: Add timeout for internally routed messages.

### LINKED ISSUE #821: Ensure timeouts are under 5 seconds in endpoint
Worst case timings for our endpoints right now are around 5+ seconds. This is likely because we have not manually dropped the timeout for intranode messaging from the endpoint. We should ensure none of the intranode timeouts are more than 5 seconds.

jrconlin:
closes #821

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -134,6 +134,9 @@ def add_shared_args(parser):
     parser.add_argument('--router_write_throughput',
                         help="DynamoDB router write throughput",
                         type=int, default=5, env_var="ROUTER_WRITE_THROUGHPUT")
+    parser.add_argument('--connection_timeout',
+                        help="Seconds to wait for connection timeout",
+                        type=int, default=1, env_var="CONNECTION_TIMEOUT")
     parser.add_argument('--max_data', help="Max data segment length in bytes",
                         default=4096, env_var='MAX_DATA')
     parser.add_argument('--env',
@@ -448,6 +451,7 @@ def make_settings(args, **kwargs):
         ami_id=ami_id,
         client_certs=client_certs,
         msg_limit=args.msg_limit,
+        connect_timeout=args.connection_timeout,
         **kwargs
     )
 
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -18,6 +18,7 @@ from twisted.internet.threads import deferToThread
 from twisted.internet.defer import (
     inlineCallbacks,
     returnValue,
+    CancelledError,
 )
 from twisted.internet.error import (
     ConnectError,
@@ -81,13 +82,14 @@ class SimpleRouter(object):
             try:
                 result = yield self._send_notification(uaid, node_id,
                                                        notification)
-            except (ConnectError, ConnectionClosed, ResponseFailed) as exc:
+            except (ConnectError, ConnectionClosed, ResponseFailed,
+                    CancelledError) as exc:
                 self.metrics.increment("updates.client.host_gone")
                 yield deferToThread(router.clear_node,
                                     uaid_data).addErrback(self._eat_db_err)
                 if isinstance(exc, ConnectionRefusedError):
                     # Occurs if an IP record is now used by some other node
-                    # in AWS.
+                    # in AWS or if the connection timesout.
                     self.log.debug("Could not route message: {exc}", exc=exc)
             if result and result.code == 200:
                 self.metrics.increment("router.broadcast.hit")
--- a/autopush/protocol.py
+++ b/autopush/protocol.py
@@ -20,7 +20,8 @@ class IgnoreBody(Protocol):
     def ignore(cls, response):
         """Class method helper for ignoring the response"""
         d = Deferred()
-        response.deliverBody(cls(response, d))
+        if response is not None:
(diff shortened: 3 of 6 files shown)


pjenvey:
Funny thing here is ap_settings.agent already has a connectTimeout feature we're using and it's already hardcoded to 5 seconds.

That explains the 5 second requests from the original issue https://github.com/mozilla/fxa-auth-server/issues/1657

.. but why do a few take 10 seconds?

jrconlin:
I'm not sure, but I believe the problem is that `connectTimeout` is only for the actual connection, not the entire transaction. So if a connection takes 1s and the actual transaction takes 10s (for whatever reason), then `connectTimeout` wouldn't fire. Adding the `callLater` here absolutely kills a connection, regardless of connection state. 

Would this solve the problem?

autopush/router/simple.py
```diff
@@ -81,13 +83,17 @@ def route_notification(self, notification, uaid_data):
...
+            except CancelledError as exc:
+                # Log the timeout, but don't clear the node.
+                self.log.debug("Could not route message: Timedout {exc}",
+                               exc=exc)
             except (ConnectError, ConnectionClosed, ResponseFailed) as exc:
                 self.metrics.increment("updates.client.host_gone")
                 yield deferToThread(router.clear_node,
                                     uaid_data).addErrback(self._eat_db_err)
-                if isinstance(exc, ConnectionRefusedError):
+                if isinstance(exc, (ConnectionRefusedError, CancelledError)):
```
pjenvey: CancelledError never occurs here with the new except block above

I think we now want some additional CancelledError handling around _send_notification_check as well, though

pjenvey:
WebPushRouter.send_notification needs the timing out too

bbangert:
@pjenvey because if a node fails to respond, that's 5 seconds, we stash the message in storage, then attempt to tell the node to check storage (another 5 seconds).

autopush/main.py
```diff
@@ -134,6 +134,10 @@ def add_shared_args(parser):
     parser.add_argument('--router_write_throughput',
                         help="DynamoDB router write throughput",
                         type=int, default=5, env_var="ROUTER_WRITE_THROUGHPUT")
+    parser.add_argument('--router_timeout',
+                        help="Seconds to wait for a router to process a "
+                             "message",
```
bbangert: This isn't really timing out the entire router, is it? It's just changing the timeout on requests the router might make?

autopush/router/simple.py
```diff
@@ -188,21 +194,42 @@ def _send_notification(self, uaid, node_id, notification):
...
             "PUT",
             url.encode("utf8"),
             bodyProducer=FileBodyProducer(StringIO(payload)),
         )
-        d.addCallback(IgnoreBody.ignore)
-        return d
+        # Only time out actual calls, not mocks.
+        if isinstance(request, defer.Deferred):
+            timeout = reactor.callLater(self.ap_settings.router_timeout,
+                                        request.cancel)
```
bbangert: I guess I had assumed there'd be away to set a socket-level timeout, rather than having to cancel it from Python. Am I reading this correctly that twisted's HTTP agent has no way to set a time-out on the socket connection portion?

pjenvey: Like I said there's connectTimeout

https://github.com/mozilla-services/autopush/blob/master/autopush/settings.py#L108

bbangert: I had figured we'd just drop the connect Timeout, rather than use more code to cancel requests like this (and even though they're cancelled, afaik twisted is still waiting on trying to connect to them).

jrconlin: Well, [connectionTimeout](https://twistedmatrix.com/documents/current/api/twisted.web.client.Agent.html) is just how long to wait until a peer accepts a connection. The timeout is handed down to the connection protocol, which is essentially doing [the same sort of timeout](https://github.com/twisted/twisted/blob/twisted-17.1.0/src/twisted/internet/protocol.py#L271) as what I've got for the whole thing. 

Dropping the connectTimeout means that we don't timeout at all, which I don't think we want. Connection nodes bounce and we probably want to have some detection of that, right?

autopush/main.py
jrconlin: No, it's timing out the whole router transaction. The `connectTimeout` parameter limits how long it takes to connect to a peer, where this would limit the entire transaction time. Which, I believed, was the problem. If not, then I'm not sure I understand what the issue was.

autopush/router/simple.py
bbangert: I meant we'd drop the value of the connectTimeout substantially, ie maybe 500ms, this means if it ends up making two dead connect calls (our worst-case node situation in routing), our max time taken is 1sec, not 10sec. Sorry, wasn't clear enough what I meant to 'drop'.

I'm wary of canceling the entire routing call because the logic in it wasn't made with the assumption that we'd nuke the whole run-through externally. It only was created with the assumption that talking to nodes sometimes might fail.

jrconlin: Ah, ok. So I'll pull this PR in favor of just a simple config change.

jrconlin:
Opting for just reducing PR to a new config argument for connection timeouts

Issue #844, opened 2017-03-14
Return Channel IDs for a Bridged connection


jrconlin:
for a given registration URL 
`GET /v1/{router_type}/{token}/registration/{uaid}/` (auth'd by secret token), return the server known channel IDs.

Issue #843, opened 2017-03-13
Web Push form Firefox Mobile


bjjclimbin:
Hello.

I have been implementing web push notifications through my portal. The application works great for firefox Desktop and other browsers (such as chrome desktop and mobile) but when i try to push notifications to a mobile user, the server's response is {"code": 410, "errno": 105, "error": "", "more_info": "http://autopush.readthedocs.io/en/latest/http.html#error-codes", "message": "Request did not validate Critical Failure: Client is unreachable due to a configuration error. Unable to send message."} . I have tried several different methods to push notifications even paid service providers and the notifications do not arrive on the mobile devices (firefox permissions have been set correctly) .

jrconlin:
Hi, I presume that you're talking about a portal that you're running yourself (our logs aren't showing a 410:105 error in the past day or so). 

If you are running your own instance of autopush on your own server, then you'll need to configure the various external bridges yourself. The [shared configuration file](https://github.com/mozilla-services/autopush/blob/master/configs/autopush_shared.ini.sample#L86) has information about what you need to set up, Each bridge acts as it's own GCM app, so you'll need to configure a senderID (with associated keys and permissions).

Be aware that bridges are a proprietary extension for autopush. This means that you will need to modify the Firefox mobile client to include the corresponding credentials as well.  While you're free to experiment with this interface, it's not officially supported for third party use. The code is obviously available and is commented, but unfortunately, we don't really have the available resources to help you navigate all the hurdles. (Obviously, we'd LOVE if you would publish any notes or comments about your experience so that others can benefit!)

kitcambridge:
Could you also paste the endpoint, please?

bjjclimbin:
Hello.

This is one of the endpoints i have been trying.

https://updates.push.services.mozilla.com/wpush/v2/gAAAAABYwtR-
Bz7ldT4XuZSZ3vBAI83cefC1s7vCk9VPRoUSjhOn6VsUPmN5rx0-_
fF42XllPznMIZGwMgpR80jAsOpBW-sMiYM3G41JUDpkp4hSEOR3SGdBSl0Q
7Tiz9A8kXOfvFFif821fAwUYrGgOYI0Nib51xdHt46Aa2U4mHrY6lRq-hFs" --request POST
--header "TTL: 60" --header "Content-Length: 0" --header "Authorization:
[...]

PR #847, opened 2017-03-16
feat: Add ChannelID report for UAID

jrconlin:
Adds new bridge system HTTP endpoint to fetch a list of server known
CHIDs for a given UAID. Useful for remote clients to check status. (see
documentation for calling structure and return).

Also fixed an edge case where clients that may have been forced offline
due to external system errors could still register new endpoints. The
behavior will now be to return a 410 error if a client has been flagged
as disconnected by the server. It is up to the client to check the local
bridge connection and re-establish any endpoints.

closes: #844, #843

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -495,7 +495,8 @@ class Message(object):
         # Functions that call store_message() would be required to
         # update that list as well using register_channel()
         try:
-            result = self.table.get_item(consistent=True, uaid=hasher(uaid),
+            result = self.table.get_item(consistent=True,
+                                         uaid=hasher(uaid),
                                          chidmessageid=" ")
             return (True, result["chids"] or set([]))
         except ItemNotFound:
--- a/autopush/web/registration.py
+++ b/autopush/web/registration.py
@@ -51,16 +51,27 @@ class RegistrationSchema(Schema):
         chid = req['path_kwargs'].get('chid', params.get("channelID"))
         if uaid:
             try:
-                uuid.UUID(uaid)
+                u_uuid = uuid.UUID(uaid)
             except (ValueError, TypeError):
                 raise InvalidRequest("Invalid Request UAID",
                                      status_code=401, errno=109)
+            # Check if the UAID has a 'critical error' which means that it's
+            # probably invalid and should be reset/re-registered
+            try:
+                record = self.context['settings'].router.get_uaid(u_uuid.hex)
+                if record.get('critical_failure'):
+                    raise InvalidRequest("Invalid Request UAID",
+                                         status_code=410, errno=105)
+            except ItemNotFound:
+                pass
+
         if chid:
             try:
                 uuid.UUID(chid)
             except (ValueError, TypeError):
                 raise InvalidRequest("Invalid Request Channel_id",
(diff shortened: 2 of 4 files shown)


autopush/db.py
```diff
@@ -495,7 +495,8 @@ def all_channels(self, uaid):
         # Functions that call store_message() would be required to
         # update that list as well using register_channel()
         try:
-            result = self.table.get_item(consistent=True, uaid=hasher(uaid),
+            result = self.table.get_item(consistent=True,
+                                         uaid=hasher(str(uaid)),
```
bbangert: str again? The type sig indicates its already a str? Does that mean its being used incorrectly somewhere?

autopush/web/registration.py
```diff
@@ -307,6 +336,15 @@ def _return_endpoint(self, endpoint_data, new_uaid, router=None):
                        client_info=self._client_info)
         self.finish()
 
+    def _list_channels(self):
+        _, channels = self.ap_settings.message.all_channels(self.uaid)
+        dashed = [str(uuid.UUID(x)) for x in channels]
+        self.write(json.dumps(
```
pjenvey: _list_channels is called in a separate thread so you really want to go back to the event loop for write/finish

autopush/web/registration.py
```diff
@@ -307,6 +336,15 @@ def _return_endpoint(self, endpoint_data, new_uaid, router=None):
...
+            {"uaid": self.uaid.hex,
```
pjenvey: nitpick: returning uaid's superfluous

jrconlin: Not necessarily. For some "state challenged" remote calls, it's not always easy to associate a response back to it's request. Having a shared ID between the two can simplify things. In this case, the UAID is required for the URL and the response, plus it's how the client organizes.

jrconlin:
@pjenvey Well, they need to have a "secret" bearer token which is only provided when calling HTTP registration (not websocket), so it's already limited.

autopush/web/registration.py
```diff
@@ -307,6 +337,15 @@ def _return_endpoint(self, endpoint_data, new_uaid, router=None):
                        client_info=self._client_info)
         self.finish()
 
+    def _list_channels(self, uaid):
+        _, channels = self.ap_settings.message.all_channels(str(uaid))
```
pjenvey: this db call should definitely be in a thread though, we just don't really want the write/finish in one

jrconlin: ah, misread how Deferred() works. switch to deferToThread()

autopush/web/registration.py
```diff
@@ -216,6 +227,26 @@ def _register_channel(self, router_data=None):
...
 
     @threaded_validate(RegistrationSchema)
+    def get(self, *args, **kwargs):
+        """HTTP GET
+
+        Return a list of known channelIDs for a given UAID
+
+        """
+        self.uaid = self.valid_input['uaid']
+        if not self.uaid:
```
bbangert: The validation doesn't actually validate what is required (the UAID in this case)? We added validation schema's to avoid if/else checks in the code here.

PR #850, opened 2017-03-23
Removed typo in URL

tarekziade:


pjenvey:
FYI this doesn't follow our commit message guidelines:

https://github.com/mozilla-services/autopush/blob/master/CONTRIBUTING.md

(but I'm not going to hold up a one char fix over that)

bbangert:
Yea, it won't show correctly on the changelog as is, but given its a 1 character change it's ok to not appear on the changelog.

PR #854, opened 2017-03-28
feat: Use cryptography based JWT parser for increased speed

### LINKED ISSUE #785: Convert pycrypto to cryptography
based on comments left in https://github.com/mozilla-services/autopush/issues/560#issuecomment-273321239 and observation of a slow memory leak in pypy, it may be highly valuable to convert the pycrypto call-outs to use native cryptography libraries.

jrconlin:
Switches to using python cryptography library instead of jose/jwt
(which relied on python ecdsa library). Also discovered lots of fun
discrepencies between how ecdsa and libssl sign things.

closes #785

### DIFF
--- a/autopush/jwt.py
+++ b/autopush/jwt.py
@@ -1,0 +1,107 @@
+import base64
+import binascii
+import json
+
+from cryptography.exceptions import InvalidSignature
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import ec, utils
+from cryptography.hazmat.primitives import hashes
+from pyasn1.error import PyAsn1Error
+from twisted.logger import Logger
+
+
+def repad(string):
+    # type: (str) -> str
+    """Adds padding to strings for base64 decoding"""
+    if len(string) % 4:
+        string += '===='[len(string) % 4:]
+    return string
+
+
+class VerifyJWT(object):
+    """Minimally verify a Vapid JWT object.
+
+        Why hand roll? Most python JWT libraries either use a python elliptic
+        curve library directly, or call one that does, or is abandoned, or a
+        dozen other reasons.
+
+        After spending half a day looking for reasonable replacements, I
+        decided to just write the functions we need directly.
+
+        THIS IS NOT A FULL JWT REPLACEMENT.
+
+    """
+
+    @staticmethod
+    def extract_signature(auth):
+        # type: (str) -> tuple()
+        """Fix the JWT auth token.
+
+        The JWA spec defines the signature to be a pair of 32octet encoded
+        longs.
+        The `ecdsa` library signs using a raw, 32octet pair of values (s, r).
+        Cryptography, which uses OpenSSL, uses a DER sequence of (s, r).
+        This function converts the raw ecdsa to DER.
+
+        :param auth: A JWT authorization token.
+        :type auth: str
+
+        :return tuple containing the signature material and signature
+
+        """
+        payload, asig = auth.encode('utf8').rsplit(".", 1)
+        sig = base64.urlsafe_b64decode(repad(asig))
+        if len(sig) != 64:
+            return payload, sig
+
+        encoded = utils.encode_dss_signature(
+            s=int(binascii.hexlify(sig[32:]), 16),
+            r=int(binascii.hexlify(sig[:32]), 16)
+        )
+        return payload, encoded
+
+    @staticmethod
+    def decode(token, key):
+        # type (str, str) -> dict()
+        """Decode a web token into a assertion dictionary.
+
+        This attempts to rectify both ecdsa and openssl generated
+        signatures. We use the built-in cryptography library since it wraps
+        libssl and is faster than the python only approach.
+
+        :param token: VAPID auth token
+        :type token: str
+        :param key: bitarray containing public key
+        :type key: str or bitarray
+
+        :return dict of the VAPID claims
+
+        :raise InvalidSignature
+
+        """
+        # convert the signature if needed.
+        try:
+            sig_material, signature = VerifyJWT.extract_signature(token)
+            pkey = ec.EllipticCurvePublicNumbers.from_encoded_point(
+                ec.SECP256R1(),
+                key
+            ).public_key(default_backend())
+            # NOTE: verify() will take any string as the signature. It appears
+            # to be doing lazy verification and matching strings rather than
+            # comparing content values. If the signatures start failing for
+            # some unknown reason in the future, decode the signature and
+            # make sure it matches how we're reconstructing it.
+            # This will raise an InvalidSignature exception if failure.
+            # It will be captured externally.
+            pkey.verify(
+                signature,
+                sig_material.encode('utf8'),
+                ec.ECDSA(hashes.SHA256()))
+            return json.loads(
+                base64.urlsafe_b64decode(
+                    repad(sig_material.split('.')[1]).encode('utf8')))
+        except (ValueError, TypeError, binascii.Error, PyAsn1Error):
+            raise InvalidSignature()
+        except Exception:  # pragma: no cover
+            Logger().failure("Unexpected error processing JWT")
+            raise InvalidSignature()
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -3,7 +3,7 @@ import time
 
 from boto.dynamodb2.exceptions import ItemNotFound
 from cryptography.fernet import InvalidToken
-from jose import JOSEError
+from cryptography.exceptions import InvalidSignature
 from marshmallow import (
     Schema,
     fields,
@@ -321,7 +321,7 @@ class WebPushRequestSchema(Schema):
 
         try:
             jwt = extract_jwt(token, public_key)
-        except (AssertionError, ValueError, JOSEError):
+        except (ValueError, InvalidSignature, Exception):
             raise InvalidRequest("Invalid Authorization Header",
                                  status_code=401, errno=109,
                                  headers={"www-authenticate": PREF_SCHEME})
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -1,11 +1,10 @@
 import json
 import uuid
 
-import jose
 import twisted.internet.base
 from cryptography.fernet import Fernet, InvalidToken
 from cyclone.web import Application
-from mock import Mock, patch
+from mock import Mock
 from moto import mock_dynamodb2
 from nose.tools import eq_, ok_
 from twisted.internet.defer import Deferred
@@ -534,28 +533,6 @@ class RegistrationTestCase(unittest.TestCase):
         self.finish_deferred.addCallback(handle_finish)
         self.reg.request.headers["Authorization"] = "WebPush Invalid"
         self.reg.post(self._make_req(router_type="simplepush",
-                                     uaid=dummy_uaid.hex,
-                                     chid=str(dummy_chid)))
-        return self.finish_deferred
-
-    @patch('jose.jws.verify', side_effect=jose.exceptions.JWTError)
-    def test_post_bad_jwt(self, *args):
-        self.reg.request.body = json.dumps(dict(
-            channelID=str(dummy_chid),
-        ))
-
-        def handle_finish(value):
-            self._check_error(401, 109, 'Unauthorized')
-
-        def restore(*args, **kwargs):
-            uuid.uuid4 = old_func
-
-        old_func = uuid.uuid4
-        uuid.uuid4 = lambda: dummy_uaid
-        self.finish_deferred.addBoth(restore)
-        self.finish_deferred.addCallback(handle_finish)
-        self.reg.request.headers["Authorization"] = "WebPush Dummy"
-        self.reg.post(self._make_req(router_type="webpush",
                                      uaid=dummy_uaid.hex,
                                      chid=str(dummy_chid)))
         return self.finish_deferred
--- a/autopush/tests/test_web_validation.py
+++ b/autopush/tests/test_web_validation.py
@@ -8,8 +8,8 @@ from boto.dynamodb2.exceptions import (
     ItemNotFound,
 )
 from cryptography.fernet import InvalidToken
+from cryptography.exceptions import InvalidSignature
 from jose import jws
-from jose.exceptions import JWTClaimsError
 from marshmallow import Schema, fields
 from mock import Mock, patch
 from moto import mock_dynamodb2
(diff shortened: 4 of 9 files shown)


autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        # It's ok, though, because neither label them, even though openssl
+        # uses namedtypes, with the names set to ""
+        #
+        # Oh frabjous day!
+        ds = univ.Sequence(
+            componentType=namedtype.NamedTypes()
+        ).setComponents(
+            univ.Integer(int(binascii.hexlify(sig[:32]), 16)),
+            univ.Integer(int(binascii.hexlify(sig[32:]), 16)),
+        )
```
alex: cryptography has a function for taking `r` and `s` and generating the DER encoded body: `cryptography.hazmat.primitives.asymmetric.utils.encode_dss_signature` (path from memory, hopefully correct!)

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+import json
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives import hashes
+
+from pyasn1.type import univ, namedtype
+from pyasn1.codec.der.encoder import encode
+
+
+"""Why hand roll?
```
bbangert: nit: Module docs go above the imports

autopush/jwt.py
alex: (In addition to being less code for you, it uses a different python asn.1 lib which is faster, so that'll be helpful if perf is a concern)

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+
+After spending half a day looking for reasonable replacements, I decided to
+just write the functions we need directly.
+
+THIS IS NOT A FULL JWT REPLACEMENT.
+
+"""
+
+
+def repad(string):
```
bbangert: Wasn't there another padding altering function somewhere?

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+    # type: (str) -> str
+    """Adds padding to strings for base64 decoding"""
+    if len(string) % 4:
+        string += '===='[len(string) % 4:]
+    return string
+
+
+class VerifyJWT:
```
bbangert: Should inherit from (object) still (in Py3 that's implicit).

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            ec.SECP256R1(),
+            key
+        ).public_key(default_backend())
+        # eventually this would be a map of algs to signing algorithms.
+        # NOTE: verifier() will take any string as the signature. It appears
+        # to be doing lazy verification and matching strings rather than
+        # comparing content values. If the signatures start failing for
+        # some unknown reason in the future, decode the signature and
+        # make sure it matches how we're reconstructing it.
+        verifier = pkey.verifier(
```
alex: Recent cryptography's have a helper `pkey.verify()` method, which will save you this comment and a few lines of code :-) https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePublicKey.verify

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+    # type: (str) -> str
+    """Adds padding to strings for base64 decoding"""
+    if len(string) % 4:
+        string += '===='[len(string) % 4:]
+    return string
+
+
+class VerifyJWT:
+
+    def __init__(self):  # pragma: no cover
```
bbangert: Why is an empty init needed?

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        pass
+
+    @classmethod
```
bbangert: Why not ``@staticmethod``? Since these don't seem to use cls.

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        # make sure it matches how we're reconstructing it.
+        verifier = pkey.verifier(
+            base64.urlsafe_b64decode(repad(signature.encode())),
+            ec.ECDSA(hashes.SHA256()))
+        verifier.update(sig_material.encode())
+        # This will raise an InvalidSignature exception if failure.
+        # It will be captured externally.
+        verifier.verify()
+        return json.loads(
+            base64.urlsafe_b64decode(repad(sig_material.split('.')[1])))
```
alex: You can probably factor out the `base64.urlsafe_b64decode(repad(signature.encode()))` above into a `message` var, to safe teh re-decoding.

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            componentType=namedtype.NamedTypes()
+        ).setComponents(
+            univ.Integer(int(binascii.hexlify(sig[:32]), 16)),
+            univ.Integer(int(binascii.hexlify(sig[32:]), 16)),
+        )
+        encoded = encode(ds)
+        new_sig = base64.urlsafe_b64encode(encoded)
+        return "{}.{}".format(payload, new_sig)
+
+    @classmethod
```
bbangert: staticmethod?

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        # Oh frabjous day!
+        ds = univ.Sequence(
+            componentType=namedtype.NamedTypes()
+        ).setComponents(
+            univ.Integer(int(binascii.hexlify(sig[:32]), 16)),
+            univ.Integer(int(binascii.hexlify(sig[32:]), 16)),
+        )
+        encoded = encode(ds)
+        new_sig = base64.urlsafe_b64encode(encoded)
+        return "{}.{}".format(payload, new_sig)
```
alex: Since the caller immediately calls `split()` on this, is there a reason not to return a tuple?

This saves you the splitting, and removes the need to to base64-encode (and then decode) the signature.

autopush/jwt.py
jrconlin: It's in utils, which includes jwt.py, leading to a dependency loop.

alex: Work around that by moving the function here and changing that file to import it?

autopush/jwt.py
jrconlin: absolutely! Thanks!

alex:
Usage of the cryptography API here LGTM. I don't know anything else about the autopush codebase to say further :-)

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        # to be doing lazy verification and matching strings rather than
+        # comparing content values. If the signatures start failing for
+        # some unknown reason in the future, decode the signature and
+        # make sure it matches how we're reconstructing it.
+        verifier = pkey.verifier(
+            base64.urlsafe_b64decode(repad(signature.encode())),
+            ec.ECDSA(hashes.SHA256()))
+        verifier.update(sig_material.encode())
+        # This will raise an InvalidSignature exception if failure.
+        # It will be captured externally.
```
pjenvey: we're not catching it yet --

I think we can overhaul the JOSEError and AssertionError except clause in webpush now for InvalidSignature (IIRC ecdsa had all those asserts we had to catch?)

This func can definitely trigger ValueErrors (from_encoded_point, maybe the rsplits?) which is ok I guess, we're already catching them in that clause.

jose attempts to catch some ValueErrors and convert them to JOSEErrors -- but maybe it wasn't perfect at it because we still had to catch them. 

Looking at jose.jws._load, it's also careful about TypeErrors/binascii.Errors, might we trigger any of those too?

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        # uses namedtypes, with the names set to ""
+        #
+        # Oh frabjous day!
+        ds = univ.Sequence(
+            componentType=namedtype.NamedTypes()
+        ).setComponents(
+            univ.Integer(int(binascii.hexlify(sig[:32]), 16)),
+            univ.Integer(int(binascii.hexlify(sig[32:]), 16)),
+        )
+        encoded = encode(ds)
```
pjenvey: Could this ever raise a PyAsn1Error?

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            key
+        ).public_key(default_backend())
+        # eventually this would be a map of algs to signing algorithms.
+        # NOTE: verifier() will take any string as the signature. It appears
+        # to be doing lazy verification and matching strings rather than
+        # comparing content values. If the signatures start failing for
+        # some unknown reason in the future, decode the signature and
+        # make sure it matches how we're reconstructing it.
+        verifier = pkey.verifier(
+            base64.urlsafe_b64decode(repad(signature.encode())),
```
pjenvey: do you really want encode() w/ default encoding here and line 97? Should explicitly specify ascii then

autopush/jwt.py
```diff
@@ -0,0 +1,84 @@
+"""Why hand roll?
```
bbangert: This will show up as the module's title in the API docs.... oh, maybe we should add a .rst so that there'll be docs generated for this module?

autopush/jwt.py
```diff
@@ -0,0 +1,84 @@
...
+After spending half a day looking for reasonable replacements, I decided to
+just write the functions we need directly.
+
+THIS IS NOT A FULL JWT REPLACEMENT.
+
+"""
+
+import base64
+import binascii
+import json
```
bbangert: nit: line-break between module imports from std lib to 3rd party.

autopush/jwt.py
jrconlin: Partly. Mostly I wanted to keep the basic concept the same so I didn't have to go do a massive code rewrite.

autopush/web/webpush.py
```diff
@@ -321,7 +321,7 @@ def validate_auth(self, d):
 
         try:
             jwt = extract_jwt(token, public_key)
-        except (AssertionError, ValueError, JOSEError):
+        except (ValueError, InvalidSignature, Exception):
```
pjenvey: Exception here definitely not needed

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            return payload, sig
+
+        encoded = utils.encode_dss_signature(
+            s=int(binascii.hexlify(sig[32:]), 16),
+            r=int(binascii.hexlify(sig[:32]), 16)
+        )
+        return payload, encoded
+
+    @staticmethod
+    def decode(token, key=None, *args, **kwargs):
```
pjenvey: key should be required and don't need *args/**kwargs

autopush/jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            # This will raise an InvalidSignature exception if failure.
+            # It will be captured externally.
+            pkey.verify(
+                signature,
+                sig_material.encode('utf8'),
+                ec.ECDSA(hashes.SHA256()))
+            return json.loads(
+                base64.urlsafe_b64decode(
+                    repad(sig_material.split('.')[1]).encode('utf8')))
+        except Exception as ex:
```
pjenvey: It'd be nice to restrict this somewhat but it's a damn chore tracking all the potential exceptions. Maybe we should log totally unexpected ones to sentry?

I'll list the ones I know about, do with it what you want =]

splits: ValueError
base64: (TypeError, binascii.Error)
encode_dss_signature (underneath is pyasn stuff): not sure put probably that PyAsn1Error or whatever it was
verify: InvalidSignature
json.loads: ValueError

alex: `encode_dss_signature` shouldn't ever throw

alex: (if it is, let us know, we need to either fix a bug or update teh docs: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/utils/#cryptography.hazmat.primitives.asymmetric.utils.encode_dss_signature)

jrconlin:
Crap, missed pulling that test. Good catch, thanks!

autopush/jwt.py
jrconlin: @alex, encode_dss_signature shouldn't throw, but hexlify absolutely will. Granted, it's caught by the parent function, but it's also still in the critical path.

autopush/jwt.py
jrconlin: yeah, I had those in because I was trying to make the function a drop in replacement and was being paranoid.

autopush/jwt.py
jrconlin: @pjenvey I like your idea about having a end Exception that reports to sentry. That will let us spot errors, but still have the code behave properly.

autopush/jwt.py
```diff
@@ -0,0 +1,108 @@
...
+                sig_material.encode('utf8'),
+                ec.ECDSA(hashes.SHA256()))
+            return json.loads(
+                base64.urlsafe_b64decode(
+                    repad(sig_material.split('.')[1]).encode('utf8')))
+        except (ValueError, TypeError, binascii.Error, PyAsn1Error,
+                InvalidSignature):
+            raise InvalidSignature()
+        except Exception as ex:  # pragma: no cover
+            Logger().error("Unexpected error processing JWT: ", repr(ex))
```
pjenvey: you want failure("Unexpected error processing JWT"), that'll grab exc_info and send to sentry

autopush/jwt.py
```diff
@@ -0,0 +1,108 @@
...
+            # It will be captured externally.
+            pkey.verify(
+                signature,
+                sig_material.encode('utf8'),
+                ec.ECDSA(hashes.SHA256()))
+            return json.loads(
+                base64.urlsafe_b64decode(
+                    repad(sig_material.split('.')[1]).encode('utf8')))
+        except (ValueError, TypeError, binascii.Error, PyAsn1Error,
+                InvalidSignature):
```
pjenvey: nitpick don't need InvalidSignature here

pjenvey: Ah, probably don't need PyAsn1Error either according to alex

pjenvey:
One more related thing (I notice as I'm looking at sentry):

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/383686/

let's start catching TypeError/binascii.Error in extract_jwt's base64 call

autopush/jwt.py
```diff
@@ -0,0 +1,107 @@
...
+        Cryptography, which uses OpenSSL, uses a DER sequence of (s, r).
+        This function converts the raw ecdsa to DER.
+
+        :param auth: A JWT authorization token.
+        :type auth: str
+
+        :return tuple containing the signature material and signature
+
+        """
+        payload, asig = auth.rsplit(".", 1)
```
pjenvey: Another thing: I think we can just auth.encode('utf-8') here to begin with: then the other 2 later encode calls aren't needed. jose.jws _load does similar. (then you can even utilize our utils.base64_urldecode)

jrconlin: yes on calling `.encode('utf8')` no on using `utils.base64_urldecode`. the latter would set up a circular dependency.

Issue #855, opened 2017-03-30
APNS is not using the log.failure() method


jrconlin:
We're leaking some valuable data from being logged because we're not calling `log.failure()` to capture full state and exceptions for some APNS calls.

PR #856, opened 2017-03-30
bug: Use log.failure() to better track exceptions in APNS

jrconlin:
fixes: #855

### DIFF
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -4,6 +4,7 @@ import uuid
 from hyper.http20.exceptions import ConnectionError, HTTP20Error
 from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
+from twisted.python.failure import Failure
 
 from autopush.exceptions import RouterException
 from autopush.router.apns2 import (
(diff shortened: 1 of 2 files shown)


autopush/router/apnsrouter.py
```diff
@@ -146,18 +147,22 @@ def _route(self, notification, router_data):
         try:
             apns_client.send(router_token=router_token, payload=payload,
                              apns_id=apns_id)
-        except ConnectionError:
+        except ConnectionError as ex:
             self.ap_settings.metrics.increment(
                 "updates.client.bridge.apns.connection_err",
                 self._base_tags
             )
+            self.log.error("Connection Error sending to APNS",
```
bbangert: afaik using log.failure will automatically include any current exceptions.

bbangert:
Kind of confused that the commit is "Use log.failure" and then this doesn't use log.failure.

autopush/router/apnsrouter.py
pjenvey: yea we talked about this yesterday. As horrible as the log_failure=Failure() syntax is, twisted logging recommends typically not using failure (it's more along the lines of a critical, not handled well log)

jrconlin: pjenvey and I discussed that initially in channel. Not sure that this qualifies as a failure, since it's not killing the process, just invoking an error for a transaction.

jrconlin:
Fair point. It did originally. I'll change the commit message.

autopush/router/apnsrouter.py
pjenvey: ok there's 2 options to improve that:

log.failure("..", level=LogLevel.error) # sigh
log.error("..", log_failure=Failure()) # Failure grabs it from exc_info()

Issue #695, opened 2016-10-12
further refine Handler validation


pjenvey:
Improvements we can do here:

o how we pass validation results to handlers. @threaded_validate currently attaches valid_input to the handler obj yet continues calling the handler method w/ unvalidated kwargs (which we always ignore and are thus confusing/dangerous to even keep around).

We should probably pass **valid_input as the kwargs instead

o We can also pass thru richer objects (like a fully constructed WebPushNotification object for e.g. in web/message.py). Either as fields.Raw or in a @post_load (not clear to me yet what works better)?

jrconlin:
I'd prefer if we used `fields.Raw`. passing back as part of `@post_load` feels too "magical" for my tastes. We declare the fields at the top of the class and should stick to that so we don't lose data.

pjenvey:
While shaking this out it seemed like @post_load is maybe more appropriate (webpush.py already uses it). fields.Raw would require the @post_load type code to live inside of the @validates_schema methods most of the time (and we already have some larger ones).

Almost tempted to document what the schema produces to the handler in schema class docstring to make it very clear

PR #859, opened 2017-04-04
refactor: cleanup some deferred arg bookkeeping

pjenvey:
prepping some cleanup before the actual #695 handler change


autopush/web/registration.py
```diff
@@ -145,45 +149,43 @@ def post(self, *args, **kwargs):
...
-                          uri=self.request.uri)
-            d.addCallback(self._save_router_data,
-                          self.valid_input["router_type"])
-            d.addCallback(self._create_endpoint)
-            d.addCallback(self._return_endpoint, new_uaid, router)
+
+        if not uaid:
+            uaid = uuid.uuid4()
+            d = defer.execute(
+                router.register,
```
jrconlin: We are inconsistent with how we're calling router.register (sometimes via named args, sometimes via ordered.) We had gone to named args prior because ordered had lead to some sleeper bugs. I would recommend we stay with that practice so that future coders don't hit the same problems.

PR #863, opened 2017-04-07
bug: APNS may close a socket prematurely, resulting in an AttributeError

### LINKED ISSUE #862: Sudden socket closure may result in AttributeError

jrconlin:
closes #862

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -124,6 +124,9 @@ class APNSClient(object):
             # re-established.
             stream_id = connection.request(
                 'POST', url=url, body=body, headers=headers)
+            # get_response() may return an AttributeError. Not really sure
+            # how it happens, but the connected socket may get set to None.
+            # We'll treat that as a premature socket closure.
             response = connection.get_response(stream_id)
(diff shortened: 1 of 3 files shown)


pjenvey:
Wanna just upgrade to the latest hyper with its fix instead? Not a whole lot of difference in 0.6.2 vs 0.7.0 and it gets ping support which we were interested in anyway

(it's about time to upgrade our dependencies anyway)

jrconlin:
I'm fine doing both. Update hyper and leave the attribute trap in place.

jrconlin:
pjenvey: I was tracking https://github.com/Lukasa/hyper/issues/272, but I agree that https://github.com/Lukasa/hyper/commit/4d9262444b43b7b76cf649d853fc04b834d63545 should resolve the root issue. We can remove the AttributeError check in the future.

PR #864, opened 2017-04-07
fix: Hyphenate `content-available`; don't send `alert` for APNs

kitcambridge:
The [docs](https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#//apple_ref/doc/uid/TP40008194-CH10-SW8) suggests `content-available` is hyphenated, not underscored, and says we mustn't include `alert`, `sound`, or `badge` if we're sending a silent notification with `content-available`.


kitcambridge:
[`mutable-content`](https://developer.apple.com/reference/usernotifications/unnotificationserviceextension) also looks interesting. But it looks like that still takes an `alert` dict (`title` and `body`; only the *contents* of those fields are encrypted)...whereas `content-available` lets you send any kind of payload you want.

kitcambridge:
(╯°□°）╯︵ ┻━┻uoᴉʇɐʇuǝɯnɔop

kitcambridge:
Although, reading [more about `mutable-content`](https://developer.apple.com/reference/usernotifications/unnotificationserviceextension/1648229-didreceive)...

> You can modify any of the content from the original request. You might customize the content for the current user or replace it altogether.

And:

> You may also modify the alert text as long as you do not remove it. If the content object does not contain any alert text, the system ignores your modifications and delivers the original notification content.

So looks like it's OK to include a fake title and contents and completely replace them, as long as the app shows something?

jrconlin:
Yeah, the wording on that is really vague. Does
` If the content object does not contain any alert text, the system ignores your modifications and delivers the original notification content.` mean that there must be alert text in the content for any modification to be accepted, or that if you add alert content later, the modifications are also accepted. 

I'm fine not including it since it's trivial to add it back later. I get the feeling that this may be a point of experimentation.

PR #869, opened 2017-04-13
bug: enforce `senderID` for gcm/fcm

### LINKED ISSUE #868: Enforce senderID requirement for GCM/FCM records
We've seen a few reports of *cm records that do not include a `senderID` in their router data.
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/358097/events/9996820/
We should consider those records to be a critical failure and mark them appropriately.

NOTE: This may disable notifications to this client, but they weren't getting through anyway.

jrconlin:
closes #868

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -78,6 +78,14 @@ class WebPushSubscriptionSchema(Schema):
         if result.get("router_type") not in ["webpush", "gcm", "apns", "fcm"]:
             raise InvalidRequest("Wrong URL for user", errno=108)
 
+        if (result.get("router_type") in ["gcm", "fcm"]
+            and 'senderID' not in result.get('router_data',
+                                             {}).get("creds", {})):
+                # Make sure we note that this record is bad.
+                result['critical_failure'] = \
+                    result.get('critical_failure', "Missing SenderID")
+                settings.router.register_user(result)
+
         if result.get("critical_failure"):
             raise InvalidRequest("Critical Failure: %s" %
                                  result.get("critical_failure"),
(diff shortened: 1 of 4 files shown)


autopush/web/webpush.py
```diff
@@ -78,6 +78,14 @@ def validate_uaid_month_and_chid(self, d):
         if result.get("router_type") not in ["webpush", "gcm", "apns", "fcm"]:
             raise InvalidRequest("Wrong URL for user", errno=108)
 
+        if result.get("router_type") in ["gcm", "fcm"]:
+            if 'senderID' not in result.get('router_data',
```
bbangert: Why not combine these with an and?

bbangert:
So, the sentry error *does* have a senderID, this PR is intentionally not going to address that, and this is a separate issue, right?

autopush/web/webpush.py
jrconlin: because I broke them apart when I was fixing the tests and forgot to glue them back together?

me === idiot.

jrconlin:
following up here from irc. 

The "senderID" appearing in the sentry record is manufactured by the code. It's part of a failure trap so that the proper error return is triggered. This PR checks much earlier and also flags the record as having a critical failure. This will also let the publisher get a 410. 

I'm leaving the original trigger code in place because 1) I don't want to impact more code during a refactor than I absolutely need to and 2) the router_data comes from a DB call which is encapsulated (while it *should* run through the PR, there may be an edge case where data comes through without it and the resulting 500 would be worse)

jrconlin:
discussed in IRC, this will check and invalidate a record that is fetched from the db that may not have set a senderid. The SenderID that appeared in the message was an artifact.

PR #870, opened 2017-04-13
refactor: refine the Handler validation

### LINKED ISSUE #695: further refine Handler validation

pjenvey:
closes #695

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -80,18 +80,6 @@ class MessageTestCase(unittest.TestCase):
         d = self.finish_deferred = Deferred()
         self.message.finish = lambda: d.callback(True)
 
-    def _make_req(self, id=None, headers=None, body='',
-                  rargs=None, *args, **kwargs):
-        if headers is None:
-            headers = {}
-        self.request_mock.body = body
-        self.request_mock.headers.update(headers)
-        self.message.path_kwargs = {}
-        self.message.path_args = rargs or args or []
-        if id is not None:
-            self.message.path_kwargs = {"message_id": id}
-        return dict()
-
     def test_delete_token_invalid(self):
         self.fernet_mock.configure_mock(**{
             "decrypt.side_effect": InvalidToken})
@@ -100,7 +88,7 @@ class MessageTestCase(unittest.TestCase):
             self.status_mock.assert_called_with(400, reason=None)
         self.finish_deferred.addCallback(handle_finish)
 
-        self.message.delete(self._make_req(id=''))
+        self.message.delete(message_id='')
         return self.finish_deferred
 
     def test_delete_token_wrong_components(self):
@@ -110,7 +98,7 @@ class MessageTestCase(unittest.TestCase):
             self.status_mock.assert_called_with(400, reason=None)
         self.finish_deferred.addCallback(handle_finish)
 
-        self.message.delete(self._make_req("ignored"))
+        self.message.delete(message_id="ignored")
         return self.finish_deferred
 
     def test_delete_token_wrong_kind(self):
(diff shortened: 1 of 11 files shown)


autopush/tests/test_web_validation.py
```diff
@@ -158,13 +158,13 @@ def _make_fut(self):
         schema.context["log"] = Mock()
         return schema
 
-    def _make_test_data(self, headers=None, body="", path_args=None,
-                        path_kwargs=None, arguments=None):
+    def _make_test_data(self, headers=None, body="", args=None, kwargs=None,
+                        arguments=None):
```
jrconlin: Hrm. While it feels like a nit, `args` vs. `arguments` might be a bit confusing. Granted, `path_args` wasn't terribly great either, but at least it reflected what it was used for. 

Since it's kinda/sorta unused. I suppose we could just drop the `args` for now.

autopush/web/registration.py
```diff
@@ -133,6 +137,41 @@ def validate_data(self, data):
                                          errno=109,
                                          headers=request_pref_header)
 
+    @post_load
+    def handler_kwargs(self, data):
+        # not used
```
jrconlin: ??

autopush/tests/test_web_validation.py
pjenvey: I had the same thought and forgot all about it, I was thinking of renaming arguments (how about request_arguments?) it's rarely used anyway

autopush/web/registration.py
pjenvey: this just pops it off the handler kwargs after it's validated (not used = not used by the handlers)

autopush/tests/test_web_validation.py
pjenvey: eh, maybe still a bit confusing..

jrconlin: Yeah, like I said, a bit more nitty than needed. I'm also fine with just dropping "args" since I don't think that anything actually passes them. Calling "arguments" `request_arguments` might be good. Granted, shortening down 'kwargs' and tossing it into the middle of the arg list might cause a few python folk to twitch, but none of that is super critical.

pjenvey: Punting on this for now, I'll revisit it later. cyclone apparently sets up path_args/kwargs for use by the prepare method, so I guess there's a small case to keep the naming, but I wouldn't mind shortening. the websocket handlers still use regular args (though they don't validate), should probably switch those over

autopush/tests/test_endpoint.py
```diff
@@ -246,17 +219,24 @@ def setUp(self):
...
-        if headers is None:
-            headers = {}
-        self.request_mock.body = body or self.request_mock.body
-        self.request_mock.headers.update(headers)
-        self.reg.path_kwargs = {"router_type": router_type,
-                                "router_token": router_token,
-                                "uaid": uaid,
-                                "chid": chid}
-        return dict()
+    def _req(self, m, router_type="", router_token="", uaid=None, chid=None):
```
bbangert: m?

autopush/web/registration.py
```diff
@@ -133,6 +137,41 @@ def validate_data(self, data):
...
+            token=data.pop('router_token'),
+            data=data.pop('router_data')
+        )
+
+
+@attrs(slots=True)
+class RouterInfo(object):
+    """Bundle of Router registration information"""
+
+    router = attrib()  # type: Any
```
bbangert: It's really Any? I thought there were only a few valid sets of things it could be.

pjenvey: It's IRouter, however, mypy doesn't yet understand zope.interface

PR #876, opened 2017-04-19
feat: add tracking for content-encoding

### LINKED ISSUE #875: Add metrics for encryption schemes
Martin asked about this in https://bugzilla.mozilla.org/show_bug.cgi?id=1345665#c6, so that we'll know when it's safe to remove "aesgcm128" and "aesgcm". It seems like it'd be easier to record these on the server, if we don't already.

jrconlin:
closes #875

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -416,6 +416,9 @@ class WebPushHandler(BaseWebHandler):
                 self._client_info["jwt_" + i] = jwt["jwt_data"][i]
 
         user_data = subscription["user_data"]
+        encoding = ''
+        if notification.data and notification.headers:
+            encoding = notification.headers.get('encoding', '')
         self._client_info.update(
             message_id=notification.message_id,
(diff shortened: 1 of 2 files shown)


autopush/web/webpush.py
```diff
@@ -426,6 +426,10 @@ def post(self,
             version=notification.version
         )
 
+        if notification.data:
+            self.metrics.increment("updates.notification.encoding.{}".format(
+                notification.headers['encoding']))
```
pjenvey: isn't it headers['content-encoding']?

could use a test

jrconlin: Nope, it's 'encoding' once we finish schema processing. 
https://github.com/mozilla-services/autopush/blob/master/autopush/utils.py#L352

Will add the test.

bbangert: Nope, cleanup_headers changes it to 'encoding'.

Issue #867, opened 2017-04-12
Don't send empty string for required param


bbangert:
Per Sentry issue, https://sentry.prod.mozaws.net/operations/autopush-prod/issues/383827/

```
ValidationException: ValidationException: 400 Bad Request
{u'__type': u'com.amazon.coral.validate#ValidationException', u'message': u'ExpressionAttributeValues contains invalid value: One or more parameter values were invalid: An AttributeValue may not contain an empty string for key :router_data'}
```

Traceback:
```
ValidationException: ValidationException: 400 Bad Request
{u'__type': u'com.amazon.coral.validate#ValidationException', u'message': u'ExpressionAttributeValues contains invalid value: One or more parameter values were invalid: An AttributeValue may not contain an empty string for key :router_data'}
  File "twisted/python/threadpool.py", line 246, in inContext
    result = inContext.theWork()
  File "twisted/python/threadpool.py", line 262, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "autopush/db.py", line 292, in wrapper
    return func(self, *args, **kwargs)
  File "autopush/db.py", line 718, in register_user
    return_values="ALL_OLD",
  File "boto/dynamodb2/layer1.py", line 2765, in update_item
    body=json.dumps(params))
  File "boto/dynamodb2/layer1.py", line 2842, in make_request
    retry_handler=self._retry_handler)
  File "boto/connection.py", line 954, in _mexe
    status = retry_handler(response, i, next_sleep)
  File "boto/dynamodb2/layer1.py", line 2882, in _retry_handler
    response.status, response.reason, data)
```

pjenvey:
already fixed in #870, test added in #881

PR #881, opened 2017-04-24
fix: revert registration's strict critical failure check

### LINKED ISSUE #880: Revert registration's critical_failure UAID check
This is causing more trouble than it's worth for clients, e.g.:

https://bugzilla.mozilla.org/show_bug.cgi?id=1358689

Reregistering should clear out these errors but the critical_failure prevents them from doing so.

pjenvey:
was causing more trouble than it's worth for clients

also add an already passing test for #867 (fixed in b1312db)

fixes #880, #867

### DIFF
--- a/autopush/web/registration.py
+++ b/autopush/web/registration.py
@@ -59,20 +59,10 @@ class RegistrationSchema(Schema):
         chid = req['path_kwargs'].get('chid', router_data.get("channelID"))
         if uaid:
             try:
-                u_uuid = uuid.UUID(uaid)
+                uuid.UUID(uaid)
             except (ValueError, TypeError):
                 raise InvalidRequest("Invalid Request UAID",
                                      status_code=401, errno=109)
-            # Check if the UAID has a 'critical error' which means that it's
-            # probably invalid and should be reset/re-registered
-            try:
-                record = self.context['settings'].router.get_uaid(u_uuid.hex)
-                if record.get('critical_failure'):
-                    raise InvalidRequest("Invalid Request UAID",
-                                         status_code=410, errno=105)
-            except ItemNotFound:
-                pass
-
         if chid:
             try:
                 uuid.UUID(chid)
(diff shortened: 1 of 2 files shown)


bbangert:
Commit message should be 'Closes'

autopush/web/registration.py
```diff
@@ -59,20 +59,10 @@ def extract_data(self, req):
...
-                u_uuid = uuid.UUID(uaid)
+                uuid.UUID(uaid)
             except (ValueError, TypeError):
                 raise InvalidRequest("Invalid Request UAID",
                                      status_code=401, errno=109)
-            # Check if the UAID has a 'critical error' which means that it's
-            # probably invalid and should be reset/re-registered
-            try:
-                record = self.context['settings'].router.get_uaid(u_uuid.hex)
-                if record.get('critical_failure'):
```
bbangert: If it's a critical failure, what is the appropriate behavior? I'm worried about the reasoning for these changes being lost since it happened purely in an IRC chat afaik.

jrconlin:
The problem occurred because there were some records that were recorded into our database that did not have proper senderIDs specified. This was due partly to the mixup regarding stage vs. prod senderids and some very early devices not specifying the senderid they required. This was resolved, and many of the effected devices cycled out, but there are still occasional records that occur, leading to Sentry errors. 

The only work around is to completely uninstall firefox from the remote device, and then reinstall so that the proper configuration can be fetched and stored.

This is a "critical failure" because there is no way for us to communicate to these clients. The original fix attempted to resolve this problem by invoking a force reset when a device attempted to register a new channel. Unfortunately, that code path on the client does not support force resets at this time, causing the device to simply try again with the invalid data. That is why this patch was considered "not worth the effort" and the revert was recommended.

bbangert:
Thanks. So in this case we will accept it, even though its doomed, to avoid having the client continue to retry, yes?

jrconlin:
Yes. The number of impacted clients is small, and diminishing, so we may occasionally see Sentry errors about this, but it's not something that can be addressed with the current systems interactions.

Issue #882, opened 2017-04-26
Use consistent log field names and avoid mixing types


bbangert:
For logging, if its the same info, it should have the same field name everywhere. Otherwise it's easy to exclude it from our parquet files accidentally. I have no strong opinion on which name is used in any of these cases, but they need a consistent name.

Here are the ones I've identified thus far:
- channel_id
- channelID
- uaid (https://github.com/mozilla-services/autopush/blob/master/autopush/web/webpush.py#L419)
- uaid_hash (https://github.com/mozilla-services/autopush/blob/master/autopush/web/webpush.py#L447)
- ttl (appears to be the same as below, but logged in a different spot)
- message_ttl

For mixed types, we use an empty string when no status or logged_status is present which can mess up analysis. We should use a known 'null' value like 0 instead of "" when logging numbers and want to log a lack of one.

bbangert:
Closed by #883.

PR #883, opened 2017-04-27
feat: add parquet unified schema and cleanup logging messages

### LINKED ISSUE #874: Create single parquet schema for autopush
The PR here (#832) should be modified and re-opened by our team with a single schema. Starting from the code in this closed PR is a good start, and instead of separate schemas for each index as the PR has, a single schema including all the fields we want to capture should be made.

We could do this by starting from the more thorough of the existing schemas, and just ensure it has all the optional fields from both places.

bbangert:
Adds a unified parquet stream for all log messages with a message
matcher for necessary endpoint and connection node messages.
Logging fields were cleaned up for consistency and status code
logging uses a 0 instead of "" to avoid mixing types.

Closes #874, #882

### DIFF
--- a/parquet_schemas/autopush/push.cfg
+++ b/parquet_schemas/autopush/push.cfg
@@ -1,0 +1,46 @@
+-- vim:ft=lua
+message_matcher = message_matcher .. " && (Fields[Fields.message] == 'Successful delivery' || Fields[Fields.message] == 'Router miss, message stored.' || Fields[Fields.message] == 'Request timings'"
+message_matcher = message_matcher .. " || Fields[Fields.message] == 'hello' || Fields[Fields.message] == 'Register' || Fields[Fields.message] == 'Unregister' || Fields[Fields.message] == 'Ack' || Fields[Fields.message] == 'Nack')"
+
+s3_path_dimensions = {
+	{name="log", source="Type"},
+	{name="date", source="Fields[Date]"},
+	{name="hour", source="Fields[Hour]"},
+}
+
+parquet_schema = [=[
+message Log {
+    required int64 Timestamp;
+    optional binary Type (UTF8);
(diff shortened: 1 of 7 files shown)


parquet_schemas/autopush/push.cfg
```diff
@@ -0,0 +1,46 @@
...
+    optional int64 fields_request_time;
+    optional int64 fields_route_time;
+    optional int64 fields_validation_time;
+    optional binary fields_router_key (UTF8);
+    optional int64 fields_status_code;
+    optional binary fields_ua_browser_family (UTF8);
+    optional binary fields_ua_browser_ver (UTF8);
+    optional binary fields_ua_os_family (UTF8);
+    optional binary fields_ua_os_ver (UTF8);
+    optional binary fields_system (UTF8);
```
jrconlin: Didn't see `errno`,
(also didn't see `uaid_record`, ' but guessing that's intentional)
Should we differentiate better between `code` and `status_code`?

Also, do you want to fix up autopush.base.BaseHandler._init_info()? 
(I'm guessing we want to set some of those to integer values)

Not sure if this would screw anything up, but sorting these was really useful for spotting what's missing.

pjenvey: system was also in relud's but it looks pretty useless.

seems to be twisted logger's "log_system". typically our logs get a default value of "-" otherwise an occasional random class logPrefix like "HTTP11ClientProtocol,client", "HTTPConnection,0,127.0.0.1"

bbangert: I tweaked the base_info a little to avoid populating the TTL with a "", otherwise it looks fine.

bbangert: Also added errno.

PR #885, opened 2017-05-02
refactor: rearrange main around app (MultiService) objects

### LINKED ISSUE #465: Refactor main.py into well structured server setup and separate CLI parser

pjenvey:
- drive apps from settings as much as possible
- push some initialization into customized factories
- split arg parsing out into its own module

fixes #465

### DIFF
--- a/autopush/main_argparse.py
+++ b/autopush/main_argparse.py
@@ -1,0 +1,258 @@
+"""autopush/autoendpoint/etc script command line parsing"""
+import configargparse
+
+
+def add_shared_args(parser):
+    """Add's a large common set of shared arguments"""
+    parser.add_argument('--config-shared',
+                        help="Common configuration file path",
+                        dest='config_file', is_config_file=True)
+    parser.add_argument('--debug', help='Debug Info.', action="store_true",
+                        default=False, env_var="DEBUG")
+    parser.add_argument('--log_level', help='Log level to log', type=str,
+                        default="", env_var="LOG_LEVEL")
+    parser.add_argument('--log_output', help="Log output, stdout or filename",
+                        default="stdout", env_var="LOG_OUTPUT")
+    parser.add_argument('--firehose_stream_name', help="Firehose Delivery"
+                        " Stream Name", default="", env_var="STREAM_NAME",
+                        type=str)
+    parser.add_argument('--crypto_key', help="Crypto key for tokens",
+                        default=[], env_var="CRYPTO_KEY", type=str,
+                        action="append")
+    parser.add_argument('--key_hash', help="Key to hash IDs for storage",
+                        default="", env_var="KEY_HASH", type=str)
+    parser.add_argument('--datadog_api_key', help="DataDog API Key", type=str,
+                        default="", env_var="DATADOG_API_KEY")
+    parser.add_argument('--datadog_app_key', help="DataDog App Key", type=str,
+                        default="", env_var="DATADOG_APP_KEY")
+    parser.add_argument('--datadog_flush_interval',
+                        help="DataDog Flush Interval", type=int,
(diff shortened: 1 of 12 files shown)


autopush/tests/test_integration.py
```diff
@@ -354,135 +348,75 @@ def wait_for(self, func):
...
-            skip_request_logging
-        )
-        from autopush.settings import AutopushSettings
-        from autopush.web.message import MessageHandler
-        from autopush.web.registration import RegistrationHandler
-        from autopush.web.simplepush import SimplePushHandler
-        from autopush.web.webpush import WebPushHandler
+    endpoint_port = 9020
+
+    _endpoint_defaults = dict(
```
jrconlin: 👍 I like gathering the defaults like this. Nice and readable.

autopush/tests/test_integration.py
```diff
@@ -706,7 +638,7 @@ def test_basic_last_connect(self):
         yield client.disconnect()
         times = 0
         while times < 10:
-            c = yield deferToThread(self._settings.router.get_uaid,
+            c = yield deferToThread(self.conn.settings.router.get_uaid,
```
jrconlin: nit: Should we indent the same way we do with l#624 above?

autopush/main_argparse.py
```diff
@@ -0,0 +1,258 @@
...
+    # Apple Push Notification system (APNs) for iOS
+    # credentials consist of JSON struct containing a channel type
+    # followed by the settings,
+    # e.g. {'firefox':{'cert': 'path.cert', 'key': 'path.key',
+    #                  'sandbox': false}, ... }
+    parser.add_argument('--apns_creds', help="JSON dictionary of "
+                                             "APNS settings",
+                        type=str, default="",
+                        env_var="APNS_CREDS")
+    # UDP
```
pjenvey: I could nuke it in this PR but I'd prefer logging a separate issue for this. This change is so large that I'd rather play it safe to keep the main -> main_args delta as straightforward as possible

autopush/tests/test_integration.py
```diff
@@ -354,135 +348,75 @@ def wait_for(self, func):
                 break
 
 
+ROUTER_TABLE = os.environ.get("ROUTER_TABLE", "router_int_test")
+STORAGE_TABLE = os.environ.get("STORAGE_TABLE", "storage_int_test")
+MESSAGE_TABLE = os.environ.get("MESSAGE_TABLE", "message_int_test")
```
pjenvey: Great question but these were already here before this change

autopush/settings.py
```diff
@@ -163,9 +178,28 @@ def __init__(self,
             self.endpoint_hostname,
             endpoint_port
         )
+
+        # not accurate under autoendpoint
```
pjenvey: Sorry, this is basically a TODO since I'll be touching Settings next. Some settings are specific to endpoint or connection, e.g. "router_url" -- only used in websocket and produces a bogus value under the endpoint. I'm adding ws_url in the same manner, for now.

I will be cleaning that up -- pretty tempted to separate the Settings per endpoint/connection to follow suit w/ the App objects (test_integration's already half way there)

autopush/main_argparse.py
jrconlin: agreed. filed https://github.com/mozilla-services/autopush/issues/886

autopush/tests/test_integration.py
jrconlin: Ah, indeed, didn't see them further down.

PR #891, opened 2017-05-09
feat: add session log summary on connection close

### LINKED ISSUE #448: Add KPI Metrics logging statement
A new metric statement should be added when the client drops its connection, that logs some session statistics for the user-session.

Example logging message:

``` python
self.log.info("Session", connection_type="websocket", messages=3, register=0,
              unregister=0, new_uaid=False, uaid_hash="...", user-agent="..."
              )
```

This is needed for the following KPI's:
1. Daily Activations - Knowing new_uaid for a session indicates if the uaid is a reconnect or new.
2. DIU’s - Daily Inactive Users, to know one, I'd need to see that for all the users sessions for a day, there were no messages.

It's possible to get both of these by alternatively logging when a new uaid is issued, and logging 'hello' for a uaid_hash, which would let me group them to see there were no Ack's. That's likely just as efficient, but correlating commands issued per 'session', and other per-session stats would be hard to evaluate. As such, it'd be easier to do analysis on a Session logging statement that had it all.

A side-effect of this is that long-lived connections get a bit.... weird. This shouldn't be that bad for our stats, as the max typical life span of a connection right now is around 50 minutes.

bbangert:
Logs a summary of session statistics when a client drops its client
connection. Also cleans up several places where the PushState had
its uaid set but not its uaid_obj or uaid_hash.

Closes #448

### DIFF
--- a/autopush/tests/test_websocket.py
+++ b/autopush/tests/test_websocket.py
@@ -59,6 +59,7 @@ dummy_headers = {
 dummy_chid = uuid.uuid4()
 dummy_chid_str = str(dummy_chid)
 dummy_uaid = uuid.uuid4()
+dummy_uaid_str = dummy_uaid.hex
 
 
 def dummy_notif(**kwargs):
@@ -340,7 +341,7 @@ class WebsocketTestCase(unittest.TestCase):
 
     def test_no_messagetype_after_hello(self):
         self._connect()
-        self.proto.ps.uaid = "asdf"
+        self.proto.ps.uaid = dummy_uaid_str
         self._send_message(dict(data="wassup"))
 
         def check_result(close_args):
@@ -353,7 +354,7 @@ class WebsocketTestCase(unittest.TestCase):
 
     def test_unknown_messagetype(self):
         self._connect()
-        self.proto.ps.uaid = "asdf"
+        self.proto.ps.uaid = dummy_uaid_str
         self._send_message(dict(messageType="wassup"))
 
         def check_result(close_args):
@@ -366,8 +367,8 @@ class WebsocketTestCase(unittest.TestCase):
 
     def test_close_with_cleanup(self):
         self._connect()
-        self.proto.ps.uaid = "asdf"
-        self.proto.ap_settings.clients["asdf"] = self.proto
+        self.proto.ps.uaid = dummy_uaid_str
+        self.proto.ap_settings.clients[dummy_uaid_str] = self.proto
 
         # Stick a mock on
         notif_mock = Mock()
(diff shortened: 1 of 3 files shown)


autopush/tests/test_websocket.py
```diff
@@ -59,6 +59,7 @@
 dummy_chid = uuid.uuid4()
 dummy_chid_str = str(dummy_chid)
 dummy_uaid = uuid.uuid4()
+dummy_uaid_str = uuid.uuid4().hex
```
jrconlin: Silly question, but why not convert the dummy_uaid like we do for the chid? (take the `dummy_uaid.hex` value)

If we ever need to really check the string value of dummy_uaid, might make it more consistent.

bbangert: Was mainly just following convention of the _str above it for chid.

autopush/tests/test_websocket.py
```diff
@@ -59,6 +59,7 @@
 dummy_chid = uuid.uuid4()
 dummy_chid_str = str(dummy_chid)
 dummy_uaid = uuid.uuid4()
+dummy_uaid_str = dummy_uaid.hex
```
jrconlin: 👍

autopush/websocket.py
```diff
@@ -139,6 +144,41 @@ def wrapper(self, *args, **kwargs):
...
+    ua_browser_family = attrib(default="")  # type: str
+    ua_browser_ver = attrib(default="")  # type: str
+    connection_time = attrib(default=0)  # type: int
+
+    # Usage data
+    direct_acked = attrib(default=0)  # type: int
+    direct_storage = attrib(default=0)  # type: int
+    stored_retrieved = attrib(default=0)  # type: int
+    stored_acked = attrib(default=0)  # type: int
+    total_messages = attrib(default=0)  # type: int
```
pjenvey: total_messages isn't used

autopush/websocket.py
```diff
@@ -179,6 +224,7 @@ class PushState(object):
         'direct_updates',
 
         'msg_limit',
+        '_reset_uaid',
         'reset_uaid',
```
pjenvey: methods/properties don't need __slot__ definitions (reset_uaid, uaid/_hash_obj, etc).

I've been meaning to remove remove pause/resume/stopProducing from this list..

autopush/websocket.py
```diff
@@ -257,8 +307,55 @@ def message(self):
...
 
+    @property
+    def reset_uaid(self):
+        # type: () -> bool
+        return self._reset_uaid
+
+    @reset_uaid.setter
+    def reset_uaid(self, value):
+        if value:
+            self._reset_uaid = True
```
pjenvey: kind of annoying to keep this and other things in 2 places, can PushState be solely responsible and insert it into the logging some other way?

bbangert: Not that I saw, due to how reset is retained.

autopush/websocket.py
```diff
@@ -803,6 +890,7 @@ def _verify_user_record(self):
         if ("record_version" not in record or
                 int(record["record_version"]) < USER_RECORD_VERSION):
             self.ps.reset_uaid = True
+            self.ps.stats.uaid_reset = True
```
pjenvey: This isn't needed

PR #892, opened 2017-05-11
feat: handle JSONResponse errors like provisioned errors

### LINKED ISSUE #744: JSONResponseError: 400 Bad Request part 2
Follow up to #703. *Any* db call can trigger these, so we need to handle it in more cases. Here's one triggered from a router:

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/369293/

We need a fix similar to https://github.com/mozilla-services/autopush/pull/718 covering other db calls

bbangert:
When handling provisioned errors, the same logic should be used as when
a JSONResponseError occurs.

Closes #744

### DIFF
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -11,10 +11,8 @@ from StringIO import StringIO
 from typing import Any  # noqa
 
 import requests
-from boto.dynamodb2.exceptions import (
-    ItemNotFound,
-    ProvisionedThroughputExceededException,
-)
+from boto.dynamodb2.exceptions import ItemNotFound
+from boto.exception import JSONResponseError
 from twisted.internet.threads import deferToThread
 from twisted.internet.defer import (
     inlineCallbacks,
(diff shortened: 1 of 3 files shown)


autopush/web/base.py
```diff
@@ -236,7 +236,7 @@ def _overload_err(self, fail):
 
     def _boto_err(self, fail):
         """errBack for random boto exceptions"""
-        fail.trap(BotoServerError)
+        fail.trap(BotoServerError, JSONResponseError)
```
pjenvey: not needed: JSONResponse is a BotoServerError subclass

autopush/router/simple.py
```diff
@@ -107,8 +108,8 @@ def route_notification(self, notification, uaid_data):
             if result is False:
                 self.metrics.increment("router.broadcast.miss")
                 returnValue(self.stored_response(notification))
-        except ProvisionedThroughputExceededException:
-            raise RouterException("Provisioned throughput error",
+        except (ProvisionedThroughputExceededException, JSONResponseError):
```
pjenvey: turns out ProvisionedThroughputExceededException is a subclass of JSONResponseError: so you could just catch that one

bbangert:
Yup, as @pjenvey pointed out.

jrconlin:
Apparently it is.

PR #896, opened 2017-05-15
feat: temporarily allow aps data via registration message

### LINKED ISSUE #887: Allow ios registration messages to provide optional 'aps' configuration
In order to allow for faster bug handling, allow iOS registrations to specify the aps options as part of the new channel registration message.

jrconlin:
closes #887

### DIFF
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -133,7 +133,7 @@ class APNSRouter(object):
                 payload["cryptokey"] = notification.headers["crypto_key"]
             elif "encryption_key" in notification.headers:
                 payload["enckey"] = notification.headers["encryption_key"]
-            payload['aps'] = {"content-available": 1}
+            payload['aps'] = router_data.get('aps', {"content-available": 1})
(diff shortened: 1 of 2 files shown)


autopush/router/apnsrouter.py
```diff
@@ -133,7 +133,7 @@ def _route(self, notification, router_data):
                 payload["cryptokey"] = notification.headers["crypto_key"]
             elif "encryption_key" in notification.headers:
                 payload["enckey"] = notification.headers["encryption_key"]
-            payload['aps'] = {"content-available": 1}
+            payload['aps'] = router_data.get('aps', {"content-available": 1})
```
bbangert: If there's an aps entry, then content-available doesn't need to be present?

jrconlin: Yes.
There needs to be an "aps" block, thus why I default to `"{content-available": 1}`, however, 'content-available is equivalent to our silent notifications, so it can be problematic. I presume that the folks providing the override `aps` data are knowledgeable and allow for full customization.

PR #900, opened 2017-05-20
bug: Serialize Decimal correctly for ios aps

### LINKED ISSUE #899: Serialize Decimal values in iOS APNs output
See https://sentry.prod.mozaws.net/operations/autopush-dev/issues/605747/

jrconlin:
closes #899

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -1,5 +1,6 @@
 import json
 from collections import deque
+from decimal import Decimal
 
 import hyper.tls
 from hyper import HTTP20Connection
@@ -19,6 +20,16 @@ APNS_MAX_CONNECTIONS = 20
 # see https://developer.apple.com/search/?q=%22apns-priority%22
 APNS_PRIORITY_IMMEDIATE = '10'
 APNS_PRIORITY_LOW = '5'
+
+
+class ComplexEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, Decimal):
+            return int(obj.to_integral_value())
+        # for most data types, this function isn't called.
+        # the following is added for safety, but should not
+        # be required.
+        return json.JSONEncoder.default(self, obj)  # pragma nocover
 
 
 class APNSException(Exception):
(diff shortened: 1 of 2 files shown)


autopush/router/apns2.py
```diff
@@ -21,6 +22,16 @@
 APNS_PRIORITY_LOW = '5'
 
 
+class ComplexEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, Decimal):
+            return int(float(obj.to_integral()))
```
bbangert: Why not use to_integral_value? The docs indicate it should already return an int.

jrconlin: I'll blame pycharm a bit for this one. When I just used object, i was getting warnings. To be safe, i figured being explicit would be safer.

jrconlin: Going to switch this out for less juggling

jrconlin: Hrm, ok, so reading the docs and thinking about this, there are a few possible methods that I should take. While the outcome is "dump this out as an integer" and this is absolutely temporary while the iOS folk are poking at it, there's a few things at play. 

`to_integral` (or properly `to_integral_value`) does the rounding inside the decimal without flagging an exception. float converts the decimal to a numeric and int converts to the expected JSON serializable element. Having decimal do it's rounding is probably the most accurate thing. The other two steps are pretty much required.

jrconlin: And, this is why reading a code review comment while following a keynote is probably not a good idea. Misread your comment. *sigh*

autopush/router/apns2.py
```diff
@@ -21,6 +22,16 @@
 APNS_PRIORITY_LOW = '5'
 
 
+class ComplexEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, Decimal):
+            return int(float(obj.to_integral_value()))
```
pjenvey: should kill the float() here: to_integral_value already clips it

jrconlin: Ah, thanks!

PR #903, opened 2017-05-23
bug: Trap for missing Vapid AUTH header token

### LINKED ISSUE #902: Handle missing bearer token in VAPID auth
Passing just "Bearer" (or presumably any other valid Auth scheme word) causes an index error. This should return a 401.

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/605058/

jrconlin:
* Fix some flake8 spacing

closes #902

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -327,7 +327,8 @@ class WebPushRequestSchema(Schema):
             else:
                 public_key = d["subscription"].get("public_key")
             jwt = extract_jwt(token, public_key)
-        except (KeyError, ValueError, InvalidSignature, TypeError):
+        except (KeyError, ValueError, InvalidSignature, TypeError,
+                VapidAuthException):
             raise InvalidRequest("Invalid Authorization Header",
                                  status_code=401, errno=109,
                                  headers={"www-authenticate": PREF_SCHEME})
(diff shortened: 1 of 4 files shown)


autopush/web/webpush.py
```diff
@@ -327,7 +327,8 @@ def validate_auth(self, d):
             else:
                 public_key = d["subscription"].get("public_key")
             jwt = extract_jwt(token, public_key)
-        except (KeyError, ValueError, InvalidSignature, TypeError):
+        except (KeyError, ValueError, InvalidSignature, TypeError,
+                VapidAuthException):
```
pjenvey: unless I'm mistaken VapidAuthException doesn't seem necessary here

jrconlin: parse_auth_header at https://github.com/mozilla-services/autopush/pull/903/files/2fccec64b166c6487a598362b1242456e1846f4a#diff-ccbbd52ff880cf4ddb71b18d126d189cR321 can return a VapidAuthException now. (See https://github.com/mozilla-services/autopush/pull/903/files/2fccec64b166c6487a598362b1242456e1846f4a#diff-0457b7bca45dadd77752fbfd0ee3e74aR570)

Issue #871, opened 2017-04-13
refactor tests


pjenvey:
Our tests could be refined, some improvements along the lines of:

- inlineCallbacks everywhere (like test_integration)
- some higher level abstractions could remove a lot of mocks

Cyclone's master branch actually a test client along those lines, we could possibly utilize it (though it's a bit odd in it returns handlers from the request methods.. I suppose we could tweak it)

https://github.com/fiorix/cyclone/blob/master/cyclone/tests/test_testing.py
https://github.com/fiorix/cyclone/blob/master/cyclone/testing/client.py

PR #905, opened 2017-05-24
refactor: convert handler tests to client/inlineCallbacks

pjenvey:
and remove mypy's deprecated fast_parser option

issue #871

### DIFF
--- a/autopush/http.py
+++ b/autopush/http.py
@@ -92,7 +92,9 @@ class BaseHTTPFactory(cyclone.web.Application):
         thrown.
 
         """
-        for pattern, handler in cls.ap_handlers:
+        if 'handlers' in kwargs:  # pragma: nocover
+            raise ValueError("handler_cls incompatibile with handlers kwarg")
+        for pattern, handler in cls.ap_handlers + cls.health_ap_handlers:
             if handler is handler_cls:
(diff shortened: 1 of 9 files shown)


autopush/logging.py
```diff
@@ -60,15 +60,18 @@
...
 
     Registers via the global LogBeginner the first time called.
     """
     global began_logging
     if not began_logging:
-        globalLogBeginner.beginLoggingTo([observer], redirectStandardIO=False)
+        globalLogBeginner.beginLoggingTo(
+            [observer],
+            redirectStandardIO=redirectStandardIO,
+            **kwargs)
```
bbangert: Do we want to drop the ) when we immediately drop-line after starting, or keep it this way? I'm fine with either way, just want to know which one we should use everywhere.

pjenvey: we've standardized dropping it lately, I forgot it here

autopush/http.py
```diff
@@ -92,7 +92,9 @@ def for_handler(cls, handler_cls, *args, **kwargs):
         thrown.
 
         """
-        for pattern, handler in cls.ap_handlers:
+        if 'handlers' in kwargs:  # pragma: nocover
```
jrconlin: Not a stopper, but we can't test this?

PR #906, opened 2017-05-26
feat: avoid triggering validation errors from bridge API

### LINKED ISSUE #893: Avoid triggering validation errors in dynamodb
Fields that can't be empty are being left empty. See:
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/395245/
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/603910/
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/394586/

bbangert:
The existing bridge API was overly permissive on parameters and
difficult to follow on necessary values. As a result this became
a bit larger than originally anticiated and is a refactor of the
bridge API into distinct handlers for each distinct URL path with
re-usable and specific validation schema's.

Closes #893

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -5,7 +5,6 @@ import twisted.internet.base
 from cryptography.fernet import Fernet, InvalidToken
 from cyclone.web import Application
 from mock import Mock, patch
-from moto import mock_dynamodb2
 from nose.tools import eq_, ok_
 from twisted.internet.defer import inlineCallbacks
 from twisted.trial import unittest
@@ -31,21 +30,22 @@ from autopush.utils import (
     generate_hash,
 )
 from autopush.web.message import MessageHandler
-from autopush.web.registration import RegistrationHandler
+from autopush.web.registration import NewRegistrationHandler
 
-mock_dynamodb2 = mock_dynamodb2()
 dummy_uaid = uuid.UUID("abad1dea00000000aabbccdd00000000")
 dummy_chid = uuid.UUID("deadbeef00000000decafbad00000000")
 dummy_token = dummy_uaid.hex + ":" + str(dummy_chid)
 
 
 def setUp():
-    mock_dynamodb2.start()
+    from .test_integration import setUp
+    setUp()
     create_rotating_message_table()
 
 
 def tearDown():
-    mock_dynamodb2.stop()
+    from .test_integration import tearDown
+    tearDown()
 
 
 class FileConsumer(object):  # pragma: no cover
@@ -142,10 +142,8 @@ class MessageTestCase(unittest.TestCase):
         eq_(resp.get_status(), 503)
 
 
-CORS_HEAD = "GET,POST,PUT,DELETE"
-
-
 class RegistrationTestCase(unittest.TestCase):
+    CORS_HEAD = "POST"
 
     def setUp(self):
         twisted.internet.base.DelayedCall.debug = True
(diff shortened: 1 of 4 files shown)


autopush/tests/test_endpoint.py
```diff
@@ -649,7 +620,8 @@ def test_delete_uaid(self):
     @inlineCallbacks
     def test_delete_bad_uaid(self):
         resp = yield self.client.delete(
-            self.url(router_type="test", router_token="test", uaid="invalid"),
+            self.url(router_type="test", router_token="test",
+                     uaid=uuid.uuid4().hex),
```
jrconlin: tempted to suggest using a static UUID like `"0"*32`, or a comment that the generated, valid UUID is bad because there's no associated record. Might save some future confusion about why this returns a 401.

autopush/web/registration.py
```diff
@@ -34,242 +32,356 @@
...
-        return d
-
-    @threaded_validate(RegistrationSchema)
-    def put(self, rinfo, uaid=None, chid=None):
-        # type: (RouterInfo, Optional[uuid.UUID], Optional[str]) -> Deferred
-        """HTTP PUT
-
-        Update router type/data for a UAID.
+        uaid = uuid.uuid4()
+        chid = router_data.get("channelID", uuid.uuid4().hex)
```
jrconlin: Please add warning that the value must match what's specified by the bridged client. We were bit pretty hard by this one and I'd like to make sure it doesn't come back.

autopush/web/registration.py
```diff
@@ -34,242 +32,356 @@
...
-            # mark channel as dead
-            self.ap_settings.metrics.increment("updates.client.unregister",
-                                               tags=self.base_tags())
-            d = deferToThread(self._delete_channel, uaid, chid)
-            d.addCallback(self._success)
-            d.addErrback(self._chid_not_found_err)
-            d.addErrback(self._response_err)
-            return d
-        # nuke all records for the UAID
         d = deferToThread(self._delete_uaid, uaid)
```
jrconlin: We should make sure we return the right content-type here (see PUT and POST)

autopush/web/registration.py
bbangert: I don't understand what you mean.

jrconlin: Right now the clients use a hex formatted UUIDs for the CHID. We need to return values that match that format. Previously we would return str formatted UUIDs, which would cause significant problems for the client. I want to make sure that mistake is not made again in the future. 

It's an external dependency that is dictated by a customer.

autopush/web/registration.py
```diff
@@ -34,242 +32,356 @@
...
         d.addCallback(self._write_channels, uaid)
         d.addErrback(self._uaid_not_found_err)
         d.addErrback(self._response_err)
         return d
 
-    @threaded_validate(RegistrationSchema)
-    def delete(self, uaid=None, chid=None, **kwargs):
-        # type: (Optional[uuid.UUID], Optional[str], **Any) -> Deferred
+    @threaded_validate(TokenUpdateSchema)
+    def put(self, router_data, router_type, uaid):
```
pjenvey: swap these: ordered as router_type/router_data everywhere else

autopush/web/registration.py
pjenvey: The response rendering functions should really handle setting content type

autopush/web/registration.py
```diff
@@ -34,242 +32,356 @@
...
+        self.ap_settings.message.register_channel(uaid.hex, chid)
+        return self.ap_settings.make_endpoint(uaid.hex, chid, app_server_key)
 
-    #############################################################
-    #                    Cyclone HTTP Methods
-    #############################################################
-    @threaded_validate(RegistrationSchema)
-    def post(self, rinfo, uaid=None, chid=None):
-        # type: (RouterInfo, Optional[uuid.UUID], Optional[str]) -> Deferred
+    def _register_user_and_channel(self, uaid, chid, router_type, router_data):
```
pjenvey: only used in NewRegistrationHandler: it should really live there

autopush/tests/test_endpoint.py
```diff
@@ -446,37 +426,15 @@ def test_post_bad_params(self):
...
-    @inlineCallbacks
     def test_post_nochid(self, *args):
         self.patch('uuid.uuid4', return_value=dummy_chid)
 
         self.fernet_mock.configure_mock(**{
             'encrypt.return_value': 'abcd123',
         })
         resp = yield self.client.post(
-            self.url(router_type="simplepush", uaid=dummy_uaid.hex),
+            self.url(router_type="simplepush", uaid=dummy_uaid.hex) +
```
bbangert: We could do that when we pull them apart, not in a rush right now.

autopush/web/registration.py
bbangert: Removed due to using write with a dict.

autopush/tests/test_endpoint.py
jrconlin: 👍

PR #909, opened 2017-05-29
bug: Ensure 'aps' passes through marshmallow schema check

### LINKED ISSUE #908: Ensure APS data is passed through registration system unfiltered
A recent refactor strips 'aps' data from passing through the marshmalllow schema. Patch returns it and adjusts integration test to check.

jrconlin:
closes #908

### DIFF
--- a/autopush/web/registration.py
+++ b/autopush/web/registration.py
@@ -68,8 +68,11 @@ class SubInfoSchema(Schema):
 
 
 class TokenSchema(SubInfoSchema):
-    """Validates that token and optionally key/channelID"""
+    """Filters allowed values from body data"""
     token = fields.Str(allow_none=True)
+    # Temporarily allow 'aps' definition data for iOS.
+    # TODO: lock down dict content to just allowed extra values.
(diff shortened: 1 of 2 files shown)


autopush/web/registration.py
```diff
@@ -68,8 +68,11 @@ def convert_chid(self, data):
 
 
 class TokenSchema(SubInfoSchema):
-    """Validates that token and optionally key/channelID"""
+    """Filters allowed values from body data"""
     token = fields.Str(allow_none=True)
+    # TODO: Temporarily allow 'aps' definition data for iOS.
```
bbangert: Since this now allows 'aps' with this change, wouldn't the TODO be the locking down underneath?

jrconlin: Yes, exactly. We don't know the values that iOS will require, thus we can't lock them down right now. We have to allow everything. When iOS completes and we understand what we need to allow, then the `TODO` would be addressed.

bbangert: Ok, how about removing the comment entirely? Better to say nothing than mislead, no?

PR #914, opened 2017-06-10
refactor: convert to inlineCallbacks

pjenvey:
and cleanup NotificationHandler delete's args/regex

issue #871

### DIFF
--- a/autopush/http.py
+++ b/autopush/http.py
@@ -153,7 +153,7 @@ class InternalRouterHTTPFactory(BaseHTTPFactory):
 
     ap_handlers = (
         (r"/push/([^\/]+)", RouterHandler),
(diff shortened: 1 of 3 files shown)


autopush/http.py
```diff
@@ -153,7 +153,7 @@ class InternalRouterHTTPFactory(BaseHTTPFactory):
 
     ap_handlers = (
         (r"/push/([^\/]+)", RouterHandler),
-        (r"/notif/([^\/]+)(/([^\/]+))?", NotificationHandler),
+        (r"/notif/([^\/]+)(?:/([^\/]+))?", NotificationHandler),
```
bbangert: We're going to int the last arg, maybe only match on digits?

PR #929, opened 2017-06-20
bug: Return 400 if routing token is blank or unspecified

### LINKED ISSUE #921: Error out if routing token is blank
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/614025/

Routing tokens cannot be empty.

jrconlin:
closes #921

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -1865,6 +1865,23 @@ class TestGCMBridgeIntegration(IntegrationBase):
         eq_(ca_data['enc'], salt)
         eq_(ca_data['body'], base64url_encode(data))
 
+    @inlineCallbacks
+    def test_registration_no_token(self):
+        self._add_router()
+        # get the senderid
+        url = "{}/v1/{}/{}/registration".format(
+            self.ep.settings.endpoint_url,
+            "gcm",
+            self.senderID,
+        )
+        response, body = yield _agent('POST', url, body=json.dumps(
+            {
+                "chid": str(uuid.uuid4()),
+                "token": '',
+             }
+        ))
+        eq_(response.code, 400)
+
 
 class TestFCMBridgeIntegration(IntegrationBase):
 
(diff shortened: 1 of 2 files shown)


autopush/web/registration.py
```diff
@@ -71,7 +74,19 @@ def convert_chid(self, data):
 class TokenSchema(SubInfoSchema):
     """Filters allowed values from body data"""
     token = fields.Str(allow_none=True)
-    # Temporarily allow 'aps' definition data for iOS.
+
+
+def validate_token(val):
```
bbangert: Why not use the http://marshmallow.readthedocs.io/en/latest/api_reference.html#marshmallow.validate.Length validator to ensure there's something here?

jrconlin: I was more thinking about someone using a token like " " by accident, and the fact that we don't control tokens, so we don't really have any idea what would be considered a valid size. 

I suppose I could somewhat arbitrarily decide there has to be a minimum of 8 characters and hope someone doesn't send "                ".

bbangert: There's also a RegExp validator in there, if we have some idea what they could be. Is it possible for a token to be nested structure like a dict?

jrconlin: Huh, didn't know about the RegExp validator, might be useful. Tokens should be strings, but again, are opaque to us. (Granted, I don't see google or apple going to making these transparent, complex structures for a lot of reasons.)

autopush/tests/test_integration.py
```diff
@@ -1865,6 +1865,22 @@ def test_registration(self):
...
+    def test_registration_no_token(self):
+        self._add_router()
+        # get the senderid
+        url = "{}/v1/{}/{}/registration".format(
+            self.ep.settings.endpoint_url,
+            "gcm",
+            self.senderID,
+        )
+        response, body = yield _agent('POST', url, body=json.dumps(
+            {"chid": str(uuid.uuid4()),
```
pjenvey: nitpick odd whitepsace in this dict

PR #932, opened 2017-06-22
feat: accept aes128gcm content encoding

### LINKED ISSUE #930: Accept messages with Content-Encoding: aes128gcm
The latest encryption drafts specify that the Content-Encoding HTTP header for messages encrypted with this scheme needs to set to `aes128gcm`.

https://tools.ietf.org/html/draft-ietf-webpush-encryption
https://tools.ietf.org/html/draft-ietf-httpbis-encryption-encoding

When pushing a payload to autopush with this header set, I get the following error:

> Request did not validate Unknown Content-Encoding

Given that Nightly gained support for decrypting such payloads early May, it'd be nice if your server could be updated to accept them too.

I'm testing with the following page. Chrome supports the new scheme starting Chrome 60 (currently in beta):

https://tests.peter.sh/push-generator/#padding=0;;encryption=1

jrconlin:
closes #930 

NOTE: pjenvey: If this conflicts with your refactor, this has lower priority.

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -13,7 +13,7 @@ from marshmallow import (
     validates_schema,
 )
 from marshmallow_polyfield import PolyField
-from marshmallow.validate import OneOf
+from marshmallow.validate import Equal
 from twisted.logger import Logger  # noqa
 from twisted.internet.defer import Deferred  # noqa
 from twisted.internet.defer import maybeDeferred
@@ -170,7 +170,7 @@ class WebPushCrypto01HeaderSchema(Schema):
     content_encoding = fields.String(
         required=True,
         load_from="content-encoding",
-        validate=OneOf(["aesgcm128"])
+        validate=Equal("aesgcm128")
     )
     encryption = fields.String(required=True)
     encryption_key = fields.String(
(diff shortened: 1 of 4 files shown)


autopush/web/webpush.py
```diff
@@ -255,6 +255,40 @@ def reject_encryption_key(self, data, original_data):
...
+    """Validates WebPush Message Encryption
+
+    Uses draft-ietf-httpbis-encryption-encoding-06 rules for validation
+
+    """
+
+    content_encoding = fields.String(
+        required=True,
+        load_from="content-encoding",
+        validate=OneOf(["aes128gcm"])
```
pjenvey: nitpick: think I've mentioned this before but we always end up w/ OneOf([onething]), should just be Equals(onething)

bbangert:
If the commit doesn't use a capital c for 'closes', then clog doesn't see it. Which is unfortunately kind of sad, since our contributing doc makes it sound like all lowercase 'closes' is fine. This is why some of our changelog is missing Issue linkage for what it closed.

autopush/web/webpush.py
jrconlin: Fair point. I'll change the other ones that do the same thing.

Issue #632, opened 2016-08-25
Refactor object initializers


bbangert:
We have multiple, large, hideous object initializers. I.e. AutopushSettings, PushState, etc.

Refactor should use attrs:
https://glyph.twistedmatrix.com/2016/08/attrs.html
https://attrs.readthedocs.io/en/stable/index.html

pjenvey:
* make_settings should move to Settings.from_argparse, and hopefully some of the attr's validation can aid in cleaning it up
* Settings' constructor defaults should really sync up w/ argparse defaults in some fashion
* Likely rename AutopushSettings -> AutopushConfig because referring to it as "ap_settings" in handlers (to not clash w/ cyclone's "settings") is slightly annoying
* extra credit: separation of connection/endpoint settings

PR #933, opened 2017-06-23
refactor: begin tearing apart AutopushSettings

pjenvey:
- move db concerns (incl. metrics) into a simple DatabaseManager.
heavier db init's moved into its own setup() method that most tests
don't even call
- split out routers, now connection doesn't even create them
- kill PushState.msg_limit, nor did it need a metrics attrib

issue #632

### DIFF
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -1,9 +1,15 @@
 import sys
 import uuid
+from typing import TYPE_CHECKING
 
 import cyclone.web
 from twisted.logger import Logger
 from twisted.python import failure
+
+if TYPE_CHECKING:  # pragma: nocover
+    from autopush.db import DatabaseManager  # noqa
+    from autopush.metrics import IMetrics  # noqa
+    from autopush.settings import AutopushSettings  # noqa
 
 
 class BaseHandler(cyclone.web.RequestHandler):
@@ -11,10 +17,24 @@ class BaseHandler(cyclone.web.RequestHandler):
 
     log = Logger()
 
-    def initialize(self, ap_settings):
-        """Setup basic attributes from AutopushSettings"""
-        self.ap_settings = ap_settings
+    def initialize(self):
+        """Initialize info from the client"""
         self._client_info = self._init_info()
+
+    @property
+    def ap_settings(self):
+        # type: () -> AutopushSettings
+        return self.application.ap_settings
+
+    @property
+    def db(self):
+        # type: () -> DatabaseManager
+        return self.application.db
+
+    @property
+    def metrics(self):
+        # type: () -> IMetrics
+        return self.db.metrics
 
     def _init_info(self):
         return dict(
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -37,6 +37,11 @@ import time
 import uuid
 from functools import wraps
 
+from attr import (
+    attrs,
+    attrib,
+    Factory
+)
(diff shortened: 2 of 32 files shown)


autopush/router/__init__.py
```diff
@@ -12,3 +12,18 @@
 
 __all__ = ["APNSRouter", "FCMRouter", "GCMRouter", "SimpleRouter",
            "WebPushRouter"]
+
+
+def routers_from_settings(settings, db):
+    """Create a dict of IRouters for the given settings"""
+    router_conf = settings.router_conf
+    routers = dict(
+        simplepush=SimpleRouter(
```
pjenvey: Sure but not right now

autopush/db.py
```diff
@@ -853,3 +861,130 @@ def clear_node(self, item):
...
+        """Initializes a dict of the initial rotating messages tables.
+
+        An entry for last months table, an entry for this months table,
+        an entry for tomorrow, if tomorrow is a new month.
+
+        """
+        today = datetime.date.today()
+        last_month = get_rotating_message_table(self._message_prefix, -1)
+        this_month = get_rotating_message_table(self._message_prefix)
+        self.current_month = today.month
```
pjenvey: exactly what I'm trying to get at -- I actually did forget that Router/Storage are still being initialized upon creation here. I'll fix this later in another commit. DatabaseManager() creation should be light weight (no network/db calls), until you explicitly tell it to initialize

autopush/db.py
```diff
@@ -853,3 +861,130 @@ def clear_node(self, item):
...
+
+    @message.setter
+    def message(self, value):
+        """Setter to set the current message table"""
+        self.message_tables[self.current_msg_month] = value
+
+    def _tomorrow(self):
+        return datetime.date.today() + datetime.timedelta(days=1)
+
+    def create_initial_message_tables(self):
```
pjenvey: forgot some of these and we do have some circular imports. this is going to start getting typing.TYPE_CHECKING ugly

bbangert: Ah, ok.

autopush/db.py
bbangert: Sounds good.

autopush/base.py
```diff
@@ -1,21 +1,41 @@
 import sys
 import uuid
+from typing import TYPE_CHECKING
 
 import cyclone.web
 from twisted.logger import Logger
 from twisted.python import failure
 
+if TYPE_CHECKING:  # pragma: nocover
```
pjenvey: sweet

autopush/db.py
```diff
@@ -853,3 +861,132 @@ def clear_node(self, item):
...
+    metrics = attrib()  # type: IMetrics
+
+    message_tables = attrib(default=Factory(dict))  # type: Dict[str, Message]
+    current_msg_month = attrib(default=None)        # type: Optional[str]
+    current_month = attrib(default=None)            # type: Optional[int]
+
+    _message_prefix = attrib(default="message")  # type: str
+
+    @classmethod
+    def from_settings(cls, settings):
```
pjenvey: we get a light weight `__init__` automatically w/ @attrs (some tests use it). these might change around a bit though in a later commit, re the deferred initialization comments above

jrconlin: Ah! ok, kinda like how we have to have `ap_settings`.

PR #934, opened 2017-06-23
feat: log EC2 instance ID when using AWS for hostname

### LINKED ISSUE #931: Use internal EC2 instance ID for datadog Host value
By default the datadog Python agent is picking up the machines external hostname and using it as the host value. This causes a mismatch in the Host values in Datadog, that also doesn't match the logging infra Host (internal hostname).

This value can be retrieved on startup from http://169.254.169.254/latest/meta-data/instance-id

bbangert:
Closes #931

### DIFF
--- a/autopush/tests/test_utils.py
+++ b/autopush/tests/test_utils.py
@@ -1,5 +1,6 @@
 import unittest
 
+from mock import patch, Mock
 from nose.tools import eq_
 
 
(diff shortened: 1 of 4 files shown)


autopush/tests/test_utils.py
```diff
@@ -29,3 +30,21 @@ def test_other_os_and_browser(self):
         eq_(raw["ua_os_family"], "BlackBerry OS")
         eq_(dd["ua_browser_family"], "Other")
         eq_(raw["ua_browser_family"], "BlackBerry")
+
+    @patch("requests.get")
+    def test_get_ec2_instance_id_unknwon(self, request_mock):
```
jrconlin: spelling "unknwon"

bbangert: oh for....

PR #941, opened 2017-06-28
Use cryptography from git to fix a memory leak

alex:


bbangert:
Nit, our commit format uses angular style commit messages. ;)

alex:
@bbangert fixed.

PR #945, opened 2017-06-29
refactor: split clients and agent off settings

pjenvey:
issue #632

### DIFF
--- a/autopush/http.py
+++ b/autopush/http.py
@@ -10,6 +10,12 @@ from typing import (  # noqa
 )
 
 import cyclone.web
+from twisted.internet import reactor
+from twisted.web.client import (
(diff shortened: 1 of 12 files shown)


autopush/http.py
```diff
@@ -109,18 +112,21 @@ def for_handler(cls,
...
                 if db is None:
                     db = DatabaseManager.from_settings(ap_settings)
-                if routers is None:
-                    routers = routers_from_settings(ap_settings, db)
-                return cls(
+                return cls._for_handler(
                     ap_settings,
-                    db=db,
-                    routers=routers,
+                    db,
```
bbangert: Keep the keyword declaration for clarity?

PR #948, opened 2017-07-05
feat: Use modern metrics

### LINKED ISSUE #943: Update metrics calls for modern tracking
See https://docs.google.com/a/mozilla.com/spreadsheets/d/18RL-Brz-Q-jR-iClNPCwDMVPYS0P2CvkHSJ2WZK-ceA/edit?usp=sharing

jrconlin:
Closes #943 

(Metric updates from https://docs.google.com/spreadsheets/d/18RL-Brz-Q-jR-iClNPCwDMVPYS0P2CvkHSJ2WZK-ceA/edit#gid=0)

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -1,5 +1,5 @@
 """Metrics interface and implementations"""
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Sequence, Any  # noqa
 
 from twisted.internet import reactor
 from txstatsd.client import StatsDClientProtocol, TwistedStatsDClient
@@ -72,6 +72,14 @@ class TwistedMetrics(object):
 
     def timing(self, name, duration, **kwargs):
         self._metric.timing(name, duration)
+
+
+def make_tags(base=None, **kwargs):
+    # type: (Sequence[str], **Any) -> Sequence[str]
+    """Generate a list of tag values"""
+    tags = list(base or [])
+    tags.extend('{}:{}'.format(key, val) for key, val in kwargs.iteritems())
+    return tags
 
 
 class DatadogMetrics(object):
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -8,6 +8,7 @@ from twisted.logger import Logger
 from twisted.python.failure import Failure
 
 from autopush.exceptions import RouterException
+from autopush.metrics import make_tags
 from autopush.router.apns2 import (
     APNSClient,
     APNS_MAX_CONNECTIONS,
@@ -62,7 +63,7 @@ class APNSRouter(object):
         self.ap_settings = ap_settings
         self._config = router_conf
         self.metrics = metrics
-        self._base_tags = []
+        self._base_tags = ["platform:apns"]
         self.apns = dict()
         for rel_channel in self._config:
             self.apns[rel_channel] = self._connect(rel_channel,
@@ -144,8 +145,10 @@ class APNSRouter(object):
             apns_client.send(router_token=router_token, payload=payload,
                              apns_id=apns_id)
         except (ConnectionError, AttributeError) as ex:
-            self.metrics.increment("updates.client.bridge.apns.connection_err",
-                                   self._base_tags)
+            self.metrics.increment("notification.bridge.error",
+                                   tags=make_tags(self._base_tags,
+                                                  application=rel_channel,
+                                                  reason="connection_error"))
             self.log.error("Connection Error sending to APNS",
                            log_failure=Failure(ex))
             raise RouterException(
@@ -165,11 +168,19 @@ class APNSRouter(object):
 
         location = "%s/m/%s" % (self.ap_settings.endpoint_url,
                                 notification.version)
+        self.metrics.increment("notification.bridge.sent",
+                               tags=make_tags(self._base_tags,
+                                              application=rel_channel))
+
         self.metrics.increment(
             "updates.client.bridge.apns.%s.sent" %
             router_data["rel_channel"],
             self._base_tags
         )
+        self.metrics.gauge("notification.message_data",
+                           notification.data_length,
+                           tags=make_tags(self._base_tags,
+                                          destination='Direct'))
         return RouterResponse(status_code=201, response_body="",
                               headers={"TTL": notification.ttl,
                                        "Location": location},
--- a/autopush/router/simple.py
+++ b/autopush/router/simple.py
@@ -29,6 +29,7 @@ from twisted.web._newclient import ResponseFailed
 from twisted.web.client import FileBodyProducer
 
 from autopush.exceptions import RouterException
+from autopush.metrics import make_tags
 from autopush.protocol import IgnoreBody
 from autopush.router.interface import RouterResponse
 from autopush.types import JSONDict  # noqa
@@ -61,9 +62,15 @@ class SimpleRouter(object):
         """Stubbed out for this router"""
 
     def stored_response(self, notification):
+        self.metrics.gauge("notification.message_data",
+                           notification.data_length,
+                           tags=make_tags(destination='Stored'))
         return RouterResponse(202, "Notification Stored")
 
     def delivered_response(self, notification):
+        self.metrics.gauge("notification.message_data",
+                           notification.data_length,
+                           tags=make_tags(destination='Direct'))
         return RouterResponse(200, "Delivered")
 
     @inlineCallbacks
@@ -98,7 +105,6 @@ class SimpleRouter(object):
                     # in AWS or if the connection timesout.
                     self.log.debug("Could not route message: {exc}", exc=exc)
             if result and result.code == 200:
-                self.metrics.increment("router.broadcast.hit")
                 returnValue(self.delivered_response(notification))
 
         # Save notification, node is not present or busy
@@ -108,7 +114,6 @@ class SimpleRouter(object):
         try:
             result = yield self._save_notification(uaid_data, notification)
             if result is False:
-                self.metrics.increment("router.broadcast.miss")
                 returnValue(self.stored_response(notification))
         except JSONResponseError:
             raise RouterException("Error saving to database",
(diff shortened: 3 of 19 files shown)


autopush/websocket.py
```diff
@@ -1120,8 +1105,10 @@ def finish_webpush_notifications(self, result):
             if self.sent_notification_count > self.ap_settings.msg_limit:
                 raise MessageOverloadException()
             if notif.topic:
-                self.metrics.increment("updates.notification.topic",
+                self.metrics.increment("notification.topic",
                                        tags=self.base_tags)
+            self.metrics.gauge('ua.message_data', len(msg.get('data', '')),
+                               tags={"source": msg.pop('source', 'Direct')})
```
pjenvey: you want notif.source vs msg.pop('source'). this could use a test

autopush/tests/test_web_base.py
```diff
@@ -57,8 +58,11 @@ def setUp(self):
                                  headers={"ttl": "0"},
                                  host='example.com:8080')
 
+        mock_db = Mock()
```
pjenvey: I have a test_db() in support you can also use, but it's not important here/this is fine

autopush/tests/test_websocket.py
```diff
@@ -435,10 +423,10 @@ def test_close_with_delivery_cleanup_and_get_no_result(self):
...
         # Close the connection
         self.proto.onClose(True, None, None)
-        yield self._wait_for(lambda: len(self.metrics.mock_calls) > 2)
-        eq_(len(self.metrics.mock_calls), 3)
-        self.metrics.increment.assert_called_with(
-            "client.notify_uaid_failure", tags=None)
+        yield self._wait_for(lambda: len(self.metrics.mock_calls) > 0)
+        eq_(self.metrics.timing.call_args[0][0], 'ua.connection.lifespan')
+        # Wait for final cleanup (no error or metric produced)
+        yield sleep(1)
```
pjenvey: are these sleeps really necessary?

autopush/router/simple.py
```diff
@@ -61,9 +61,15 @@ def amend_endpoint_response(self, response, router_data):
         """Stubbed out for this router"""
 
     def stored_response(self, notification):
+        self.metrics.gauge("notification.message_data",
```
pjenvey: these would be simplepush only as webpush's router overrides these methods.

Would this stat be interesting on external bridges too? If so it might make sense for the WebHandlers to be emitting these

autopush/tests/test_websocket.py
jrconlin: Sadly, yes. Otherwise we get coverage issues.

autopush/web/base.py
```diff
@@ -283,12 +284,26 @@ def _router_fail_err(self, fail):
                 self.log.debug(format="Success", status_code=exc.status_code,
                                logged_status=exc.logged_status or 0,
                                client_info=self._client_info)
+                success = True
+                self.metrics.increment('notification.message.success',
+                                       tags=dict(
```
jrconlin: Ah, I thought these were being normalized. Will correct the others as well.

autopush/router/simple.py
jrconlin: True, but I wanted to get these calls about as close to the point that the action occurs as possible. Everything goes through `route_notification`, and that's really the only time we can say with authority what the disposition is & have the data size. 

I'll add the calls to the other methods as well. (Although bridge would pretty much only be "Direct")

... and just realized there's a bunch of metrics we missed for the audit.

autopush/web/base.py
```diff
@@ -283,12 +284,26 @@ def _router_fail_err(self, fail):
...
                                status_code=exc.status_code,
                                logged_status=exc.logged_status or 0,
                                errno=exc.errno or 0,
                                client_info=self._client_info)
+        if not success:
+            self.metrics.increment('notification.message.error',
+                                   tags=dict(
+                                       code=exc.status_code,
+                                       router=router_type,
+                                       vapid=(vapid is True)
```
pjenvey: forgot this one and one in webpush

autopush/tests/test_websocket.py
pjenvey: if you keep the older _wait_for on more metrics emitted + the client.notify_uaid_failure check that should do it instead (and a little quicker than a full sleep(1) -- coverage is ultimately waiting on that client.notify_uaid_failure

autopush/websocket.py
```diff
@@ -852,8 +836,9 @@ def _verify_user_record(self):
                                uaid_record=dump_uaid(record))
                 self.force_retry(self.db.router.drop_user,
                                  self.ps.uaid)
-                self.metrics.increment("client.drop_user",
-                                       tags={"errno": 105})
+                tags = list('code:105')
```
pjenvey: forgot ['code:105'] here

autopush/metrics.py
```diff
@@ -89,19 +89,34 @@ def __init__(self, api_key, app_key, hostname, flush_interval=10,
...
                            roll_up_interval=self._flush_interval)
 
     def increment(self, name, count=1, **kwargs):
+        if 'tags' in kwargs:
+            kwargs['tags'] = self._normalize_tags(kwargs['tags'])
         self._client.increment(self._prefix_name(name), count, host=self._host,
                                **kwargs)
 
     def gauge(self, name, count, **kwargs):
+        if 'tags' in kwargs:
```
pjenvey: nitnitpick: tags=None explicitly in the func prototype is a lil cleaner

autopush/tests/test_websocket.py
jrconlin: Hrm, this might be another bit that got messed up in the patch I had to do to catch up. And looks like `client.notify_uaid_failure` isn't listed in the table. The metric had previously been removed.

autopush/web/base.py
jrconlin: handled in subsequent update

jrconlin:
Reduced the sleeps to .25s. Granted, it's guesswork as to the optimal time required, but I'll admit that 1s is overkill.

autopush/web/base.py
pjenvey: still here

autopush/metrics.py
```diff
@@ -89,19 +89,32 @@ def __init__(self, api_key, app_key, hostname, flush_interval=10,
     def _prefix_name(self, name):
         return "%s.%s" % (self._namespace, name)
 
+    def _normalize_tags(self, tags):
```
pjenvey: kill all this now that we'll be consistent w/ always passing lists of strs

autopush/router/apnsrouter.py
```diff
@@ -165,11 +168,19 @@ def _route(self, notification, router_data):
...
+                               tags=tags)
+
         self.metrics.increment(
             "updates.client.bridge.apns.%s.sent" %
             router_data["rel_channel"],
             self._base_tags
         )
+        self.metrics.gauge("notification.message_data",
+                           len(notification.data or ""),
+                           tags={'destination': 'Direct'})
```
pjenvey: tags=['destination:Direct'] -- there's about 8 of these to fix in this diff

autopush/web/webpush.py
```diff
@@ -502,18 +508,30 @@ def _router_completed(self, response, uaid_data, warning=""):
...
             elif response.status_code == 202 or response.logged_status == 202:
                 self.log.debug(
                     format="Router miss, message stored.",
                     client_info=self._client_info)
-            self.metrics.timing("updates.handled", duration=time_diff)
+                dest = 'Stored'
+            self.metrics.timing("notification.request_time",
+                                duration=time_diff)
+            self.metrics.increment('notification.message.success',
+                                   tags=dict(
```
pjenvey: convert to list too

autopush/web/base.py
```diff
@@ -283,12 +284,26 @@ def _router_fail_err(self, fail):
...
                                status_code=exc.status_code,
                                logged_status=exc.logged_status or 0,
                                errno=exc.errno or 0,
                                client_info=self._client_info)
+        if not success:
+            self.metrics.increment('notification.message.error',
+                                   tags=[
+                                        "code:{}".format(exc.status_code),
+                                        "router:{}".format(router_type),
+                                        "vapid:{}".format(vapid is True)
```
pjenvey: there's still 2 bogus "vapid is True" in this patch: here and in webpush _router_completed. actually, both of these methods don't even need the full vapid dict. they just need a bool flag

webpush.py can be repsonsible for "vapid = jwt is not None" and pass that along to them

autopush/web/webpush.py
```diff
@@ -502,18 +508,30 @@ def _router_completed(self, response, uaid_data, warning=""):
...
                                client_info=self._client_info)
             elif response.status_code == 202 or response.logged_status == 202:
                 self.log.debug(
                     format="Router miss, message stored.",
                     client_info=self._client_info)
-            self.metrics.timing("updates.handled", duration=time_diff)
+                dest = 'Stored'
+            self.metrics.timing("notification.request_time",
+                                duration=time_diff)
+            self.metrics.increment('notification.message.success',
```
pjenvey: just occurred to me that web.simplepush's missing this one

jrconlin: added

jrconlin:
Went ahead and created the `make_tags` func since you've asked that we drop all dicts from tags. 

which means closing #950

autopush/metrics.py
```diff
@@ -74,6 +74,14 @@ def timing(self, name, duration, **kwargs):
         self._metric.timing(name, duration)
 
 
+def make_tags(base=[], **kwargs):
```
pjenvey: although this is a safe use of a list as a default kwarg, might just wanna use base=None instead (kinda surprised flake8 didn't complain, I think PyCharm will)

jrconlin: Huh, neither did. But, yeah, I keep forgetting about not doing that.

autopush/web/base.py
```diff
@@ -283,12 +284,26 @@ def _router_fail_err(self, fail):
                 self.log.debug(format="Success", status_code=exc.status_code,
                                logged_status=exc.logged_status or 0,
                                client_info=self._client_info)
+                success = True
+                self.metrics.increment('notification.message.success',
+                                       tags=[
+                                           'destination:"Direct"',
```
bbangert: Value has double quotes around it but none of the others do?

bbangert:
I think only the last comment really needs to be addressed before merge, as we should be consistent in value quoting for tags.

PR #951, opened 2017-07-07
feat: Do not do VAPID auth on trusted connections

### LINKED ISSUE #938: Don't check Vapid sig value for partner endpoint users
Partners using cert'd endpoints have already validated themselves. 
Don't bother checking the signature block.

jrconlin:
Closes #938

### DIFF
--- a/autopush/jwt.py
+++ b/autopush/jwt.py
@@ -78,7 +78,21 @@ class VerifyJWT(object):
         return payload, encoded
 
     @staticmethod
-    def decode(token, key):
+    def extract_assertion(token):
+        # type (str) -> JSONDict
+        """Extract the assertion dictionary from the passed token. This does
+        NOT do validation.
+
+        :param token: Partial or full VAPID auth token
+        :return dict of the VAPID claims
+
+        """
+        return json.loads(
+            base64.urlsafe_b64decode(
+                repad(token.split('.')[1]).encode('utf8')))
+
+    @staticmethod
+    def validate_and_extract_assertion(token, key):
         # type (str, str) -> JSONDict
         """Decode a web token into a assertion dictionary.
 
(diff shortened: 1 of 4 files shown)


autopush/web/webpush.py
```diff
@@ -367,7 +367,11 @@ def validate_auth(self, d):
                 public_key = vapid_auth['k']
             else:
                 public_key = d["subscription"].get("public_key")
-            jwt = extract_jwt(token, public_key)
+            jwt = extract_jwt(
+                token,
+                public_key,
+                is_trusted=self.context['settings'].client_certs is not None
```
pjenvey: we have a settings.enable_tls_auth to use instead

autopush/jwt.py
```diff
@@ -78,6 +78,20 @@ def extract_signature(auth):
         return payload, encoded
 
     @staticmethod
+    def extract(token):
```
bbangert: Maybe rename to ``extract_assertion`` ?

On first glance why you'd use decode vs. extract makes little sense without reading docs. Having slightly more info in the method name would help readability. Maybe they should both be ``decode_*`` for consistency, or ``extract_*``?

autopush/utils.py
```diff
@@ -188,12 +188,14 @@ def decipher_public_key(key_data):
     raise ValueError("Unknown public key format specified")
 
 
-def extract_jwt(token, crypto_key):
-    # type: (str, str) -> Dict[str, str]
+def extract_jwt(token, crypto_key, is_trusted=False):
```
bbangert: Maybe this new option is called ``verify``?

autopush/utils.py
```diff
@@ -188,12 +188,14 @@ def decipher_public_key(key_data):
...
+    # type: (str, str, bool) -> Dict[str, str]
     """Extract the claims from the validated JWT. """
     # first split and convert the jwt.
     if not token or not crypto_key:
         return {}
+    if is_trusted:
+        return jwt.extract(token)
     return jwt.decode(token, decipher_public_key(crypto_key.encode('utf8')))
```
bbangert: And maybe we either ``unverified_extract`` and ``verified_extract`` so its clear?

bbangert:
Something about the method naming here feels prone to confusion, I think cleaning up the names slightly should alleviate that. Looks good other than that!

autopush/jwt.py
jrconlin: Yeah, I was keeping with the original library which I am emulating, but no reason not to be clearer about things.

Issue #949, opened 2017-07-06
enable pyup for security updates


pjenvey:
We might as well utilize pyup now that it's available to us for security updates

Issue #454, opened 2016-04-26
Add SHA dependency checks


bbangert:
pip now lets us add SHA hashes to dependencies, we should use these:
- https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode

pjenvey:
These can be potentially bothersome but should be worth it. You usually include a hash of the .tar.gz and a platform's .whl, but can include as many as needed

The hashin utility can produce these for us. But I'm piptools' pip-compile could get support for what hashin does soon:

https://github.com/nvie/pip-tools/issues/303

w/ pip-compile you can keep the most bare minimum requirements in a '.in' file (a lot like what you'd have in setuptools requires).

It then version pins everything from it including all the sub dependencies. It mentions in the final requirements file where those sub dependencies are from.

e.g.:
six==1.10.0               # via autobahn, cryptography, pyopenssl, python-dateutil, python-jose, txaio

Then it has a --upgrade feature, which is similar to what pyup provides

oberstet:
You could steel from [this](https://github.com/crossbario/crossbar/blob/master/Makefile#L42) .. using [hashin](https://pypi.python.org/pypi/hashin)

pjenvey:
FYI @oberstet pip-tools actually has support for handling these now (that nvie/pip-tools#303 should be closed out). When be switch to it we'll be gaining the .in file format/upgrading and now the hash support too

pjenvey:
We can't enable hash checks while still using editable url specs (txstatsd and temporarily crytography).

I've pinged txstatsd's lauchpad for a new release.. but will probably have to get in touch with them via some other means

pjenvey:
endpoint still needs it, no?

bbangert:
Aw right, we're still on twisted for the endpoint, doh.

bbangert:
Won't be addressing this as we're moving entirely to Rust soonish.

PR #955, opened 2017-07-13
chore: add a pyup config file

pjenvey:
also upgrade deps via pip-tools: and kill the unneeded future dep and
move pycrypto into tests only

also title case the Issue/Closes footer in CONTRIBUTING

Closes: #949
Issue: #454

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,10 @@
+update: insecure
+schedule: "every day"
+search: False
+requirements:
+  - requirements.txt
+branch_prefix: chore/pyup/
+assignees:
+ - bbangert
+ - jrconlin
+ - pjenvey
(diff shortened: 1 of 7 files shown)


jrconlin:
so we have to include all the referenced libraries too? Granted, not super opposed, but seems like things could get weighty fast, particularly if a dependent library removes dependencies later.

pjenvey:
You *need* all referenced deps for hash checking mode, it's all or nothing. You really want them anyway, hashes or not, so the environment's reproducible.

We're already including most referenced deps (likely via pip freeze) but I think we've missed some recently, depending on how we were updating requirements.txt.

pip-compile makes it much more manageable especially w/ the comments on where the deps came from

PR #959, opened 2017-07-19
fix: move metric increments to lowest callbacks

### LINKED ISSUE #958: Message metrics are inaccurate
The new message metrics fail to record message success if exc.log_exception is False under _router_fail_err. Since _router_fail_err ends up calling _router_response, that should probably be responsible for adding the metric increments.

bbangert:
The metric increments were being called for registration API
calls due to an error callback. They weren't called for success
cases as well. Moving them to the lower callbacks with a new
flag should help ensure they're incremented correctly.

Closes #958

### DIFF
--- a/autopush/web/base.py
+++ b/autopush/web/base.py
@@ -154,6 +154,7 @@ class BaseWebHandler(BaseHandler):
         self._base_tags = {}
         self._start_time = time.time()
         self._timings = {}
+        self._handling_message = False
 
     @property
     def routers(self):
@@ -186,7 +187,9 @@ class BaseWebHandler(BaseHandler):
     #############################################################
     def _write_response(self, status_code, errno, message=None, error=None,
                         headers=None,
-                        url=DEFAULT_ERR_URL):
+                        url=DEFAULT_ERR_URL,
+                        router_type=None,
+                        vapid=None):
         """Writes out a full JSON error and sets the appropriate status"""
         self.set_status(status_code, reason=error)
         error_data = dict(
@@ -207,6 +210,13 @@ class BaseWebHandler(BaseHandler):
         if status_code == 410:
             self.set_header("Cache-Control", "max-age=86400")
 
+        if self._handling_message and status_code >= 300:
+            self.metrics.increment('notification.message.error',
+                                   tags=[
+                                       "code:{}".format(status_code),
+                                       "router:{}".format(router_type),
+                                       "vapid:{}".format(vapid is not None)
+                                   ])
         self._track_timing()
         self.finish()
 
@@ -255,7 +265,7 @@ class BaseWebHandler(BaseHandler):
         self._write_response(503, errno=202,
                              message="Communication error, please retry")
 
-    def _router_response(self, response):
+    def _router_response(self, response, router_type, vapid):
         for name, val in response.headers.items():
             if val is not None:
                 self.set_header(name, val)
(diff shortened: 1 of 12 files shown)


autopush/web/base.py
```diff
@@ -207,6 +210,13 @@ def _write_response(self, status_code, errno, message=None, error=None,
         if status_code == 410:
             self.set_header("Cache-Control", "max-age=86400")
 
+        if self._handling_message and status_code >= 300:
```
jrconlin: This will only metric messages that are potentially stored and handled by Autopush and not passed via a bridge, correct?

autopush/web/base.py
```diff
@@ -288,27 +312,13 @@ def _router_fail_err(self, fail, router_type=None, vapid=False):
...
                                client_info=self._client_info)
-        if not success:
-            self.metrics.increment('notification.message.error',
-                                   tags=[
-                                        "code:{}".format(exc.status_code),
-                                        "router:{}".format(router_type),
-                                        "vapid:{}".format(vapid is not None)
-                                   ])
-        self._router_response(exc)
+        self._router_response(exc, router_type, vapid)
```
jrconlin: Should we pass just the boolean state of VAPID rather than the whole structure? Are we planning on extracting more from the vapid data?

pjenvey:
_router_fail_err's no longer needed in registration, so we should nuke those calls. Then it and router_response are only called by web.webpush/simple and they can assume metrics should always be emitted (without the need for the _handling_message flag). The router_type and vapid args can become required.

only difference of simplepush's usage is it doesn't emit a vapid tag. not a big deal or easily hardcoded around if router_type == "simplepush"

..also leftover from when this landed, I thought we fixed the vapid arg but it seems still a bit off (sometimes a bool, sometimes not)

autopush/web/base.py
bbangert: Only the simplepush/webpush handlers set this flag, yes.

autopush/web/base.py
bbangert: We are not, but it does kind of help for consistency.

bbangert:
@pjenvey agreed, I think that should be done as a refactor to clean up all these dangling callbacks though.

autopush/web/base.py
jrconlin: Ah, I didn't see any metrics for `autopush.notification.bridge.sent` in datadog. I presume a separate issue will be filed for that?

bbangert: That's probably because those metrics were sent passing the base_tags as the name, not the proper values. I'm guessing they were dropped as a result. I'll fix that too.

bbangert: Ok, updated to pass in the tags as the tags= value, that should fix the bridge.sent metrics.

autopush/web/base.py
```diff
@@ -255,27 +265,41 @@ def _boto_err(self, fail):
         self._write_response(503, errno=202,
                              message="Communication error, please retry")
 
-    def _router_response(self, response):
+    def _router_response(self, response, router_type=None, vapid=False):
```
pjenvey: router_type and vapid should be required (more importantly, webpush and simplepush aren't calling it w/ these args and should be)

autopush/web/webpush.py
```diff
@@ -441,6 +441,11 @@ class WebPushHandler(BaseWebHandler):
                             "authorization")
     cors_response_headers = ("location", "www-authenticate")
 
+    def initialize(self):
+        """Must run on initialization to set ahead of validation"""
+        super(WebPushHandler, self).initialize()
+        self._handling_message = True
```
pjenvey: FYI this could have been a class var, but I think we'll shuffle this around later so don't care

autopush/web/base.py
```diff
@@ -207,6 +210,13 @@ def _write_response(self, status_code, errno, message=None, error=None,
         if status_code == 410:
             self.set_header("Cache-Control", "max-age=86400")
 
+        if self._handling_message and status_code >= 300:
+            self.metrics.increment('notification.message.error',
```
pjenvey: we will get the metric in all cases here, but router_type and vapid will be None in certain cases

autopush/web/webpush.py
```diff
@@ -520,24 +525,15 @@ def _router_completed(self, response, uaid_data, warning="",
...
                 self.log.debug(format="Successful delivery",
                                client_info=self._client_info)
             elif response.status_code == 202 or response.logged_status == 202:
                 self.log.debug(
                     format="Router miss, message stored.",
                     client_info=self._client_info)
-                dest = 'Stored'
             self.metrics.timing("notification.request_time",
                                 duration=time_diff)
-            self.metrics.increment('notification.message.success',
```
pjenvey: simplepush.py needs a similar diff

pjenvey:
What do we get out of a message size Counter?

autopush/web/base.py
bbangert: That is correct, sometimes it'll fail before getting that knowledge. So it will just be logged as a None value.

Issue #799, opened 2017-02-06
End server support for SImplePush


jrconlin:
Simplepush is no longer supported and is not actively being used in field. Requirements for simplepush compatibility restrict improvements and advancements for WebPush. It is in the projects best interest to officially drop support for the long obsolete tech.

jrconlin:
Intent to Disable note sent to dev planning

https://groups.google.com/forum/#!topic/mozilla.dev.planning/7jdtLa5JMko

bbangert:
According to our docs, we will be supporting SimplePush through Q1/2018.

PR #981, opened 2017-07-24
feat: make simplepush protocol optional

jrconlin:
Preliminary work for removing simplepush protocol by making it optional
(default off)

Issue #799

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -208,7 +208,6 @@ class EndpointApplication(AutopushMultiService):
             cors=not ns.no_cors,
             bear_hash_key=ns.auth_key,
             proxy_protocol_port=ns.proxy_protocol_port,
-            use_cryptography=ns.use_cryptography,
         )
 
 
@@ -282,5 +281,4 @@ class ConnectionApplication(AutopushMultiService):
             auto_ping_timeout=ns.auto_ping_timeout,
             max_connections=ns.max_connections,
             close_handshake_timeout=ns.close_handshake_timeout,
-            use_cryptography=ns.use_cryptography,
         )
--- a/autopush/main_argparse.py
+++ b/autopush/main_argparse.py
@@ -103,6 +103,10 @@ def add_shared_args(parser):
                         help="Use the cryptography library vs. JOSE",
                         action="store_true",
                         default=False, env_var="USE_CRYPTOGRAPHY")
+    parser.add_argument('--enable_simplepush',
+                        help="Enable the deprecated Simplepush protocol",
+                        action="store_true", default=True,
+                        env_var="ENABLE_SIMPLEPUSH")
     # No ENV because this is for humans
     _add_external_router_args(parser)
     _obsolete_args(parser)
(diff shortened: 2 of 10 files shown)


autopush/main.py
```diff
@@ -209,6 +209,7 @@ def from_argparse(cls, ns):
             bear_hash_key=ns.auth_key,
             proxy_protocol_port=ns.proxy_protocol_port,
             use_cryptography=ns.use_cryptography,
+            enable_simplepush=ns.enable_simplepush,
```
pjenvey: FYI this and use_cryptography could go in AutopushSettings.from_argparse since both endpoint/conn need it

jrconlin: Oh, good point!

autopush/main_argparse.py
```diff
@@ -103,6 +103,10 @@ def add_shared_args(parser):
                         help="Use the cryptography library vs. JOSE",
                         action="store_true",
                         default=False, env_var="USE_CRYPTOGRAPHY")
+    parser.add_argument('--enable_simplepush',
+                        help="Enable the deprecated Simplepush protocol",
+                        action="store_true", default=False,
```
bbangert: Might want to default to true first?

autopush/router/__init__.py
```diff
@@ -17,19 +17,20 @@
...
     router_conf = settings.router_conf
     routers = dict(
-        simplepush=SimpleRouter(
-            settings, router_conf.get("simplepush"), db, agent),
         webpush=WebPushRouter(settings, None, db, agent)
     )
+    if settings.enable_simplepush:
+        routers['simplepush'] = SimpleRouter(
+            settings, router_conf.get("simplepush"), db, agent)
+        # __all__.append("SimpleRouter")
```
bbangert: Commented out?

autopush/router/__init__.py
```diff
@@ -17,19 +17,20 @@
 from autopush.router.fcm import FCMRouter
 from autopush.settings import AutopushSettings  # noqa
 
-__all__ = ["APNSRouter", "FCMRouter", "GCMRouter", "SimpleRouter",
-           "WebPushRouter"]
+__all__ = ["APNSRouter", "FCMRouter", "GCMRouter", "WebPushRouter"]
```
bbangert: This is mainly used for imports, and should likely remain.

autopush/main_argparse.py
jrconlin: Well, I figured that we were soft-killing Simplepush and that we could optionally re-enable it if we needed to. Making this default true means we have to mindfully disable it instead. I'll admit that both are equivalent, but one is more aggressive.

But, sure, fine, ruin my hopes for an early death for simplepush... ;)

autopush/router/__init__.py
jrconlin: Yeah, I was sorta debating that. We don't have anything that doesn't explicitly import `SimpleRouter`, and since it's deprecated, I figured it would be useful to make sure that nothing is implicitly importing it either. 

I'm fine adding it back in since it technically still exists.

autopush/router/__init__.py
jrconlin: whoops, sorry.

PR #987, opened 2017-07-31
feat: remove use of moto dynamodb2 from tests

bbangert:
Moves the local DynamoDB run to the package level and all tests use
it instead of running moto's incomplete dynamodb mock.

This also fixes a bug in the tests that left autopush.db.key_hash
set after an endpoint test ran.


pjenvey:
this should really also kill the moto req in test-requirements.txt

PR #993, opened 2017-08-17
bug: Add missing metric for notification encoding

### LINKED ISSUE #992: add a metric for ECE crypto header
We only log the crypto header type in client_info (which is no longer always logged). We need to measure its usage, so it should be tallied as a metric

jrconlin:
Added "updates.notification.encoding.{content-encoding}",
we already track vapid as "updates.vapid.{version}"

Closes #992

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -477,6 +477,9 @@ class WebPushHandler(BaseWebHandler):
         encoding = ''
         if notification.data and notification.headers:
             encoding = notification.headers.get('encoding', '')
+            self.metrics.increment(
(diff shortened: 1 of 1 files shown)


autopush/web/webpush.py
```diff
@@ -477,6 +477,9 @@ def post(self,
         encoding = ''
         if notification.data and notification.headers:
             encoding = notification.headers.get('encoding', '')
+            self.metrics.increment(
```
pjenvey: When there's no body we'll get an empty string and thus a "updates.notification.encoding." metric which is a bit odd, but I don't actually care :)

PR #995, opened 2017-08-22
bug: change "enable_simplepush" flag to "disable_simplepush"

### LINKED ISSUE #994: Configuration flag for simplepush conflicted
Due to a last minute change, the configuration flag for "enable_simplepush" was improperly initiated.

jrconlin:
Due to a last minute change in the prior patch, the initial state of the
config option was improperly set. In addition the simplepush
registration was succeeding when it should have been rejected. Added
testing to verify both conditions are satisfied.

Closes #994

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -116,12 +116,12 @@ keyid="http://example.org/bob/keys/123;salt="XZwpw6o37R-6qoZjw6KwAw"\
         self.ws.send(msg)
         result = json.loads(self.ws.recv())
         log.debug("Recv: %s", result)
+        eq_(result["status"], 200)
         if self.uaid and self.uaid != result["uaid"]:  # pragma: nocover
             log.debug("Mismatch on re-using uaid. Old: %s, New: %s",
                       self.uaid, result["uaid"])
             self.channels = {}
         self.uaid = result["uaid"]
-        eq_(result["status"], 200)
         return result
 
     def register(self, chid=None, key=None):
@@ -309,7 +309,7 @@ class IntegrationBase(unittest.TestCase):
         storage_table=dict(tablename=STORAGE_TABLE),
         message_table=dict(tablename=MESSAGE_TABLE),
         use_cryptography=True,
-        enable_simplepush=True,
+        disable_simplepush=False,
     )
 
     _conn_defaults = dict(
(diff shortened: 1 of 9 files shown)


autopush/websocket.py
```diff
@@ -1215,6 +1215,8 @@ def process_register(self, data):
             d = self.deferToThread(self.conf.make_endpoint,
                                    self.ps.uaid, chid, data.get("key"))
         else:
+            if self.conf.disable_simplepush:
+                return self.bad_message("register", "No longer supported")
```
bbangert: Shouldn't we have dropped them by now?

autopush/tests/test_integration.py
```diff
@@ -619,6 +619,18 @@ def test_simplepush_data_delivery(self):
...
     @inlineCallbacks
+    def test_disabled_simplepush(self):
+        state = self.conn.conf.disable_simplepush
+        self.conn.conf.disable_simplepush = True
+        try:
+            yield self.quick_register()
+        except AssertionError as ex:
+            # Check the raised assertion, since the true one is buried.
+            ok_("401 " in ex.message)
+            pass
```
pjenvey: kill this line. To be safe I always go with an else: assert False, "Expected AssertionError w/ 401" block on try excepts like this. the last line resetting the flag should be in a finally block too.

the rarely seen try-except-else-finally

autopush/websocket.py
jrconlin: We should have. This ensures that any older clients do not get a URL. Previously, it was possible that this could occur.

autopush/tests/test_integration.py
jrconlin: Awesome. Wasn't super thrilled about that, and the suggestion is much better. (Plus, I learned about additional try states, so bonus.)

autopush/websocket.py
bbangert: How is it possible for a non-webpush client to connect and issue a register command if we drop them after hello?

bbangert: Ok, I see, my bad, I was under the impression we dropped them at hello if they weren't using webpush. Why not just drop at hello?

autopush/tests/test_integration.py
jrconlin: had to remove this test since I could not get a clean run without a reactor error.

PR #998, opened 2017-09-01
bug: allow only known protocols for /register calls

### LINKED ISSUE #997: 410 error just after registration
I'm trying to  get hands on the API, and I don't get why I'm getting a `410` error.

Register to webpush, this works:
```bash
curl 'https://updates.push.services.mozilla.com/v1/webpush/11-TESTING-11/registration' --data-binary '{"token":"11-TESTING-11-11"}'
```

I get a successful response:
```json
{
"channelID": "MY_CHID",
"endpoint": "https://updates.push.services.mozilla.com/wpush/v1/HIDDEN",
"uaid": "MY_UAID",
"secret": "MY_SECRET"
}
```

But if I'm trying to send a message to the endpoint (with web-push), I'm getting the following error:
```json
{
"code": 410,
"errno": 106,
"error": "",
"more_info": "http://autopush.readthedocs.io/en/latest/http.html#error-codes",
"message": "Request did not validate No such subscription"
}
```

Why would the endpoint be instantly invalid ?

jrconlin:
Closes #997

### DIFF
--- a/autopush/web/registration.py
+++ b/autopush/web/registration.py
@@ -174,9 +174,16 @@ class AuthorizationCheckSchema(Schema):
 
 
 def conditional_token_check(object_dict, parent_dict):
-    if parent_dict['path_kwargs']['type'] in ['gcm', 'fcm']:
+    ptype = parent_dict['path_kwargs']['type']
+    # Basic "bozo-filter" to prevent customer surprises later.
+    if ptype not in ['apns', 'fcm', 'gcm', 'webpush', 'simplepush', 'test']:
+        raise InvalidRequest("Unknown registration type",
+                             status_code=400,
+                             errno=108,
+                             )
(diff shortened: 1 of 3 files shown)


autopush/web/registration.py
```diff
@@ -174,9 +174,16 @@ def validate_auth(self, data):
 
 
 def conditional_token_check(object_dict, parent_dict):
-    if parent_dict['path_kwargs']['type'] in ['gcm', 'fcm']:
+    ptype = parent_dict['path_kwargs']['type'].lower()
```
pjenvey: This may sound nitpicky but I'd rather *not* be lax about case here since it's from the url path -- mostly because we're case sensitive for any other path segment we're responsible for AFAICT (and I'm picky about URIs due to my old semantic web roots). Let's not start and keep it consistent.

PR #1001, opened 2017-09-05
feat: remove unused udp wake capability

### LINKED ISSUE #886: Obsolete UDP functions from autopush
There are a number of UDP options that are holdovers from the Telefonica work with Simplepush. These have never been actually used in deployment and can be safely made obsolete.

bbangert:
Closes #886


autopush/main_argparse.py
```diff
@@ -177,17 +177,6 @@ def _add_external_router_args(parser):
                                              "APNS settings",
                         type=str, default="",
                         env_var="APNS_CREDS")
-    # UDP
```
jrconlin: Move these to `_obsolete_args` to save oremj a headache

PR #1016, opened 2017-09-12
refactor: Remove unneeded features from Rust deps

alexcrichton:
Mostly just disable the TLS features in tungstenite/tokio-tungstenite as we
don't use them anyway!

### DIFF
--- a/autopush_rs/Cargo.lock
+++ b/autopush_rs/Cargo.lock
@@ -23,15 +23,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "advapi32-sys"
-version = "0.2.0"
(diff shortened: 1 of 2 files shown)


bbangert:
Does this remove the ability to support SSL connections directly? We were looking at perhaps moving to a NLB instead of ELB, which means we'd have to terminate our own SSL. Keeping that ability would be nice.

alexcrichton:
Sort of but not particularly. Right now there's not a simple switch for turning on SSL in the code. I think these features in the upstream crates were primarily meant for connecting to URLs that look like `wss://...` (which isn't used here).

If we'd like to add SSL support it shouldn't be too bad, but it'd require local changes and I think we'd still end up disabling these features in the upstream crates regardless

PR #1018, opened 2017-09-13
refactor: Reorganize the Rust ping manager

alexcrichton:
Focus on retaining just one `Timeout` instance instead of creating many, using
the `reset` function where possible. Additionally try to streamline the number
of timers that we hit during each pass to ensure that we're at most only hitting
one on the fast path.


alexcrichton:
Rebased!

PR #1019, opened 2017-09-13
feat: SSL support to the Rust server

alexcrichton:
This commit adds support for the Rust server to be configured at runtime to
accept websocket connections over TLS rather than plaintext. This basically just
boils down to reading and respecting the already-present options in the
configuration.


autopush_rs/src/server/tls.rs
```diff
@@ -0,0 +1,129 @@
...
+        None => return None,
+    };
+    let key = PKey::private_key_from_pem(&key)
+        .expect("failed to create private key");
+    let cert = read(opts.ssl_cert.as_ref().expect("ssl_cert not configured"));
+    let cert = X509::from_pem(&cert).expect("failed to create certificate");
+    let dh_param = opts.ssl_dh_param.as_ref().expect("ssl_dh_param not configured");
+    let dh_param = Dh::from_pem(&read(dh_param)).expect("failed to create dh");
+
+    let mut builder = SslAcceptorBuilder::mozilla_intermediate(
```
alexcrichton: y'all probably know more about the desired options here, so it may be useful to browse the [`SslAcceptorBuilder`](https://docs.rs/openssl/0.9.17/openssl/ssl/struct.SslAcceptorBuilder.html) docs and [`SslContextBuilder`](https://docs.rs/openssl/0.9.17/openssl/ssl/struct.SslContextBuilder.html) docs to see if this is missing anything!

autopush_rs/src/server/tls.rs
```diff
@@ -0,0 +1,129 @@
...
+/// `SslAcceptor` type is stored globally in the `Server` structure, later used
+/// to process all accepted TCP sockets.
+pub fn configure(opts: &ServerOptions) -> Option<SslAcceptor> {
+    let key = match opts.ssl_key {
+        Some(ref key) => read(key),
+        None => return None,
+    };
+    let key = PKey::private_key_from_pem(&key)
+        .expect("failed to create private key");
+    let cert = read(opts.ssl_cert.as_ref().expect("ssl_cert not configured"));
```
jrconlin: Much better errors than before. Yay!

alexcrichton:
At @bbangert's suggestion I've pushed up a commit now which also sets the `SSL_MODE_RELEASE_BUFFERS` flag

PR #1020, opened 2017-09-13
fix: don't check storage for new uaid's

### LINKED ISSUE #1017: Fix state machine to avoid new storage checks in Rust PoC
The Rust PoC triggers a storage check even for brand new users. It should return the check flag in addition to its others, so it can properly indicate if a check is even needed (it currently defaults to assuming it is).

bbangert:
Closes #1017

### DIFF
--- a/autopush/webpush_server.py
+++ b/autopush/webpush_server.py
@@ -157,6 +157,7 @@ class OutputCommand(object):
 class HelloResponse(OutputCommand):
     uaid = attrib()  # type: Optional[str]
     message_month = attrib()  # type: str
+    check_storage = attrib()  # type: bool
     reset_uaid = attrib()  # type: bool
     rotate_message_table = attrib(default=False)  # type: bool
 
@@ -337,6 +338,7 @@ class HelloCommand(ProcessorCommand):
         # type: (Hello) -> HelloResponse
         user_item = None
         flags = dict(
+            check_storage=False,
             message_month=self.db.current_msg_month,
             reset_uaid=False
         )
@@ -359,8 +361,9 @@ class HelloCommand(ProcessorCommand):
         # type: (Hello) -> (Optional[JSONDict], JSONDict)
         flags = dict(
             message_month=None,
-            rotate_message_table=False,
+            check_storage=False,
             reset_uaid=False,
+            rotate_message_table=False,
         )
         uaid = hello.uaid.hex
         try:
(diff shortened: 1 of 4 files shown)


autopush/webpush_server.py
```diff
@@ -359,8 +361,9 @@ def lookup_user(self, hello):
         # type: (Hello) -> (Optional[JSONDict], JSONDict)
         flags = dict(
             message_month=None,
-            rotate_message_table=False,
+            check_storage=False,
```
jrconlin: nit: can we not change the order of these? It's not like they're in alphabetical order anyway, and it's generally better to put new items at the end of lists to avoid accidental flag set.

autopush_rs/src/client.rs
```diff
@@ -218,9 +218,9 @@ where
             ClientState::WaitingForProcessHello(ref mut response) => {
                 debug!("State: WaitingForProcessHello");
                 match try_ready!(response.poll()) {
-                    call::HelloResponse { uaid: Some(uaid), message_month, reset_uaid, rotate_message_table }
-                    => self.data.process_hello(uaid, message_month, reset_uaid, rotate_message_table),
-                    _ => return Err("Already connected elsewhere".into()),
+                    call::HelloResponse { uaid: Some(uaid), message_month, check_storage, reset_uaid, rotate_message_table }
+                    => self.data.process_hello(uaid, message_month, check_storage, reset_uaid, rotate_message_table),
```
jrconlin: nit: this seems hard to read quickly. Should there be an indent here to designate that `call::HelloResponse` is the match?

autopush_rs/src/client.rs
```diff
@@ -435,9 +435,10 @@ where
         Ok(Async::Ready(item))
     }
 
-    fn process_hello(&mut self, uaid: Uuid, message_month: String, reset_uaid: bool, rotate_message_table: bool) -> ClientState {
+    fn process_hello(&mut self, uaid: Uuid, message_month: String, check_storage: bool, reset_uaid: bool, rotate_message_table: bool) -> ClientState {
```
jrconlin: nit: Again, I'd be happier if the new arg was at the end of this list.

autopush/webpush_server.py
bbangert: When creating a dict? I understand it in keyword args to API methods to add it to the end, but this just makes a dict.

autopush_rs/src/client.rs
bbangert: I have an upcoming PR that will run rustfmt over all of it, which will reformat this. I didn't want to fold that into this as there's lots of little tweaks in it.

autopush_rs/src/client.rs
bbangert: Ah yes, here it matters! Changing.

autopush/webpush_server.py
jrconlin: fair, but since we do specify a set of args that match these, I thought it might be nice for consistency.

Issue #1024, opened 2017-09-22
Remove redundant `get_uaid` call from router/webpush.py


jrconlin:
In WebPushRouter.route_notification. Look for "redundant" comment.

jrconlin:
Invalid see #1025 comments

PR #1025, opened 2017-09-25
bug: Remove redundant get_uaid call from websocket.

jrconlin:
UAID is already verified by the marshmallow code

Closes: #1024

### DIFF
--- a/autopush/router/webpush.py
+++ b/autopush/router/webpush.py
@@ -10,7 +10,6 @@ import time
 from StringIO import StringIO
 from typing import Any  # noqa
 
-from boto.dynamodb2.exceptions import ItemNotFound
 from boto.exception import JSONResponseError
 from twisted.internet.threads import deferToThread
 from twisted.web.client import FileBodyProducer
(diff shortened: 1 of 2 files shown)


autopush/router/webpush.py
```diff
@@ -109,32 +108,6 @@ def route_notification(self, notification, uaid_data):
...
-            self.metrics.increment("updates.client.deleted")
-            raise RouterException("User was deleted",
-                                  status_code=410,
-                                  response_body="Invalid UAID",
-                                  log_exception=False,
-                                  errno=105)
-
-        # Verify there's a node_id in here, if not we're done
-        node_id = uaid_data.get("node_id")
-        if not node_id:
```
pjenvey: Couldn't there still be a case of a node_id of None here?

pjenvey: Also we may have successfully called clear_node above and fell through here with a now stale node_id

bbangert:
The commit says 'Bug' instead of 'fix', and it says it removes it from websocket.py, but the issue cited says 'router/webpush.py' is where it should be removed from.

autopush/router/webpush.py
jrconlin: Hrm, I think this might be a different bug, thanks!

We fetch the `uaid_data` in [the marshmallow code](https://github.com/mozilla-services/autopush/blob/master/autopush/web/webpush.py#L85), which should contain the node_id. We don't check if that node_id is valid. We then try to [send to that node_id](https://github.com/mozilla-services/autopush/blob/master/autopush/router/webpush.py#L84), and if that fails with a `connection` error, we [return a `delivered_response`](https://github.com/mozilla-services/autopush/blob/master/autopush/router/webpush.py#L155), which I don't believe actually saves the message. Does the code seem to suggest that to you too? 

I believe that we need to check for node_id earlier, and only attempt delivery if it is present, as well as make sure that we save messages that either do not have a node_id assigned, or have failed to be delivered. Correct?

jrconlin:
Invalid

PR #1027, opened 2017-09-26
bug: Remove discussion comment and useless bool return

### LINKED ISSUE #1026: Remove merge comment, clarify wording, and drop useless return value
After https://github.com/mozilla-services/autopush/pull/1023, there was some discussion about the second `get_uaid` call in `WebPushRouter.route_notification`. This question was clarified, however the comment was determined to vague about what was happening.

In addition, there's a useless bool that was being returned from `Message.store_message` that could be removed to prevent similar confusion later.

jrconlin:
Closes #1026

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -436,7 +436,7 @@ class Message(object):
 
     @track_provisioned
     def store_message(self, notification):
-        # type: (WebPushNotification) -> bool
+        # type: (WebPushNotification) -> None
         """Stores a WebPushNotification in the message table"""
         item = dict(
             uaid=hasher(notification.uaid.hex),
@@ -448,7 +448,6 @@ class Message(object):
             updateid=notification.update_id
         )
         self.table.put_item(data=item, overwrite=True)
-        return True
 
     @track_provisioned
     def delete_message(self, notification):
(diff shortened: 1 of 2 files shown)


autopush/db.py
```diff
@@ -436,7 +436,7 @@ def save_channels(self, uaid, channels):
 
     @track_provisioned
     def store_message(self, notification):
-        # type: (WebPushNotification) -> bool
+        # type: (WebPushNotification)
```
pjenvey: mypy wants -> None here (don't ask me why because I do this all the time)

autopush/db.py
```diff
@@ -448,7 +448,6 @@ def store_message(self, notification):
             updateid=notification.update_id
         )
         self.table.put_item(data=item, overwrite=True)
-        return True
```
pjenvey: now we should stop checking for this result in route_notification too (and remove the silly "result is False" line)

autopush/db.py
jrconlin: Oh, huh, and pycharm noted that. Need to make it a more important warning.

autopush/db.py
jrconlin: AH! I was looking at the wrong .`save_notification` and wondering what you meant. 

And now I'm disturbed that removing that condition and the returned error didn't cause any test to fail or coverage line to be lost. (sigh)

PR #1030, opened 2017-09-27
fix: dump_rpy_heap before calling objgraph

### LINKED ISSUE #1028: /_memusage should dump_rpy_heap before calling objgraph
This makes a lot more sense: objgraph does a fair mount of work introspecting everything. We may have left over garbage from after its run which could potentially pollute dump_rpy_heap's output.

pjenvey:
just in case objgraph pollutes its output

Closes #1028

### DIFF
--- a/autopush/memusage.py
+++ b/autopush/memusage.py
@@ -24,11 +24,13 @@ def memusage():
     buf = StringIO()
     rusage = trap_err(resource.getrusage, resource.RUSAGE_SELF)
     buf.writelines([repr(rusage), '\n\n'])
-    trap_err(objgraph.show_most_common_types, limit=0, file=buf)
-    buf.write('\n\n')
     trap_err(pmap_extended, buf)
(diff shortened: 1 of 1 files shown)


autopush/memusage.py
```diff
@@ -24,11 +24,11 @@ def trap_err(func, *args, **kwargs):
     buf = StringIO()
     rusage = trap_err(resource.getrusage, resource.RUSAGE_SELF)
     buf.writelines([repr(rusage), '\n\n'])
-    trap_err(objgraph.show_most_common_types, limit=0, file=buf)
-    buf.write('\n\n')
     trap_err(pmap_extended, buf)
     trap_err(dump_rpy_heap, buf)
     trap_err(get_stats_asmmemmgr, buf)
+    buf.write('\n\n')
+    trap_err(objgraph.show_most_common_types, limit=0, file=buf)
```
jrconlin: nit: since it's important that this be the last call, might want to add a comment to say that.

PR #1034, opened 2017-10-03
bug: Handle legacy simplepush records as candidate webpush records

### LINKED ISSUE #1033: Trap for now invalid "simplepush" protocol
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/654984/

jrconlin:
Some older records in the database appear to have been miscategorized as
"simplepush". We should try to handle those as "webpush".

Closes #1033

### DIFF
--- a/autopush/tests/test_endpoint.py
+++ b/autopush/tests/test_endpoint.py
@@ -330,7 +330,7 @@ class RegistrationTestCase(unittest.TestCase):
         self.patch('uuid.uuid4', return_value=dummy_uaid)
 
         resp = yield self.client.post(
-            self.url(router_type="foo"),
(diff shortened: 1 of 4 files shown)


bbangert:
I think we shouldn't reject them because the client thinks they're valid. We don't actually verify that router_type is correct in websocket.py. We probably should verify that and force a reset if the router_type is wrong.

jrconlin:
These seem to be edge cases. I'm not exactly sure how they got their `route_type` set to `simplepush` since they have a `/wpush` url. Passing them on as a `webpush` is fairly harmless since any additional issues would be caught by the appropriate checks later.

Eventually, when we start enforcing the global TTLs we can drop this patch.

PR #1035, opened 2017-10-04
feat: add Strict-Transport-Security header

### LINKED ISSUE #1031: Add STS Headers
Per review add Strict-Transport-Security header

jrconlin:
Closes #1031

### DIFF
--- a/autopush/web/base.py
+++ b/autopush/web/base.py
@@ -163,6 +163,12 @@ class BaseWebHandler(BaseHandler):
                             ",".join(self.cors_request_headers))
             self.set_header("Access-Control-Expose-Headers",
                             ",".join(self.cors_response_headers))
+        if self.conf.sts_max_age:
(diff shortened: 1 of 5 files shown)


autopush/web/base.py
```diff
@@ -163,6 +163,12 @@ def prepare(self):
                             ",".join(self.cors_request_headers))
             self.set_header("Access-Control-Expose-Headers",
                             ",".join(self.cors_response_headers))
+        if self.conf.sts_max_age:
+            self.set_header("Strict-Transport-Security",
+                            ";".join([
+                                "max-age={}".format(self.conf.sts_max_age),
```
pjenvey: eh could have just done it all in the .format string but no biggie

jrconlin: Thought about it. Decided on this partly because there's a few other potential options and the wrapping would have gotten ugly.

PR #1042, opened 2017-10-10
feat: move tests to pytest

### LINKED ISSUE #1039: Consider moving autopush to newer testing schema
nose is no longer actively supported, so testing should move to a more supported framework like pytest or nose2.

jrconlin:
Issue #1039

General sweep of tests to make things "pytest" compliant. This means switching nosetool calls for their assert/pytest primatives. 

Note, there were a few tests in test_cryptokey that were invalid and have been fixed because of this.
In addition, some [light type correction](https://github.com/mozilla-services/autopush/compare/bug/1033?expand=1#diff-8c7a1b5174dad635894310a2ce89eb88L104) was made where appropriate.

(Also note: this PR does not fully close #1039. tox and other functions still use nose.)

### DIFF
--- a/autopush/tests/test_router.py
+++ b/autopush/tests/test_router.py
@@ -7,7 +7,7 @@ import decimal
 
 from autopush.utils import WebPushNotification
 from mock import Mock, PropertyMock, patch
-from nose.tools import eq_, ok_, assert_raises
+import pytest
 from twisted.trial import unittest
 from twisted.internet.error import ConnectionRefusedError
 from twisted.internet.defer import inlineCallbacks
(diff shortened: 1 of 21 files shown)


autopush/tests/test_router.py
```diff
@@ -1015,18 +1028,18 @@ def test_route_to_busy_node_with_ttl_zero(self):
 
         def verify_deliver(fail):
             exc = fail.value
-            ok_(exc, RouterResponse)
-            eq_(exc.status_code, 201)
-            eq_(len(self.metrics.increment.mock_calls), 0)
-            ok_("Location" in exc.headers)
+            assert exc, RouterResponse
```
pjenvey: I'm noticing a few of these asserts (originally ok_s) are incorrect. This and a couple others should be assert isinstance(exc, RouterResponse).  A couple "assert status.status_code, 201" below.

Logged a separate issue here:

https://github.com/mozilla-services/autopush/issues/1043

autopush/conftest.py
```diff
@@ -0,0 +1,11 @@
+from autopush.tests import setUp, tearDown
```
pjenvey: can we move conftest into tests/ ? (it usually lives there, in my experience)

autopush/tests/test_router.py
jrconlin: Yeah, I had to correct a few of those. Good catch on these. I'll clean them up.

Issue #1046, opened 2017-10-16
Make snaek's rust build optional


pjenvey:
snaek's rust build along with rust installation on travis should be optional per an environment variable. This should fix the mypy build (which was previously broken due to py3 compat but now seems broken for some other unknown reason) and speed up flake8's run

PR #1047, opened 2017-10-16
chore: allow WITH_RUST=false to disable the rust build

pjenvey:
also update travis to (hopefully) latest pypy

Closes: #1046

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,13 @@ matrix:
   include:
     - python: 2.7
       env: TOXENV=py27 DDB=true CODECOV=true
-    - python: pypy-5.7.1
+    - python: pypy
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -11,13 +11,13 @@ matrix:
   include:
     - python: 2.7
       env: TOXENV=py27 DDB=true CODECOV=true
-    - python: pypy-5.7.1
+    - python: pypy
```
bbangert: Which version does this peg us to?

pjenvey: This bumps us to the latest they have, 5.8.0. I'm hoping this automatically bumps us to 5.9.0 when they add it

Issue #1038, opened 2017-10-06
Workaround our support libraries' ssl.wrap_socket anti-pattern usage


pjenvey:
Per results of https://github.com/mozilla-services/push-service/issues/76, our support libraries (boto, gcmclient) use the ssl module's top level wrap_socket func which unnecessarily creates an SSLContext per connection. pypy's GC doesn't appear to keep up w/ all of them: in one scenario we see half of a 3gb heap taken by boto's SSLContext remnants.

SSLContext's aren't cheap depending on their loaded cert store. E.g. the mozilla cert store shipped w/ ubuntu 17.04 is 632k

pypy should better handle this (there's probably a bug here) but libraries shouldn't follow this anti- pattern either. They should create one SSLContext and call its wrap_socket method instead. The hyper lib (used by our apns2) does the right thing: allows passing in a custom SSLContext otherwise defaulting its own singleton

In the meantime we can workaround this with a monkey patch to ssl.wrap_socket that caches contexts.

pjenvey:
Marking as p2, I'm considering deploying just pypy 5.9 to prod before landing this change

PR #1053, opened 2017-10-18
feat: cache SSLContexts created by ssl.wrap_socket

pjenvey:
avoids unnecessary allocation of them which appears to stress pypy's
GC (w/ their finalization needs)

also fix some type signatures and kill some now unnecessary mock patches

Closes: #1038

### DIFF
--- a/autopush/main_argparse.py
+++ b/autopush/main_argparse.py
@@ -97,6 +97,10 @@ def add_shared_args(parser):
                         help="Max Strict Transport Age in seconds",
                         type=int, default=31536000,
                         env_var="STS_MAX_AGE")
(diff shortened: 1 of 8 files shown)


autopush/main_argparse.py
```diff
@@ -97,6 +97,10 @@ def add_shared_args(parser):
                         help="Max Strict Transport Age in seconds",
                         type=int, default=31536000,
                         env_var="STS_MAX_AGE")
+    parser.add_argument('--_no_cache_sslcontexts',
```
bbangert: no_sslcontext_cache might read better.

PR #1056, opened 2017-10-19
feat: address HTTP2 errors in APNS

### LINKED ISSUE #1052: Wrap all known routing failures with a Metric call
External routing can fail. Now that we have some production candidate errors in sentry, wrap the calls with a metric so we can monitor via dashboard.

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/605720/

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/605720/

jrconlin:
Converts two errors into metrics and attempts retry on Connection or
HTTP2 errors.

Closes #1052

### DIFF
--- a/autopush/router/apns2.py
+++ b/autopush/router/apns2.py
@@ -41,7 +41,8 @@ class APNSClient(object):
                  alt=False, use_sandbox=False,
                  max_connections=APNS_MAX_CONNECTIONS,
                  logger=None, metrics=None,
-                 load_connections=True):
+                 load_connections=True,
+                 max_retry=2):
         """Create the APNS client connector.
 
         The cert_file and key_file can be derived from the exported `.p12`
@@ -76,6 +77,8 @@ class APNSClient(object):
         :type metrics: autopush.metrics.IMetric
         :param load_connections: used for testing
         :type load_connections: bool
+        :param max_retry: Number of HTTP2 transmit attempts
+        :type max_retry: int
 
         """
         self.server = SANDBOX if use_sandbox else SERVER
@@ -84,6 +87,7 @@ class APNSClient(object):
         self.metrics = metrics
         self.topic = topic
         self._max_connections = max_connections
+        self._max_retry = max_retry
         self.connections = deque(maxlen=max_connections)
         if load_connections:
             self.ssl_context = hyper.tls.init_context(cert=(cert_file,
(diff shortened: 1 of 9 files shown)


autopush/router/apnsrouter.py
```diff
@@ -139,31 +140,39 @@ def _route(self, notification, router_data):
...
+        # APNs may force close a connection on us without warning.
+        # if that happens, retry the message.
+        success = False
+        for attempt in range(0, self._max_retry):
+            try:
+                apns_client.send(router_token=router_token, payload=payload,
+                                 apns_id=apns_id)
+                success = True
+                break
+            except AttributeError as ex:
```
pjenvey: We can kill the AttributeError handling in here and the associated one in apns2 at this point. It was from that old hyper bug that was eventually fixed. Doesn't have to be in this PR, though

autopush/router/apnsrouter.py
```diff
@@ -139,31 +140,39 @@ def _route(self, notification, router_data):
...
-            self.metrics.increment("notification.bridge.error",
-                                   tags=make_tags(self._base_tags,
-                                                  application=rel_channel,
-                                                  reason="connection_error"))
-            self.log.error("Connection Error sending to APNS",
-                           log_failure=Failure(ex))
+        # APNs may force close a connection on us without warning.
+        # if that happens, retry the message.
+        success = False
+        for attempt in range(0, self._max_retry):
```
pjenvey: nitpick: extra first arg of value 0 isn't necessary (same as range(self._maxretry))

autopush/router/apnsrouter.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, conf, router_conf, metrics, load_connections=True):
         self.metrics = metrics
         self._base_tags = ["platform:apns"]
         self.apns = dict()
+        self._max_retry = router_conf.get("max_retry", 2)
```
pjenvey: This could use an example in the conf file -- which leads me to notice this key sort of clashes with apns' router_conf format (a dict of rel_channel/cert_infos). Probably should be loaded from a separate config knob?

jrconlin: Well, this should come out of the router_config dict (`apns_creds`), so there should't be any conflicts. (basically, it's already in a different config knob set). 

Will add to sample config.

jrconlin: and, yes, `apns_certs` has been horribly overloaded to include "platform" specific configuration options like this. Legacy, is a thing.

configs/autopush_shared.ini.sample
```diff
@@ -125,7 +125,8 @@ endpoint_port = 8082
 ;        "topic": "com.mozilla.org.Firefox",  // Bundle ID for associated app
 ;        "max_connections": 100, // Max number of connection pool entries.
 ;        "sandbox": False},  // Use the APNs sandbox feature
+         "max_retry": 2, // Max number of retries in event of an HTTP2 error
```
pjenvey: So going by line 130 it's a knob specific to the rel_channel. Then the '}' on the line above should move down to this line.

But in that case it wouldn't be `router_conf.get('max_retry')`, but `router_conf[rel_channel].get('max_retry')`. That's what I meant by potentially clashing rel_names at the top level

jrconlin: ah, no, that's just a typo in the comment. It's per platform so in theory `router_conf[rel_channel].get('max_retry')`

autopush/router/apnsrouter.py
pjenvey: max_retry's still assumed to be top level here (alongside the rel_channel names) but it's per rel_channel in the conf. e.g. it would be `router_conf[<some_rel_channel>].get('max_retry')`

autopush/router/apns2.py
```diff
@@ -129,33 +133,40 @@ def send(self, router_token, payload, apns_id,
...
+                        response_body="APNS could not process "
+                                      "your message {}".format(reason),
+                        log_exception=False
+                    )
+                break
+            except HTTP20Error:
+                connection.close()
+                attempt += 1
+                if attempt < self._max_retry:
+                    continue
```
pjenvey: The _return_connection in the finally block's a problem now because we only call get_connection once (outside of the loop).

So in the case here of the loop continue we're potentially returning it multiple times (per each iteration).

Probably wanna set a flag here to disable returning in the finally block, or could even pull the loop body into a separate method

jrconlin: Crap, you're right. good catch!

jrconlin: And the proper fix is to move _get_connection inside of the loop since we need to pull a new one if it closes or fails.

Issue #1050, opened 2017-10-17
Convert Message / Router tables to use TTL


jrconlin:
AWS TTL will remove the table rotation requirement as well as reduce the number of write calls we'll need. 

depends on #1049

jrconlin:
blocked on #1081

jrconlin:
Expiry added to Message and Router tables as part of https://github.com/mozilla-services/autopush/pull/1061

PR #1061, opened 2017-10-25
feat: convert to use AWS boto3

### LINKED ISSUE #1049: Migrate to boto3
TTL support is only allowed via boto3. We make a number of complex requests and tests that limit the utility of doing parallel boto2/boto3 calls, so might as well bite the bullet.

jrconlin:
Issue #1050
Closes #1049

### DIFF
--- a/automock/boto.cfg
+++ b/automock/boto.cfg
@@ -1,9 +1,11 @@
-[Credentials]
-aws_access_key_id =
-aws_secret_access_key =
+[default]
 
 [Boto]
 is_secure = False
 https_validate_certificates = False
 proxy_port = 8000
 proxy = 127.0.0.1
+
+[DynamoDB]
+region=us-east-1
+validate_checksums=False
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -32,6 +32,7 @@ scheme was adopted.
 from __future__ import absolute_import
 
 import datetime
+import os
 import random
 import time
 import uuid
@@ -43,16 +44,16 @@ from attr import (
     attrib,
     Factory
 )
-from boto.exception import JSONResponseError
+
 from boto.dynamodb2.exceptions import (
-    ConditionalCheckFailedException,
     ItemNotFound,
-    ProvisionedThroughputExceededException,
 )
-from boto.dynamodb2.fields import HashKey, RangeKey, GlobalKeysOnlyIndex
-from boto.dynamodb2.layer1 import DynamoDBConnection
-from boto.dynamodb2.table import Table, Item
-from boto.dynamodb2.types import NUMBER
+import boto3
+import botocore
+from boto3.dynamodb.conditions import Key
+from boto3.exceptions import Boto3Error
+from botocore.exceptions import ClientError
+
 from typing import (  # noqa
     TYPE_CHECKING,
     Any,
@@ -72,6 +73,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.threads import deferToThread
 
 import autopush.metrics
+from autopush import MAX_EXPIRY
 from autopush.exceptions import AutopushException
 from autopush.metrics import IMetrics  # noqa
 from autopush.types import ItemLike  # noqa
@@ -91,6 +93,16 @@ T = TypeVar('T')  # noqa
 key_hash = ""
 TRACK_DB_CALLS = False
 DB_CALLS = []
+
+# See https://botocore.readthedocs.io/en/stable/reference/config.html for
+# additional config options
+g_dynamodb = boto3.resource(
+    'dynamodb',
+    config=botocore.config.Config(
+        region_name=os.getenv("AWS_REGION_NAME", "us-east-1")
+    )
+)
+g_client = g_dynamodb.meta.client
 
 
 def get_month(delta=0):
@@ -121,21 +133,6 @@ def hasher(uaid):
     return uaid
 
 
-def dump_uaid(uaid_data):
-    # type: (ItemLike) -> str
-    """Return a dict for a uaid.
-
-    This is utilized instead of repr since some db methods return a
-    DynamoDB Item which does not actually show its dict key/values
-    when dumped via repr.
-
-    """
-    if isinstance(uaid_data, Item):
-        return repr(uaid_data.items())
-    else:
-        return repr(uaid_data)
-
-
 def make_rotating_tablename(prefix, delta=0, date=None):
     # type: (str, int, Optional[datetime.date]) -> str
     """Creates a tablename for table rotation based on a prefix with a given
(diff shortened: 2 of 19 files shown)


autopush/db.py
```diff
@@ -649,6 +827,11 @@ def register_user(self, data):
                         r[key] = value
                 result = r
             return (True, result)
+        except ClientError as ex:
+            # ClientErrors are generated by a factory, and while they have a
+            # class, it's dynamically generated.
+            if 'ConditionalCheckFailedException' == ex.__class__.__name__:
```
pjenvey: Since they recommend checking against `[Error']['Code']` like you did above for ResourceNotFoundException, let's just standardize on that. The test can raise ClientError({'Error': {'Code': 'ConditionalCheckFailedException'}})

autopush/db.py
```diff
@@ -658,16 +841,19 @@ def drop_user(self, uaid):
         """Drops a user record"""
         # The following hack ensures that only uaids that exist and are
         # deleted return true.
-        huaid = hasher(uaid)
-        return self.table.delete_item(uaid=huaid,
-                                      expected={"uaid__eq": huaid})
+        try:
+            self.get_uaid(uaid)
```
pjenvey: Why add the get_uaid call here? It could be dangerous as get_uaid potentially calls drop_user

autopush/tests/__init__.py
```diff
@@ -18,17 +21,26 @@
...
         "-jar", ddb_jar, "-sharedDb", "-inMemory"
     ])
+    conf = botocore.config.Config(
+        region_name=os.getenv('AWS_REGION_NAME', 'us-east-1')
+    )
     ddb_process = subprocess.Popen(cmd, shell=True, env=os.environ)
+    autopush.db.DYNAMODB = boto3.resource(
+        'dynamodb',
+        config=conf,
+        endpoint_url=os.getenv("AWS_LOCAL_DYNAMODB", "http://127.0.0.1:8000")
```
pjenvey: I think we'll have to do something about credentials here (hardcode?) so tests run locally without needing to set them

automock/boto.cfg
```diff
@@ -1,9 +1,13 @@
-[Credentials]
```
pjenvey: Seems like we should kill this entire file now, boto3 only reads "Credentials" from a BOTO_CONFIG (which it considers an old boto2 config file).

Also if we're no longer able to configure our local dynamodb through this file our docker-compose setup will no longer work. I think that's a whole separate issue entirely from this PR, though

autopush/db.py
```diff
@@ -43,16 +44,17 @@
     attrib,
     Factory
 )
-from boto.exception import JSONResponseError
+
 from boto.dynamodb2.exceptions import (
```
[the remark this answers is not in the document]
jrconlin: IIUC, `ProvisionedThroughputExceededException` is in `botocore` and still could be returned. `ConditionalCheckFailedException` appears to be wrapped into `ClientError`,so we can probably remove it.

autopush/db.py
```diff
@@ -151,12 +161,53 @@ def create_rotating_message_table(prefix="message", delta=0, date=None,
...
+    try:
+        table.meta.client.update_time_to_live(
+            TableName=tablename,
+            TimeToLiveSpecification={
+                'Enabled': True,
+                'AttributeName': 'expry'
+            }
+        )
+    except ClientError as ex:  # pragma nocover
+        if "UnknownOperationException" in ex.message:
```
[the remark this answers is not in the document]
jrconlin: I dunno. We're eating an error. I'm not a huge fan of this either, but the other option is that we introduce a potentially complicated test environment check system (mock, env check, passed flag even for integration, etc.). The try block is limited to the one function and the significant return, so I think it may be OK to do this.

autopush/db.py
jrconlin: Pretty much for the reason noted. We only delete UAIDs that already exist. If get_uaid deletes the uaid, then it was already scheduled for deletion and returning False is acceptable. Either way, it's deleted.

autopush/db.py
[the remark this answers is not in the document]
pjenvey: Yea, probably not totally necessary. I think UnknownOperationException is a pretty clear case that we should not ever see in prod. It does show up as the exception "type" (`response['Error']['Code'] == 'UnknownOperationException'`) so we can check that a little more explicitly here at least

autopush/db.py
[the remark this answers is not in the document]
pjenvey: Nope, it looks like even ProvisionedThroughputExceededException becomes a ClientError (or one of those dynamic subclass of it) now. This boto2 one is defined in that boto.dynamodb2.exceptions module which boto3/botocore don't utilize at all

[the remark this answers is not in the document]
pjenvey: e.g. https://github.com/boto/boto3/issues/597#issuecomment-323982159

automock/boto.cfg
jrconlin: Yeah, going to hardcode the credentials, since they're bogus anyway and we only need them for the test.

autopush/utils.py
```diff
@@ -301,6 +302,8 @@ class WebPushNotification(object):
     # Whether this notification should follow legacy non-topic rules
     legacy = attrib(default=False)  # type: bool
 
+    expry = attrib(default=MAX_EXPRY)  # type: int
```
pjenvey: Is this expry attribute here even necessary? We seem to always fill the expry headed towards the database from the ttl attribute. Only thing I see this doing is propagating it over the connection node via JSON, but it's not even used there

jrconlin: Actually, going to leave it. We use it for several tests and it identifies as a keyword parameter.

autopush/db.py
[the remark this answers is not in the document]
jrconlin: Revised the errors to be ClientErrors and added handling for the special cases.

autopush/db.py
```diff
@@ -256,6 +363,7 @@ def preflight_check(message, router, uaid="deadbeef00000000deadbeef00000000"):
         update_id=message_id,
         message_id=message_id,
         ttl=60,
+        expry=_expry(60),
```
bbangert: Is there some reason we can't include the i to make it a real word? It is supposed to be expiry, right?

autopush/main_argparse.py
```diff
@@ -115,7 +115,7 @@ def _obsolete_args(parser):
     parser.add_argument('--external_router', help='OBSOLETE')
     parser.add_argument('--max_message_size', type=int, help="OBSOLETE")
     parser.add_argument('--s3_bucket', help='OBSOLETE')
-    parser.add_argument('--senderid_expry', help='OBSOLETE')
+    parser.add_argument('--senderid_expiry', help='OBSOLETE')
```
pjenvey: ha

autopush/utils.py
```diff
@@ -335,6 +335,12 @@ def generate_message_id(self, fernet):
         self.update_id = self.message_id
         return self.message_id
 
+    @classmethod
+    def calc_expiry(cls, ttl=0):
```
pjenvey: this is now dead code: kill

autopush/db.py
pjenvey: This can turn into infinite recursion w/ an invalid record, though. See the case in test_db.test_incomplete_uaid -- if you disable the mock of drop_user in there it'll blow up the stack

autopush/db.py
```diff
@@ -92,6 +94,16 @@
 TRACK_DB_CALLS = False
 DB_CALLS = []
 
+# See https://botocore.readthedocs.io/en/stable/reference/config.html for
+# additional config options
+g_dynamodb = boto3.resource(
```
bbangert: What does the g_ mean?

autopush/db.py
```diff
@@ -525,11 +655,15 @@ def fetch_timestamp_messages(
         else:
             sortkey = "01;"
 
-        results = list(self.table.query_2(uaid__eq=hasher(uaid.hex),
-                                          chidmessageid__gt=sortkey,
-                                          consistent=True, limit=limit))
+        response = self.table.query(
+            KeyConditionExpression=Key('uaid').eq(hasher(uaid.hex))
+            & Key('chidmessageid').gt(sortkey),
```
bbangert: nit: maybe indent so only the keys line up.

autopush/db.py
jrconlin: Well, i wanted to make sure that there was no potential for symbol conflict, and I picked "g_" because it's effectively a global.

autopush/db.py
jrconlin: flake8 complains about that, but I can move the & to the previous line.

pjenvey: can also wrap the whole statement in parens if that helps

PR #1065, opened 2017-10-27
fix: drop users with no router_type recorded

### LINKED ISSUE #1058: Require a router_type to validate a record in the endpoint
Per https://sentry.prod.mozaws.net/operations/autopush-prod/issues/671942/ we should require a router_type in the record or drop it.

Traceback:
```
KeyError: None
  File "twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/web/base.py", line 222, in _validation_err
    fail.trap(InvalidRequest)
  File "twisted/python/failure.py", line 359, in trap
    self.raiseException()
  File "twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/web/base.py", line 258, in _boto_err
    fail.trap(BotoServerError)
  File "twisted/python/failure.py", line 359, in trap
    self.raiseException()
  File "twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/web/base.py", line 250, in _overload_err
    fail.trap(ProvisionedThroughputExceededException)
  File "twisted/python/failure.py", line 359, in trap
    self.raiseException()
  File "twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/web/base.py", line 76, in _call_func
    return func(request_handler, **output_kwargs)
  File "autopush/web/webpush.py", line 496, in post
    router = self.routers[router_type]
```

bbangert:
Closes #1058

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -54,6 +54,8 @@ MAX_TTL = 60 * 60 * 24 * 60
 # Base64 URL validation
 VALID_BASE64_URL = re.compile(r'^[0-9A-Za-z\-_]+=*$')
 
+VALID_ROUTER_TYPES = ["simplepush", "webpush", "gcm", "fcm", "apns"]
+
 
 class WebPushSubscriptionSchema(Schema):
     uaid = fields.UUID(required=True)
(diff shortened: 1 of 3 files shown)


autopush/web/webpush.py
```diff
@@ -54,6 +54,8 @@
 # Base64 URL validation
 VALID_BASE64_URL = re.compile(r'^[0-9A-Za-z\-_]+=*$')
 
+VALID_ROUTER_TYPES = ["simplepush", "webpush", "gcm", "fcm", "apns"]
```
jrconlin: Really thought I had something like this before. There's a similar list in `autopush.registration.conditional_token_check` Wonder if it might make sense to have this list somewhere common and use it in both locations?

bbangert: Yea, or heck, maybe a constants.py for the shared universal constants we end up using all over.

jrconlin: absolutely 👍 to a constants.py. I've got one for boto I put in `__init__.py`, but happy to move it in favor of just keeping `version` there.

bbangert: Will open a separate PR to refactor some of the constants together.

Issue #1057, opened 2017-10-23
Update GCM endpoint/library


jrconlin:
GCM delivery appears to be failing for Gecko 58. In addition, it looks like the GCM endpoint may have moved and no library currently points to the new location.

from `android.googleapis.com` to `gcm-http.googleapis.com` (see https://developers.google.com/cloud-messaging/http-server-ref)

PR #1068, opened 2017-11-06
feat: assimilate the gcm client code

jrconlin:
Closes: #1057

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -1,13 +1,13 @@
 """GCM Router"""
 from typing import Any  # noqa
 
-import gcmclient
 from requests.exceptions import ConnectionError
 from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
 
 from autopush.exceptions import RouterException
 from autopush.metrics import make_tags
+from autopush.router import gcmclient
 from autopush.router.interface import RouterResponse
 from autopush.types import JSONDict  # noqa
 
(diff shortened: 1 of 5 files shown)


autopush/router/gcm.py
```diff
@@ -115,19 +115,24 @@ def _route(self, notification, uaid_data):
         try:
             gcm = self.gcm[router_data['creds']['senderID']]
             result = gcm.send(payload)
+        except RouterException:
```
pjenvey: these 2 lines are redundant

autopush/router/gcmclient.py
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        response = self._sender(
+            url=self._endpoint,
+            headers=headers,
+            data=json.dumps(payload.payload),
+            **self._options
+        )
+
+        if response.status_code == 400:
```
pjenvey: nitpick: could collapse the next section with this one `response.status_code in (400, 404)`

autopush/tests/test_router.py
```diff
@@ -296,9 +296,177 @@ def check_results(result):
         return d
 
 
+class GCMClientTestCase(unittest.TestCase):
```
pjenvey: Could we put this TestCase in a new file? test_router's already pretty hefty

autopush/router/gcm.py
jrconlin: No, they're not. There's a general Exception handler further down that transposes unexpected exceptions into RouterExceptions with logging.

autopush/tests/test_router.py
jrconlin: 👍

Issue #1054, opened 2017-10-19
Add statsd metric reporting to Rust version


bbangert:
The Rust version needs to emit statsd metrics to Datadog in the same manner as we do for the Python code. It should also emit the same metrics as we emit right now.

PR #1069, opened 2017-11-07
feat: fill in missing metrics for autopush_rs

pjenvey:
the rust side's metrics (via cadence) currently lack the ability to
pass datadog tags, currently ua.{message_data,connection.lifespan}

Issue: #1054

### DIFF
--- a/autopush/webpush_server.py
+++ b/autopush/webpush_server.py
@@ -165,6 +165,7 @@ class HelloResponse(OutputCommand):
     message_month = attrib()  # type: str
     check_storage = attrib()  # type: bool
     reset_uaid = attrib()  # type: bool
+    connected_at = attrib()  # type: int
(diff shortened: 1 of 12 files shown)


autopush/webpush_server.py
```diff
@@ -380,13 +382,13 @@ def lookup_user(self, hello):
         # All records must have a router_type and connected_at, in some odd
         # cases a record exists for some users without it
         if "router_type" not in record or "connected_at" not in record:
-            self.db.router.drop_user(uaid)
+            self.drop_user(uaid, record, 104)
```
jrconlin: We send the entire uaid record out to logging?

pjenvey: yea, this is what websocket and apparently web.webpush also do. doesn't seem necessary but at least they're all log.debug's

PR #1072, opened 2017-11-08
feat: enforce AWS expiry on data lookups

### LINKED ISSUE #1051: Transition to new router table that uses TTL instead of rotation
Convert autopush to use a TTL based router table to reduce DB writes. 
requires #1049

jrconlin:
Closes #1051

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -523,6 +523,7 @@ class Message(object):
         # Note: This only returns the chids associated with the UAID.
         # Functions that call store_message() would be required to
         # update that list as well using register_channel()
+        # TODO: once expiry is properly integrated, use FilterExpression
         result = self.table.get_item(
             Key={
                 'uaid': hasher(uaid),
@@ -533,6 +534,9 @@ class Message(object):
         if result['ResponseMetadata']['HTTPStatusCode'] != 200:
             return False, set([])
         if 'Item' not in result:
+            return False, set([])
+        now = int(time.time())
+        if result['Item'].get('expiry', now) < now:
             return False, set([])
         return True, result['Item'].get("chids", set([]))
 
@@ -609,6 +613,7 @@ class Message(object):
 
         """
         # Eagerly fetches all results in the result set.
+        # TODO: once expiry is properly integrated, use FilterExpression
         response = self.table.query(
             KeyConditionExpression=(Key("uaid").eq(hasher(uaid.hex))
                                     & Key('chidmessageid').lt('02')),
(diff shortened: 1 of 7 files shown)


autopush/db.py
```diff
@@ -654,16 +659,20 @@ def fetch_timestamp_messages(
...
             Limit=limit
         )
+        now = int(time.time())
         notifs = [
             WebPushNotification.from_message_table(uaid, x) for x in
-            response.get("Items")
+            response.get("Items") if x.get('expiry', now) >= now
         ]
+        if response.get("Items") and not notifs:
+            print "here"
```
bbangert: Huh?

autopush/db.py
```diff
@@ -523,6 +523,7 @@ def all_channels(self, uaid):
         # Note: This only returns the chids associated with the UAID.
         # Functions that call store_message() would be required to
         # update that list as well using register_channel()
+        # TODO: once expiry is properly integrated, use FilterExpression
```
bbangert: What does that mean?

autopush/router/gcm.py
```diff
@@ -115,6 +115,8 @@ def _route(self, notification, uaid_data):
         try:
             gcm = self.gcm[router_data['creds']['senderID']]
             result = gcm.send(payload)
+        except RouterException:
+            raise  # pragma nocover
```
bbangert: Did this somehow get dropped?

autopush/db.py
jrconlin: Ideally, we want to use FilterExpression in our calls to DynamoDB. Unfortunately, while really efficient, it doesn't handle our need for backwards compatibility. We have some records that don't have an expiry yet set, so those will not be fetched.

Once the expiry has been put in place for several months, and we can safely assume most current records have it, then we can remove the code I've put in place to handle records that do not have an expiry set.

autopush/db.py
jrconlin: crap, debug. sorry.

autopush/router/gcm.py
jrconlin: This may be a merge error.

bbangert: Maybe we should have a test that triggers this to avoid future merge errors?

autopush/db.py
bbangert: Ah, perhaps filing an issue for that now, and linking it here would help. We could then mark the issue blocked so we can remember when we should remove it.

autopush/tests/test_integration.py
```diff
@@ -1005,7 +1005,7 @@ def test_ttl_batch_expired_and_good_one(self):
         data2 = str(uuid.uuid4())
         client = yield self.quick_register()
         yield client.disconnect()
-        for x in range(0, 12):
```
pjenvey: was this change necessary here? IIRC the total messages churned through in these batch tests was > 10 to thoroughly test the fetch_messages call with its default limit=10

jrconlin: yes, the problem was that we would load up with the 10 expired messages and never get to the one that wasn't expired. That's highlighted now that we're actually expiring messages. The next test checks batching.

autopush/db.py
jrconlin: 👍 https://github.com/mozilla-services/autopush/issues/1073

bbangert: Oh, I meant referring to the issue in the code itself.

jrconlin:
Design assumption flawed. closing.

Issue #1048, opened 2017-10-17
Catch H2 ssl error from APNS


bbangert:
APNS can occasionally trigger an uncaught ssl exception:
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/667854/

Traceback:
```
SSLError: [SSL: BAD_WRITE_RETRY] bad write retry
  File "twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/web/base.py", line 292, in _router_fail_err
    fail.trap(RouterException)
  File "twisted/python/failure.py", line 346, in trap
    self.raiseException()
  File "twisted/python/threadpool.py", line 250, in inContext
    result = inContext.theWork()
  File "twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
  File "autopush/router/apnsrouter.py", line 144, in _route
    apns_id=apns_id)
  File "autopush/router/apns2.py", line 137, in send
    'POST', url=url, body=body, headers=headers)
  File "hyper/http20/connection.py", line 281, in request
    self.endheaders(message_body=body, final=True, stream_id=stream_id)
  File "hyper/http20/connection.py", line 555, in endheaders
    stream.send_headers(headers_only)
  File "hyper/http20/stream.py", line 99, in send_headers
    self._send_outstanding_data()
  File "hyper/http20/connection.py", line 466, in _send_outstanding_data
    self._send_cb(data, tolerate_peer_gone=tolerate_peer_gone)
  File "hyper/http20/connection.py", line 622, in _send_cb
    self._sock.sendall(data)
  File "2.7/ssl.py", line 751, in sendall
    v = self.send(data[count:])
  File "2.7/ssl.py", line 717, in send
    v = self._sslobj.write(data)
```

We should catch this since there's nothing that we can do about it.

pjenvey:
Marking as p2: we'll see if #1056 affects this

pjenvey:
[...]

PR #1078, opened 2017-11-15
fix: handle an obscure SSLError seen w/ apns

pjenvey:
not clear why this one occurs so metric it for now vs bothering sentry

Issue: #1048


pjenvey:
(fixed coverage)

PR #1085, opened 2017-11-25
feat: use structured logging and log session info out

### LINKED ISSUE #1055: Ensure Rust log output fully meets production format
The Rust log output is JSON, but not the exact format needed for production. It needs to match production logging output.

bbangert:
Closes #1055

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -11,10 +11,12 @@ use cadence::prelude::*;
 use futures::AsyncSink;
 use futures::future::Either;
 use futures::sync::mpsc;
+use futures::sync::oneshot::Receiver;
 use futures::{Stream, Sink, Future, Poll, Async};
 use tokio_core::reactor::Timeout;
 use time;
 use uuid::Uuid;
+use woothee::parser::{Parser, WootheeResult};
 
 use call;
 use errors::*;
@@ -24,6 +26,26 @@ use server::Server;
 pub struct RegisteredClient {
     pub uaid: Uuid,
     pub tx: mpsc::UnboundedSender<ServerNotification>,
+}
+
+// Websocket session statistics
+#[derive(Clone)]
+struct SessionStatistics {
+    // User data
+    uaid: String,
+    uaid_reset: bool,
+    existing_uaid: bool,
+    connection_type: String,
+    host: String,
+
+    // Usage data
+    direct_acked: i32,
+    direct_storage: i32,
+    stored_retrieved: i32,
+    stored_acked: i32,
+    nacks: i32,
+    unregisters: i32,
+    registers: i32,
 }
 
 // Represents a websocket client connection that may or may not be authenticated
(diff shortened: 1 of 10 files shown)


autopush_rs/src/client.rs
```diff
@@ -128,15 +153,26 @@ where
     /// the various state behind the server. This provides transitive access to
     /// various configuration options of the server as well as the ability to
     /// call back into Python.
-    pub fn new(ws: T, srv: &Rc<Server>) -> Client<T> {
+    pub fn new(ws: T, srv: &Rc<Server>, uarx: Receiver<Result<String>>) -> Client<T> {
         let srv = srv.clone();
         let timeout = Timeout::new(srv.opts.open_handshake_timeout.unwrap(), &srv.handle).unwrap();
+
+        // Pull out the user-agent, which we should have by now
+        let uastr = match uarx.wait() {
```
alexcrichton: In general calling `wait` can cause problems if it accidentally blocks, perhaps this could switch to `poll` which is nonblocking to assert that no blocking happens but it's still ready?

autopush_rs/src/client.rs
```diff
@@ -489,15 +527,30 @@ where
...
+            flags,
+            rx,
+            message_month,
             unacked_direct_notifs: Vec::new(),
             unacked_stored_notifs: Vec::new(),
             unacked_stored_highest: None,
-            connected_at: connected_at,
+            connected_at,
+            stats: SessionStatistics {
+                uaid: String::from(uaid.hyphenated().to_string()),
```
alexcrichton: This is actually just `String::from(String)` so I think the `from` can be left off

autopush_rs/src/util/autojson.rs
```diff
@@ -0,0 +1,287 @@
+// {{{ Crate docs
```
pjenvey: will all of this file go away w/ the slog enhancements?

bbangert: Yup.

autopush_rs/src/util/mod.rs
```diff
@@ -41,52 +44,50 @@ where
...
+                "Type" => "autopush_rs:log",
+                "EnvVersion" => "2.0",
+                "Logger" => format!("AutopushRust-{}", env!("CARGO_PKG_VERSION")),
+            ))
+            .build()
+            .fuse();
+        let drain = slog_async::Async::new(drain).build().fuse();
+        let logger = slog::Logger::root(drain, o!());
+        let _scope_guard = slog_scope::set_global_logger(logger);
+        let _log_guard = match slog_stdlog::init() {
```
pjenvey: what's with _log_guard? it's always () -- init returns an empty Result https://docs.rs/slog-stdlog/1.1.0/slog_stdlog/fn.init.html

pjenvey: You might want to propagate the Error from slog init (have init_logging return a Result) then the init_logging caller can expect() on it, like it's doing with all the other config values

bbangert: The stdlog init only errors out if its called twice due to a global it has. This only happens with our tests, so I ignored the error so our tests pass.

bbangert: Ok, ate the result.

PR #1086, opened 2017-11-28
bug: make boto3 calls thread safe

### LINKED ISSUE #1081: Convert boto3 calls to thread safe version
boto3 is not thread safe, because why would it be?

jrconlin:
Closes #1081

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -34,6 +34,7 @@ from __future__ import absolute_import
 import datetime
 import os
 import random
+import threading
 import time
 import uuid
 from functools import wraps
@@ -42,14 +43,16 @@ from attr import (
     asdict,
     attrs,
     attrib,
-    Factory
-)
+    Factory,
+    Attribute)
 
 from boto.dynamodb2.exceptions import (
     ItemNotFound,
 )
 import boto3
 import botocore
+from boto.dynamodb2.table import Table  # noqa
+from boto3.resources.base import ServiceResource  # noqa
 from boto3.dynamodb.conditions import Key
 from boto3.exceptions import Boto3Error
 from botocore.exceptions import ClientError
@@ -73,6 +76,7 @@ from twisted.internet.defer import inlineCallbacks, returnValue
 from twisted.internet.threads import deferToThread
 
 import autopush.metrics
+from autopush import constants
 from autopush.exceptions import AutopushException
 from autopush.metrics import IMetrics  # noqa
 from autopush.types import ItemLike  # noqa
@@ -96,16 +100,7 @@ key_hash = ""
 TRACK_DB_CALLS = False
 DB_CALLS = []
 
-# See https://botocore.readthedocs.io/en/stable/reference/config.html for
-# additional config options
-g_dynamodb = boto3.resource(
-    'dynamodb',
-    config=botocore.config.Config(
-        region_name=os.getenv("AWS_REGION_NAME", "us-east-1")
-    ),
-    endpoint_url=os.getenv("AWS_LOCAL_DYNAMODB")
-)
-g_client = g_dynamodb.meta.client
+MAX_DDB_SESSIONS = constants.THREAD_POOL_SIZE
 
 
 def get_month(delta=0):
@@ -145,22 +140,25 @@ def make_rotating_tablename(prefix, delta=0, date=None):
     return "{}_{:04d}_{:02d}".format(prefix, date.year, date.month)
 
 
-def create_rotating_message_table(prefix="message", delta=0, date=None,
-                                  read_throughput=5,
-                                  write_throughput=5):
-    # type: (str, int, Optional[datetime.date], int, int) -> Table
+def create_rotating_message_table(
+        prefix="message",     # type: str
+        delta=0,              # type: int
+        date=None,            # type: Optional[datetime.date]]
+        read_throughput=5,    # type: int
+        write_throughput=5,   # type: int
+        boto_resource=None):  # type: DynamoDBResource
+    # type: (...) -> Table  # noqa
     """Create a new message table for webpush style message storage"""
     tablename = make_rotating_tablename(prefix, delta, date)
 
     try:
-        table = g_dynamodb.Table(tablename)
-        if table.table_status == 'ACTIVE':  # pragma nocover
+        table = boto_resource.Table(tablename)
+        if table.table_status == 'ACTIVE':
             return table
     except ClientError as ex:
         if ex.response['Error']['Code'] != 'ResourceNotFoundException':
-            # If we hit this, our boto3 is misconfigured and we need to bail.
-            raise ex  # pragma nocover
-    table = g_dynamodb.create_table(
+            raise  # pragma nocover
+    table = boto_resource.create_table(
         TableName=tablename,
         KeySchema=[
             {
@@ -201,25 +199,32 @@ def create_rotating_message_table(prefix="message", delta=0, date=None,
     return table
 
 
-def get_rotating_message_table(prefix="message", delta=0, date=None,
-                               message_read_throughput=5,
-                               message_write_throughput=5):
-    # type: (str, int, Optional[datetime.date], int, int) -> Table
+def get_rotating_message_tablename(
+        prefix="message",            # type: str
+        delta=0,                     # type: int
+        date=None,                   # type: Optional[datetime.date]
+        message_read_throughput=5,   # type: int
+        message_write_throughput=5,  # type: int
+        boto_resource=None):         # type: DynamoDBResource
+    # type: (...) -> str  # noqa
     """Gets the message table for the current month."""
     tablename = make_rotating_tablename(prefix, delta, date)
-    if not table_exists(tablename):
-        return create_rotating_message_table(
+    if not table_exists(tablename, boto_resource=boto_resource):
+        create_rotating_message_table(
             prefix=prefix, delta=delta, date=date,
             read_throughput=message_read_throughput,
             write_throughput=message_write_throughput,
+            boto_resource=boto_resource
         )
+        return tablename
     else:
-        return g_dynamodb.Table(tablename)
+        return tablename
 
 
 def create_router_table(tablename="router", read_throughput=5,
-                        write_throughput=5):
-    # type: (str, int, int) -> Table
+                        write_throughput=5,
+                        boto_resource=None):
+    # type: (str, int, int, DynamoDBResource) -> Table
     """Create a new router table
 
     The last_connect index is a value used to determine the last month a user
@@ -233,8 +238,7 @@ def create_router_table(tablename="router", read_throughput=5,
     cost of additional queries during GC to locate expired users.
 
     """
-
-    table = g_dynamodb.create_table(
+    table = boto_resource.create_table(
         TableName=tablename,
         KeySchema=[
             {
@@ -293,20 +297,30 @@ def create_router_table(tablename="router", read_throughput=5,
     return table
 
 
-def _drop_table(tablename):
+def _drop_table(tablename, boto_resource):
+    # type: (str, DynamoDBResource) -> None
     try:
-        g_client.delete_table(TableName=tablename)
+        boto_resource.meta.client.delete_table(TableName=tablename)
     except ClientError:  # pragma nocover
         pass
 
 
-def _make_table(table_func, tablename, read_throughput, write_throughput):
-    # type: (Callable[[str, int, int], Table], str, int, int) -> Table
+def _make_table(
+        table_func,        # type: Callable[[str, int, int, ServiceResource]]
+        tablename,         # type: str
+        read_throughput,   # type: int
+        write_throughput,  # type: int
+        boto_resource      # type: DynamoDBResource
+        ):
+    # type (...) -> DynamoDBTable
     """Private common function to make a table with a table func"""
-    if not table_exists(tablename):
-        return table_func(tablename, read_throughput, write_throughput)
+    if not boto_resource:
+        raise AutopushException("No boto3 resource provided for _make_table")
+    if not table_exists(tablename, boto_resource):
+        return table_func(tablename, read_throughput, write_throughput,
+                          boto_resource)
     else:
-        return g_dynamodb.Table(tablename)
+        return DynamoDBTable(boto_resource, tablename)
 
 
 def _expiry(ttl):
@@ -314,8 +328,8 @@ def _expiry(ttl):
 
 
 def get_router_table(tablename="router", read_throughput=5,
-                     write_throughput=5):
-    # type: (str, int, int) -> Table
+                     write_throughput=5, boto_resource=None):
+    # type: (str, int, int, DynamoDBResource) -> Table
     """Get the main router table object
 
     Creates the table if it doesn't already exist, otherwise returns the
@@ -323,7 +337,7 @@ def get_router_table(tablename="router", read_throughput=5,
 
     """
     return _make_table(create_router_table, tablename, read_throughput,
-                       write_throughput)
+                       write_throughput, boto_resource=boto_resource)
 
 
 def preflight_check(message, router, uaid="deadbeef00000000deadbeef00000000"):
@@ -355,7 +369,6 @@ def preflight_check(message, router, uaid="deadbeef00000000deadbeef00000000"):
         message_id=message_id,
         ttl=60,
     )
-
     # Store a notification, fetch it, delete it
     message.store_message(notif)
     assert message.delete_message(notif)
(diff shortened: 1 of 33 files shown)


autopush/tests/support.py
```diff
@@ -44,5 +45,6 @@ def test_db(metrics=None):
         router_conf=DDBTableConfig(tablename='router'),
         router=Mock(spec=Router),
         message_conf=DDBTableConfig(tablename='message'),
-        metrics=SinkMetrics() if metrics is None else metrics
+        metrics=SinkMetrics() if metrics is None else metrics,
+        sessions=autopush.tests.sessions,
```
jrconlin: for unknown reasons, nose doesn't import `sessions` from `autopush.tests` correctly (FWIW, pytest does). Thus the fully qualified path, which does resolve to the correct global pool.

autopush/db.py
```diff
@@ -96,15 +98,30 @@
...
+SESSION_ARGS = {}
+MAX_SESSIONS = 10
+
+
+class Sessions(object):
+
+    def __init__(self, conf=SESSION_ARGS):
+        self.pool = deque(maxlen=MAX_SESSIONS)
+        session = boto3.session.Session()
+        if not conf.get("endpoint_url") and os.getenv("AWS_LOCAL_DYNAMODB"):
```
bbangert:
```
"endpoint_url" not in conf
```

autopush/db.py
```diff
@@ -96,15 +98,30 @@
...
+        self.pool.extendleft((session.resource(
+            'dynamodb',
+            config=botocore.config.Config(
+                region_name=os.getenv("AWS_REGION_NAME", "us-east-1")
+            ),
+            **conf
+            ) for x in range(0, MAX_SESSIONS)))
+
+    def fetch(self):
+        return self.pool.pop()
```
bbangert: What happens if it runs out? Should it make another?

autopush/db.py
```diff
@@ -96,15 +98,30 @@
...
-g_client = g_dynamodb.meta.client
+SESSION_ARGS = {}
+MAX_SESSIONS = 10
+
+
+class Sessions(object):
+
+    def __init__(self, conf=SESSION_ARGS):
+        self.pool = deque(maxlen=MAX_SESSIONS)
+        session = boto3.session.Session()
```
bbangert: boto_session might avoid pain in the future as we track various 'session' concepts around.

autopush/db.py
```diff
@@ -146,20 +163,21 @@ def make_rotating_tablename(prefix, delta=0, date=None):
 
 def create_rotating_message_table(prefix="message", delta=0, date=None,
                                   read_throughput=5,
-                                  write_throughput=5):
-    # type: (str, int, Optional[datetime.date], int, int) -> Table
+                                  write_throughput=5,
+                                  session=None):
```
bbangert: Above we have 'session' which is a boto session.Session, yet here its a session.resource which is not the same. Here it's actually a DynamoDB resource created with the boto session.Session.

bbangert:
Why not make a thread-local module global we initialize and use (sorta like the boto2 one) instead of passing it all over?

autopush/db.py
jrconlin: It normally shouldn't be. If it's not specified, boto3 will use it's own, else it will use whatever you note. (Passing `None` is problematic) We do pass an `endpoint_url` as part of testing.

autopush/db.py
```diff
@@ -96,15 +98,30 @@
...
+
+
+class Sessions(object):
+
+    def __init__(self, conf=SESSION_ARGS):
+        self.pool = deque(maxlen=MAX_SESSIONS)
+        session = boto3.session.Session()
+        if not conf.get("endpoint_url") and os.getenv("AWS_LOCAL_DYNAMODB"):
+            conf["endpoint_url"] = os.getenv("AWS_LOCAL_DYNAMODB")
+        self.pool.extendleft((session.resource(
```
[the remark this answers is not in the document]
jrconlin: Heh, copied what we do for apns2. (and @pjenvey usually yells at me to use list comprehension)

I'm fine either way.

jrconlin:
we tend to make a number of thread calls when we defer for the database. the boto_session is tied to an instance rather than a thread where possible to ensure that meta information remains as consistent as possible.

autopush/db.py
```diff
@@ -650,65 +741,97 @@ def fetch_timestamp_messages(
...
                         :class:`autopush.metrics.IMetrics` interface.
 
         """
-        self.table = table
+        self.conf = conf
         self.metrics = metrics
         self._max_ttl = max_ttl
+        self._session = None
+        self._cached_table = None
+        self._sessions = boto_sessions or BotoSessions()
```
pjenvey: I think boto_sessions is always specified?

autopush/main.py
```diff
@@ -148,7 +148,8 @@ def main(cls, args=None, use_files=True):
             firehose_delivery_stream=ns.firehose_stream_name
         )
         try:
-            app = cls.from_argparse(ns)
+            cls.argparse = cls.from_argparse(ns, sessions=sessions)
```
pjenvey: is this assignment needed?

tox.ini
```diff
@@ -1,18 +1,19 @@
 [tox]
-envlist = py27,pypy,flake8,py36-mypy
+#envlist = py27,pypy,flake8,py36-mypy
+envlist = pypy,flake8
```
pjenvey: !

autopush/webpush_server.py
```diff
@@ -630,9 +677,14 @@ def process(self,
         if not valid:
             return UnregisterErrorResponse(error_msg=msg)
 
-        message = self.db.message_tables[command.message_month]
+        session = self.db.sessions.fetch()
```
pjenvey: session unused here

autopush/webpush_server.py
```diff
@@ -493,46 +516,66 @@ class DeleteMessageCommand(ProcessorCommand):
...
+                          boto_sessions=self.db.sessions)
         message.delete_message(notif)
         return DeleteMessageResponse()
 
 
 class DropUserCommand(ProcessorCommand):
     def process(self, command):
         # type: (DropUser) -> DropUserResponse
-        self.db.router.drop_user(command.uaid.hex)
+        session = self.db.sessions.fetch()
```
pjenvey: unused

autopush/main.py
```diff
@@ -63,11 +63,11 @@ class AutopushMultiService(MultiService):
 
     THREAD_POOL_SIZE = 50
 
-    def __init__(self, conf):
+    def __init__(self, conf, sessions=None):
```
pjenvey: nitpicky: ideally I think DatabaseManager being more responsible for creating the BotoSessions (maybe taking a config for it that includes the endpoint_url, say for tests) would avoid the hooks here. A lot like it is responsible for the Table objects, but also allowing overrides of those objects for tests.

Unfortunately our tests around main'ish functions are currently kind of crappy and would probably require some annoying monkey patches to force the right test configuration

pjenvey:
so we definitely need to rename 'session' being that it's a resource. BotoSessions doesn't seem appropriate either. My other main issue with the naming in the current patch is "boto_session" vs "boto_sessions" seemed like a typo waiting to happen

pjenvey:
also not necessary for now but the 'session' fetch/release screams for a context manager. I imagine one that could optionally handle being specified an existing session and handling the 'release = False' noop behavior for the caller

tox.ini
jrconlin: crap. sorry, that's because I was trying to force test against pypy.

autopush/db.py
```diff
@@ -96,15 +98,42 @@
...
-# additional config options
-g_dynamodb = boto3.resource(
-    'dynamodb',
-    config=botocore.config.Config(
-        region_name=os.getenv("AWS_REGION_NAME", "us-east-1")
-    )
-)
-g_client = g_dynamodb.meta.client
+SESSION_ARGS = {}
+MAX_SESSIONS = 10
```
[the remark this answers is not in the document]
jrconlin: Oh, good call!

jrconlin:
I agree about the context manager. I mostly just wanted to get this landed and skipping the context manager made it easier for me to follow what was going on. It also let me control very tightly exactly when we fetch and release a resource handle, which was useful during debugging.

autopush/webpush_server.py
```diff
@@ -360,7 +362,12 @@ def process(self, hello):
 
         # Save the UAID as register_user removes it
         uaid = user_item["uaid"]  # type: str
-        success, _ = self.db.router.register_user(user_item)
+        session = self.db.resources.fetch()
```
bbangert: session -> ddb_resource

autopush/websocket.py
```diff
@@ -1047,17 +1086,23 @@ def _monthly_transition(self):
 
         """
         # Get the current channels for this month
-        _, channels = self.ps.message.all_channels(self.ps.uaid)
+        message = self.db.message_table(self.ps.message_month)
+        _, channels = message.all_channels(self.ps.uaid)
```
bbangert: How come these commands don't need a ddb_resource?

jrconlin: `self.db.message_table` automatically uses the resource pool already assigned in `db`. Only stand alone functions require a specific resource handle to be passed in.

autopush/websocket.py
```diff
@@ -249,7 +249,7 @@ def __attrs_post_init__(self):
     def message(self):
         # type: () -> Message
         """Property to access the currently used message table"""
-        return self.db.message_tables[self.message_month]
+        return Message(self.message_month, resource_pool=self.db.resources)
```
jrconlin: No idea why coverage doesn't see this. Maybe because it's a property?

autopush/db.py
```diff
@@ -293,28 +342,30 @@ def create_router_table(tablename="router", read_throughput=5,
...
 
 
 def _expiry(ttl):
     return int(time.time() + ttl)
 
 
 def get_router_table(tablename="router", read_throughput=5,
-                     write_throughput=5):
+                     write_throughput=5, boto_resource=None):
     # type: (str, int, int) -> Table
```
bbangert: Type sig change.

autopush/db.py
```diff
@@ -337,7 +389,7 @@ def preflight_check(message, router, uaid="deadbeef00000000deadbeef00000000"):
     # Verify tables are ready for use if they just got created
     ready = False
     while not ready:
-        tbl_status = [x.table_status() for x in [message, router]]
+        tbl_status = [x.table_status()for x in [message, router]]
```
bbangert: Lost the space after status().

autopush/db.py
```diff
@@ -694,24 +748,37 @@ def update_last_message_read(self, uaid, timestamp):
...
                         :class:`autopush.metrics.IMetrics` interface.
+        :param resource_pool: Pool of :class:`autopush.db.BotoResources`
 
         """
-        self.table = table
+        self.conf = conf
         self.metrics = metrics
         self._max_ttl = max_ttl
+        self._cached_table = None
+        self._resources = resource_pool or BotoResources()
```
bbangert: resource_pool is required, so this shouldn't happen?

autopush/db.py
```diff
@@ -694,24 +748,37 @@ def update_last_message_read(self, uaid, timestamp):
...
+        self._resource = None
```
bbangert: What is this for?

autopush/db.py
```diff
@@ -694,24 +748,37 @@ def update_last_message_read(self, uaid, timestamp):
...
+                table = get_router_table(boto_resource=boto_resource)
+        return table
 
     def table_status(self):
-        return self.table.table_status
+        return self.table().table_status
 
     def get_uaid(self, uaid):
-        # type: (str) -> Item
+        # type: (str, ServiceResource) -> Item
```
bbangert: Type sig mismatch.

autopush/db.py
```diff
@@ -851,6 +920,7 @@ def drop_old_users(self, months_ago=2):
             quickly as possible.
 
         :param months_ago: how many months ago since the last connect
+        :param boto_resource: ServiceResource for thread
```
bbangert: Param doesn't exist here.

autopush/db.py
```diff
@@ -96,16 +100,58 @@
...
+            del(conf["endpoint_url"])
+        self.conf = conf
+        self.ddb_data = threading.local()
+
+    def __enter__(self):
+        self._resource = self.fetch()
+        return self._resource
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.release(self._resource)
```
bbangert: Why is this necessary?

autopush/db.py
```diff
@@ -293,37 +297,47 @@ def create_router_table(tablename="router", read_throughput=5,
...
+        boto_resource      # type: DynamoDBResource
+        ):
+    # type (...) -> DynamoDBTable
     """Private common function to make a table with a table func"""
-    if not table_exists(tablename):
-        return table_func(tablename, read_throughput, write_throughput)
+    if not boto_resource:
+        raise AutopushException("No boto3 resource provided for _make_table")
+    if not table_exists(tablename, boto_resource):
+        return table_func(tablename, read_throughput, write_throughput,
```
pjenvey: This should return the table wrapped in DynamoDBTable like the clause below

autopush/db.py
```diff
@@ -432,45 +445,70 @@ def generate_last_connect_values(date):
...
+        :param boto_resource: DynamoDBResource for thread
 
         """
-        self.table = table
-        self.metrics = metrics
+        self.tablename = tablename
         self._max_ttl = max_ttl
+        self.resource = boto_resource
+
+    def table(self, tablename=None):
```
pjenvey: I don't think the table() method and all the associated changes for it are necessary now that the underlying 'table' object here would be the thread local wrapper

autopush/db.py
```diff
@@ -432,45 +445,70 @@ def generate_last_connect_values(date):
...
 
-def table_exists(tablename, client=None):
-    """Determine if the specified Table exists"""
-    if not client:
-        client = g_client
-    return tablename in list_tables(client)
+class DynamoDBResource(threading.local):
+    def __init__(self, **kwargs):
+        conf = kwargs
+        if not conf.get("endpoint_url"):
```
pjenvey: this is now aws_ddb_endpoint. we don't need to handle AWS_LOCAL_DYNAMODB env here any longer since configargparse is now doing it for us

autopush/db.py
```diff
@@ -432,45 +445,70 @@ def generate_last_connect_values(date):
...
+            conf["endpoint_url"] = os.getenv("AWS_LOCAL_DYNAMODB")
+        # If there is no endpoint URL, we must delete the entry
+        if "endpoint_url" in conf and not conf["endpoint_url"]:
+            del(conf["endpoint_url"])
+        region = conf.get("region_name",
+                          os.getenv("AWS_REGION_NAME", "us-east-1"))
+        if "region_name" in conf:
+            del(conf["region_name"])
+        self.conf = conf
```
pjenvey: we don't need to keep self.conf here

autopush/tests/__init__.py
```diff
@@ -4,48 +4,45 @@
...
     # Setup the necessary message tables
     message_table = os.environ.get("MESSAGE_TABLE", "message_int_test")
-
-    create_rotating_message_table(prefix=message_table, delta=-1)
-    create_rotating_message_table(prefix=message_table)
+    create_rotating_message_table(prefix=message_table, delta=-1,
+                                  boto_resource=boto_resource)
+    create_rotating_message_table(prefix=message_table,
+                                  boto_resource=boto_resource)
+    pool = reactor.getThreadPool()
```
pjenvey: we don't need the pool adjustment size any longer

autopush/db.py
jrconlin: Sorry, I'm not sure I follow.

If a table doesn't exist, it must be created. We're passing the `create_router_table` function as the `table_func` param, since it contains the full table schema, which would not be available to the generic `DynamoDBTable`.

Are you suggesting that I move `create_router_table` as a class method of `DynamoDBTable` (which I could understand), but then, shouldn't I also do the same for `create_initial_message_tables`?

autopush/db.py
jrconlin: We use `table()` to select between Message tables, so we'd still need to call DynamoDBTable() to fetch the correct reference to the table we'd want. 

For `Router` it makes more sense to drop `table()`, since there's only going to be one.

autopush/db.py
pjenvey: What I mean is I think get_router_table and get_rotating_message_table should both return a thread safe DynamoDBTable wrapper. get_router_table/_make_table as is are returning either a DynamoDBTable wrapper if the table exists, otherwise the low level boto3 Table. That doesn't make much sense

autopush/db.py
pjenvey: What I mean is originally we had autopush.db.Message/Router with a thread safe '.table' boto2 Table object.

If we're turning '.table' into a thread safe DynamoDBTable wrapper, doesn't the need for a table() method (and the change to db.message_tables) go away? All this can continue working as before as long as the .table is thread safe (unless I'm missing something)

autopush/db.py
jrconlin: Ah! Ok. I agree.

jrconlin: Modified `_create_*_table` to return `DynamoDBTable` instances.

autopush/db.py
```diff
@@ -432,42 +444,60 @@ def generate_last_connect_values(date):
...
-            yield table
-        start_table = result.get('LastEvaluatedTableName', None)
-        if not start_table:
-            break
+def table_exists(tablename, boto_resource=None):
+    # type: (str, DynamoDBResource) -> bool
+    """Determine if the specified Table exists"""
+    try:
+        return DynamoDBTable(boto_resource, tablename).table_status in [
+            'CREATING', 'UPDATING', 'ACTIVE']
```
bbangert: Might want to use the Waiter thing to wait on creating/updating to finish, as other things may try to use the table immediately after this returns, which would fail if its not ACTIVE. That would only happen on startup I think, but could be problematic. Having a separate PR is fine.

autopush/db.py
```diff
@@ -432,42 +444,60 @@ def generate_last_connect_values(date):
...
+        return getattr(self._table, name)
 
 
 class Message(object):
     """Create a Message table abstraction on top of a DynamoDB Table object"""
-    def __init__(self, table, metrics, max_ttl=MAX_EXPIRY):
-        # type: (Table, IMetrics) -> None
+    def __init__(self, table, metrics=None, boto_resource=None,
+                 max_ttl=MAX_EXPIRY):
+        # type: (DynamoDBTable, IMetrics, DynamoDBResource, int) -> None
```
bbangert: Why does it need the boto_resource?

autopush/web/health.py
```diff
@@ -58,12 +59,12 @@ def _check_error(self, failure, name):
...
         cause = self._health_checks[name] = {"status": "NOT OK"}
-        if failure.check(InternalServerError):
+        if failure.check(InternalServerError):      # pragma nocover
             cause["error"] = "Server error"
-        elif failure.check(MissingTableException):
+        elif failure.check(MissingTableException):  # pragma nocover
             cause["error"] = failure.getErrorMessage()
         else:
-            cause["error"] = "Internal error"
+            cause["error"] = "Internal error"       # pragma nocover
```
bbangert: We had tests that covered these, now we don't?

autopush/db.py
jrconlin: I'm not sure. `table_exists` is only called before a table creation command. In those situations. The command returns that the table either exists or is in the process of creation. 

`preflight_check()` already has a table ready wait routine, and is part of the startup for the process. 

Do you believe another delay would be useful?

autopush/db.py
jrconlin: mostly for consistency and testing purposes. `DynamoDBTable` still needs to pull from the resource, so having the explicit resource passed ensures that nothing "magical" happens.

autopush/web/health.py
jrconlin: Unfortunately, the test that originally triggered that had to be dropped, and getting the code to exhibit the behavior was far more difficult than expected, mostly due to the `thread.local` element isolating portions of the code from each other. I should revisit this code and see if I can enforce this behavior, but for now, I felt it's ok to have this as a "catch" behavior for edge cases.

jrconlin: Added #1114

autopush/db.py
```diff
@@ -145,22 +139,25 @@ def make_rotating_tablename(prefix, delta=0, date=None):
...
     try:
-        table = g_dynamodb.Table(tablename)
-        if table.table_status == 'ACTIVE':  # pragma nocover
+        table = DynamoDBTable(boto_resource, tablename)
+        if table.table_status == 'ACTIVE':
             return table
     except ClientError as ex:
         if ex.response['Error']['Code'] != 'ResourceNotFoundException':
-            # If we hit this, our boto3 is misconfigured and we need to bail.
-            raise ex  # pragma nocover
```
pjenvey: Might we still want to raise exceptions here (I'm not sure)? Either way this block of code isn't necessary any longer as it's a "pass" for any exception

autopush/db.py
```diff
@@ -694,18 +724,25 @@ def update_last_message_read(self, uaid, timestamp):
 
 class Router(object):
     """Create a Router table abstraction on top of a DynamoDB Table object"""
-    def __init__(self, table, metrics, max_ttl=MAX_EXPIRY):
-        # type: (Table, IMetrics, int) -> None
+    def __init__(self, conf, metrics, max_ttl=MAX_EXPIRY, resource=None):
```
pjenvey: FYI I don't we need this change from table to conf any longer nor the _cached_table var below. Changing the conf argument back will require a number of changes to test_db so I'm not going to cry over this

autopush/db.py
jrconlin: ah! Good catch, thanks!

jrconlin:
Added merge fixes for `test_rs_integration.py`

autopush/tests/test_rs_integration.py
```diff
@@ -104,14 +104,19 @@ def setUp(self):
         )
 
         # Endpoint HTTP router
-        self.ep = ep = EndpointApplication(ep_conf)
-        ep.db.client = db.g_client
+        self.ep = ep = EndpointApplication(
+            ep_conf,
+            resource=autopush.tests.boto_resource
+        )
```
jrconlin: Merge change to use boto_resource for testing.

PR #1089, opened 2017-11-30
feat: add sentry panic reporting

### LINKED ISSUE #1066: Use Sentry for error reporting in Rust code
The Rust connection node should use Sentry for panic reporting.

bbangert:
Closes #1066

### DIFF
--- a/autopush_rs/src/server/mod.rs
+++ b/autopush_rs/src/server/mod.rs
@@ -4,6 +4,7 @@ use std::ffi::CStr;
 use std::io;
 use std::net::{IpAddr, ToSocketAddrs};
 use std::panic;
+use std::panic::PanicInfo;
 use std::path::PathBuf;
 use std::rc::Rc;
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


autopush_rs/src/server/mod.rs
```diff
@@ -297,6 +300,21 @@ impl Server {
...
+        let sentry_dsn_option = option_env!("SENTRY_DSN");
+        if let Some(sentry_dsn) = sentry_dsn_option {
+            // Spin up a new thread with a new reactor core for the sentry handler
+            thread::spawn(move || {
+                let creds = sentry_dsn
+                    .parse::<sentry::SentryCredential>()
+                    .expect("Invalid Sentry DSN specified");
+                let mut core = Core::new().expect("Unable to create core");
+                let sentry = sentry::Sentry::from_settings(core.handle(), Default::default(), creds);
+                sentry.register_panic_handler(Some(|_: &PanicInfo| -> () {}));
```
alexcrichton: I think that the empty callback here may mean that a panic doesn't actually get logged to stdout/stderr, but maybe it'd be good to print something briefly or run the default panic handler perhaps? (just in case)

bbangert:
Just a note, this does not retain the JoinHandle for the new Sentry thread. For a future issue where we fix the exit to be graceful, the Join handle for the thread will need to be retained somewhere so it can be join'd on graceful shutdown.

PR #1094, opened 2017-12-02
feat: add some metrics for twisted's threadpool

### LINKED ISSUE #1093: Emit metrics for twisted's threadpool
Visibility into our threadpool usage would be interesting, in particular, how many pending jobs are queued for it at any given time on production. Being an unbounded queue we want to ensure we're not abusing it

pjenvey:
emitted every 15 seconds per datadog's preferred minimum

Closes #1093

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -129,3 +129,21 @@ def from_config(conf):
         return TwistedMetrics(conf.statsd_host, conf.statsd_port)
     else:
         return SinkMetrics()
+
+
+def periodic_reporter(metrics, prefix=''):
+    # type: (IMetrics, Optional[str]) -> None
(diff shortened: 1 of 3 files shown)


autopush/metrics.py
```diff
@@ -129,3 +129,21 @@ def from_config(conf):
         return TwistedMetrics(conf.statsd_host, conf.statsd_port)
     else:
         return SinkMetrics()
+
+
+def periodic_reporter(metrics, prefix=''):
+    # type: (IMetrics, Optional[str]) -> None
+    """Emit metrics on twisted's thread pool.
+
+    Only meant to be ran as a LoopingCall (TimerService).
```
jrconlin: nit: "run"

PR #1105, opened 2018-01-18
feat: add integration testing for Rust connection node

### LINKED ISSUE #1060: Add tests for the Rust connection handling
The Rust connection handler needs testing, both some unit tests, and the application of the integration testing.

bbangert:
Adds the following functionality for capability parity:

- Skip sending messages if the message is expired
- Properly handle legacy messages in a message stream
- Set appropriate flags for a uaid not found in the db
- Always return a timestamp when querying into timestamp messages
- Send messages in the order they're retrieved from the db
- Accept messages from endpoint while waiting for acks
- Don't save TTL:0 messages in the db if the client fails to ack them
- Allow TTL:None from endpoint and treat as TTL:0

Closes #1060

### DIFF
--- a/autopush/tests/test_z_main.py
+++ b/autopush/tests/test_z_main.py
@@ -1,3 +1,9 @@
+"""Test main instantiation
+
+This is named test_z_main.py to run it last. Due to issues in this test, the
+testing environment is unclean and no further tests can be run reliably.
+
+"""
 import unittest
 import datetime
 import json
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -72,7 +72,7 @@ pub struct WebPushClient {
     unacked_stored_notifs: Vec<Notification>,
     // Highest version from stored, retained for use with increment
     // when all the unacked storeds are ack'd
-    unacked_stored_highest: Option<i64>,
+    unacked_stored_highest: Option<u64>,
     connected_at: u64,
     stats: SessionStatistics,
 }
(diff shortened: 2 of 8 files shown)


autopush/tests/test_z_main.py
```diff
@@ -1,3 +1,9 @@
+"""Test main instantiation
```
jrconlin: Will you file an issue to clean up these tests?

autopush_rs/src/client.rs
```diff
@@ -316,17 +327,30 @@ where
...
-                    webpush.unacked_stored_notifs.extend(
-                        messages.iter().cloned(),
-                    );
-                    let message = ServerMessage::Notification(messages.pop().unwrap());
-                    ClientState::FinishSend(
-                        Some(message),
-                        Some(Box::new(ClientState::SendMessages(Some(messages)))),
-                    )
+                    // Filter out TTL expired messages
+                    let now = time::get_time().sec as u32;
```
jrconlin: In rust, we use timestamps of <u32>, but python has timestamps of <u64>. Do we need to cast up to prevent possible overruns?

autopush/tests/test_z_main.py
pjenvey: I got it: https://github.com/mozilla-services/autopush/issues/1106

autopush_rs/src/client.rs
```diff
@@ -271,7 +276,13 @@ where
                         uaid,
                         use_webpush: Some(true),
                         ..
-                    } => uaid,
+                    } => {
+                        if let Some(uaid) = uaid {
```
pjenvey: this could be `uaid.and_then(|uaid| Uuid::parse_str(uaid.as_str()).ok())`

use_webpush: Some(true) isn't necessary in the destructuring above either is it

bbangert: Thanks! I knew there was something shorter I was missing.

bbangert: I think the Some(true) is needed to enforce that it was set to True?

pjenvey: I guess but it's never false anyway

PR #1110, opened 2018-01-25
feat: add /notif/ HTTP API handler

### LINKED ISSUE #1062: Implement PUT /notif/ in Rust
The equivalent of NotificationHandler needs to be added to the rust connection server. In http.rs, triggering a ServerNotification::CheckStorage event.

bbangert:
Refactors the HTTP handling slightly for easier additions in the
future.

Closes #1062

### DIFF
--- a/autopush_rs/src/http.rs
+++ b/autopush_rs/src/http.rs
@@ -1,14 +1,18 @@
-//! Dummy module that's very likely to get entirely removed
+//! Internal router HTTP API
 //!
-//! For now just a small HTTP server used to send notifications to our dummy
-//! clients.
+//! Accepts PUT requests to deliver notifications to a connected client or trigger
+//! a client to check storage.
+//!
+//! Valid URL's:
+//!     PUT /push/UAID      - Deliver notification to a client
+//!     PUT /notify/UAID    - Tell a client to check storage
 
 use std::str;
 use std::rc::Rc;
 
-use futures::future::err;
+use futures::future::ok;
 use futures::{Stream, Future};
-use hyper::Method;
+use hyper::{Method, StatusCode};
 use hyper;
 use serde_json;
 use tokio_service::Service;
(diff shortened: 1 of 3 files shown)


autopush_rs/src/server/mod.rs
```diff
@@ -463,6 +463,20 @@ impl Server {
         Err("User not connected".into())
     }
 
+    /// A check for notification command has come for the uaid
+    pub fn check_client_storage(&self, uaid: Uuid) -> Result<()> {
+        let mut uaids = self.uaids.borrow_mut();
```
pjenvey: I don't think you need mut here or below? (use .borrow() and .get())

autopush_rs/src/server/mod.rs
```diff
@@ -463,6 +463,20 @@ impl Server {
...
+        if let Some(client) = uaids.get_mut(&uaid) {
+            client
+                .tx
+                .unbounded_send(ServerNotification::CheckStorage)
```
pjenvey: (in my quest to kill all our unwraps) I think .unbounded_send()? should work here instead of unwrap

autopush_rs/src/http.rs
```diff
@@ -1,14 +1,18 @@
...
+//!
+//! Valid URL's:
+//!     PUT /push/UAID      - Deliver notification to a client
+//!     PUT /notify/UAID    - Tell a client to check storage
 
 use std::str;
 use std::rc::Rc;
 
-use futures::future::err;
+use futures::future::{ok, err};
```
pjenvey: unused import

bbangert: It's used on line 37 and 88.

autopush_rs/src/server/mod.rs
bbangert: Ah, was just copy/pasting from the existing notify_client for the most part, which did the borrow_mut and such as well.

autopush_rs/src/http.rs
bbangert: Oh, the err.

autopush_rs/src/server/mod.rs
bbangert: @alexcrichton why was a borrow_mut() used on the notify_client func I coppied instead of borrow/get?

alexcrichton: Hm I think maybe that was from a long time ago? It may have originally done something mutable but may have been refactored away!

Hm I think maybe that was from a long time ago? It may have originally done something mutable but may have been refactored away!

autopush_rs/src/server/mod.rs
bbangert: Using ? here doesn't work as there's no From trait that satisfies the Result that unbounded_send returns, but using expect() does work fine.

bbangert: Course, expect panics.... so I'm just going to test the result and handle it properly. :)

pjenvey: > Using ? here doesn't work as there's no From trait that satisfies the Result that unbounded_send returns, but using expect() does work fine.

ah yes. we could alternatively add futures::sync::mpsc::SendError<ServerNotification> to error_chain! to get the From trait for this to work

PR #1115, opened 2018-01-31
feat: add metric tags

### LINKED ISSUE #1054: Add statsd metric reporting to Rust version

pjenvey:
relies on cadence dev w/ new tag support
    
also:

move ua parsing into utils w/ setup of metrics' specific values and
revert a stray pdb.set_trace
    
Closes #1054

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -16,12 +16,13 @@ use futures::{Stream, Sink, Future, Poll, Async};
 use tokio_core::reactor::Timeout;
 use time;
 use uuid::Uuid;
-use woothee::parser::{Parser, WootheeResult};
+use woothee::parser::Parser;
 
 use call;
 use errors::*;
 use protocol::{ClientAck, ClientMessage, ServerMessage, ServerNotification, Notification};
 use server::Server;
+use util::parse_user_agent;
 
 pub struct RegisteredClient {
     pub uaid: Uuid,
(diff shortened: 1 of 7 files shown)


autopush_rs/src/client.rs
```diff
@@ -719,8 +719,16 @@ where
 
         let now = time::precise_time_ns() / 1000;
         let elapsed = now - webpush.connected_at;
-        // XXX: tags
-        self.srv.metrics.time("ua.connection.lifespan", elapsed).ok();
+
+        let parser = Parser::new();
```
pjenvey: (note that Parser's lifetime requires it here, logged: https://github.com/woothee/woothee-rust/issues/5)

autopush_rs/src/util/user_agent.rs
```diff
@@ -0,0 +1,42 @@
...
+        category: "",
+        os: "",
+        os_version: "".to_string(),
+        browser_type: "",
+        version: "".to_string(),
+        vendor: "",
+    });
+
+    // Determine a base os/browser for metrics' tags
+    let metrics_os = if wresult.os.starts_with("Windows") {
```
jrconlin: Is case a concern? Should we check these against lowercase only?

pjenvey: we're comparing against what woothee gives us which is always title case for these (see https://github.com/woothee/woothee-rust/blob/master/src/dataset.rs). note that woothee gives us "Mac OSX" vs "Mac OS X" that the python parser gave

There is a concern here, though -- this could use some tests..

jrconlin: ah, ok 👍

PR #1125, opened 2018-02-06
bug: Use environment var AWS_DEFAULT_REGION as region specifier

### LINKED ISSUE #1124: Remove env override in boto3 code
` autopush/main.py` incorrectly pulls from the `AWS_REGION_NAME` instead of `AWS_DEFAULT_REGION`

jrconlin:
Closes #1124

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -464,7 +464,7 @@ class DynamoDBResource(threading.local):
         if "endpoint_url" in conf and not conf["endpoint_url"]:
             del(conf["endpoint_url"])
         region = conf.get("region_name",
-                          os.getenv("AWS_REGION_NAME", "us-east-1"))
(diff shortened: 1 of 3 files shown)


autopush/db.py
```diff
@@ -464,7 +464,7 @@ def __init__(self, **kwargs):
         if "endpoint_url" in conf and not conf["endpoint_url"]:
             del(conf["endpoint_url"])
         region = conf.get("region_name",
-                          os.getenv("AWS_REGION_NAME", "us-east-1"))
+                          os.getenv("AWS_DEFAULT_REGION", "us-east-1"))
```
pjenvey: we might be able to nuke this entirely (just let boto3 look it up) but we can do this later

jrconlin: Agreed.

PR #1130, opened 2018-02-12
bug: Silence remaining nuisance alarms

### LINKED ISSUE #1123: Capture exceptions that are connection loss related
Various connection breakage throws exceptions into Sentry. These are not bugs we plan on fixing, so they should be captured and dismissed as connection breaks.

Sentry bug(s):
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/1673127/
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/643290/
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/1737236/

And a variety of Connect timeout exceptions, that may need to be handled differently:
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/1814177/
- https://sentry.prod.mozaws.net/operations/autopush-prod/issues/1814153/
- etc.

jrconlin:
Closes #1123

### DIFF
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -1,5 +1,5 @@
 """APNS Router"""
-import ssl
+import socket
 import uuid
 from typing import Any  # noqa
 
@@ -157,22 +157,11 @@ class APNSRouter(object):
                                        self._base_tags,
                                        application=rel_channel,
                                        reason="connection_error"))
-        except HTTP20Error:
+        except (HTTP20Error, socket.error):
             self.metrics.increment("notification.bridge.connection.error",
                                    tags=make_tags(self._base_tags,
                                                   application=rel_channel,
                                                   reason="http2_error"))
-        except ssl.SSLError as e:
-            # can only str match this (for autopush#1048)
-            if not (e.errno == ssl.SSL_ERROR_SSL and
-                    str(e).startswith("[SSL: BAD_WRITE_RETRY]")):
-                raise  # pragma: nocover
-            self.metrics.increment(
-                "notification.bridge.connection.error",
-                tags=make_tags(self._base_tags,
-                               application=rel_channel,
-                               reason="bad_write_retry_error")
-            )
         if not success:
             raise RouterException(
                 "Server error",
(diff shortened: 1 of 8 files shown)


autopush/router/apnsrouter.py
```diff
@@ -157,7 +158,7 @@ def _route(self, notification, router_data):
                                        self._base_tags,
                                        application=rel_channel,
                                        reason="connection_error"))
-        except HTTP20Error:
+        except (HTTP20Error, socket.error):
```
pjenvey: This makes the SSLError block below dead code (SSLError derives from socket.error). Go ahead and kill it

autopush/main.py
```diff
@@ -105,7 +106,10 @@ def run(self):
         """Start the services and run the reactor"""
         reactor.suggestThreadPoolSize(constants.THREAD_POOL_SIZE)
         self.startService()
-        reactor.run()
+        try:
+            reactor.run()
+        except ConnectionDone:
```
pjenvey: this one's not right but I can see what lead you towards it: our stack trace order in sentry is screwed up (logged #1134). Look at it again in reverse =]

autopush/web/base.py
```diff
@@ -238,7 +238,10 @@ def _response_err(self, fail):
         running.
 
         """
+        from twisted.internet.error import ConnectionDone
         fmt = fail.value.message or 'Exception'
+        if fail.value.__class__ in [ConnectionDone]:
```
pjenvey: why not isinstance?

autopush/main.py
jrconlin: I'm not so sure. I was looking at https://sentry.prod.mozaws.net/operations/autopush-prod/issues/643290/. That seems to indicate that the `ConnectionDone` error is happening as part of `reactor.run()` here.

pjenvey: that one's due to web.base _response_err logging it, you can see it in the next to last stack frame

autopush/web/base.py
jrconlin: wasn't sure if there were going to be others.

pjenvey: commonly unknown but so nice: isinstance arg 2 can be a tuple of types

autopush/main.py
jrconlin: I have no idea how the hell this got reverted. Removing this try since the ConnectionDone is handled at a lower layer.

PR #1131, opened 2018-02-13
fix: make autopush_rs's ctrl-c work gracefully

### LINKED ISSUE #1090: fix ctrl-c not working under autopush_rs
Currently autopush_rs doesn't correctly handle SIGINT to quit the process. Ideally it would be joining on all the worker threads -- but thread join blocks signal handling.

We may need to make a busy loop in the main python thread to handle receiving the signal and triggering shutdown -- not ideal but I'm not sure there's another option. I think Python 3 may have some improvements for this situation

pjenvey:
also include the rust sentry thread in the shutdown process and fix
pulling SENTRY_DSN from build vs runtime env

Closes #1090

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -1,5 +1,6 @@
 """autopush/autoendpoint daemon scripts"""
 import os
+import time
 from argparse import Namespace  # noqa
 
 from twisted.application.internet import (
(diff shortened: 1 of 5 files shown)


autopush/main.py
```diff
@@ -345,9 +345,17 @@ def setup(self, rotate_tables=True):
...
         try:
-            self.push_server.run()
+            self.push_server.start()
+            while True:
+                try:
+                    # handle a graceful shutdown on SIGINT w/ a busy
+                    # loop. we can't Thread.join because SIGINT won't
+                    # interrupt it
+                    time.sleep(6000)
+                except KeyboardInterrupt:
```
jrconlin: why not signal? (it'd handle other kills too, right?)

pjenvey: signal could special handle potentially any kill but you have to register every specific signal. I could have used a signal handler but I'd actually still need this busy loop to keep the main thread going (python signal handlers should only run on the main thread -- and I even just double checked, they wouldn't have run if the main thread has died)

the special handling for KeyboardInterrupt here isn't even totally necessary, it's just to hide the traceback on the console for the common ctrl-c case

PR #1135, opened 2018-02-13
feat: drain pending notifications from the client rx queue

### LINKED ISSUE #1133: Drain notification queue on client drop
The notification queue is ignored when the client drops, its shutdown should drain the queue to ensure any notifications stuck in it are saved to storage.

bbangert:
Closes #1133

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -712,7 +712,7 @@ where
 
     pub fn shutdown(&mut self) {
         // If we made it past hello, do more cleanup
-        let webpush = match self.webpush.take() {
+        let mut webpush = match self.webpush.take() {
             Some(webpush) => webpush,
             None => return
         };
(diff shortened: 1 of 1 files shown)


autopush_rs/src/client.rs
```diff
@@ -730,6 +730,19 @@ where
             .send()
             .ok();
 
+        // If there's any notifications in the queue, move them to our unacked direct notifs
+        webpush.rx.close();
+        let rx_iter = webpush.rx.wait();
```
alexcrichton: Initially this threw me for a bit of a spin triggering the "oh no the `wait` method shouldn't be used on the event loop!" but after thinking about it I think it's fine due to the `close` above. Mind throwing a comment here saying "no worries, this won't block due to the `close` above"?

autopush_rs/src/client.rs
```diff
@@ -730,6 +730,19 @@ where
...
+        for msg in rx_iter {
+            match msg {
+                Ok(ServerNotification::CheckStorage) => continue,
+                Ok(ServerNotification::Notification(notif)) => {
+                    webpush.unacked_direct_notifs.push(notif.clone());
```
pjenvey: is this clone necessary? seems like we own it once it comes out of the queue

bbangert: Indeed, removing.

autopush_rs/src/client.rs
bbangert: Ah, yes, added a callout.

autopush_rs/src/client.rs
```diff
@@ -712,7 +712,7 @@ where
 
     pub fn shutdown(&mut self) {
         // If we made it past hello, do more cleanup
-        let webpush = match self.webpush.take() {
+        let mut webpush = match self.webpush.take() {
```
pjenvey: one more nitpick: we don't have to clone stats below any longer now that webpush is mut

bbangert: The webpush.rx and webush.unacked* calls require mutable borrows.

bbangert: Oh, nevermind, I see what you mean.

PR #1141, opened 2018-03-02
fix: have webpush router handle ClientErrors

### LINKED ISSUE #1138: Handle boto3 "InternalServerError"
From sentry:

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/2696847/
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/2710042/
https://sentry.prod.mozaws.net/operations/autopush-prod/issues/2872776/

These botocore.errorfactory.InternalServerError look analogous to the generic boto2 BotoServerError  we still handle -- generally triggered when the number of retries exceeded max

pjenvey:
it was previously handling only the old boto2 JSONResponseError in
those spots

also kill the remaining boto2 bits

Closes #1138

### DIFF
--- a/autopush/web/health.py
+++ b/autopush/web/health.py
@@ -1,7 +1,7 @@
 """Health Check HTTP Handler"""
 
 import cyclone.web
-from boto.dynamodb2.exceptions import InternalServerError
+from botocore.exceptions import ClientError
 from twisted.internet.defer import DeferredList
 from twisted.internet.threads import deferToThread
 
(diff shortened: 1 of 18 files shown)


autopush/web/health.py
```diff
@@ -61,7 +61,9 @@ def _check_error(self, failure, name):
         cause = self._health_checks[name] = {"status": "NOT OK"}
         if failure.check(MissingTableException):
             cause["error"] = failure.value.message
-        elif failure.check(InternalServerError):  # pragma nocover
+        elif (failure.check(ClientError) and
+              failure.value.response["Error"]["Code"] ==
+              "InternalServerError"):  # pragma nocover
```
jrconlin: This is becoming a reoccurring pattern. I wonder if we should create a function that just does this combined check?

pjenvey: Totally, I almost created one in this diff but figured I'd keep these short/readable. Logged https://github.com/mozilla-services/autopush/issues/1142

Issue #1129, opened 2018-02-09
Create megaphone integration


bbangert:
The megaphone API should be implemented in autopush_rs.

Doc: https://docs.google.com/document/d/1Wxqf1a4HDkKgHDIswPmhmdvk8KPoMEh2q6SPhaz4LNE/

bbangert:
As an addendum, this work includes the websocket API change to support sending out the broadcast message during ping handling, as well as having the connection node load the list on load.

PR #1146, opened 2018-03-04
feat: add megaphone service change tracker

bbangert:
Issue #1129

### DIFF
--- a/autopush_rs/src/util/megaphone.rs
+++ b/autopush_rs/src/util/megaphone.rs
@@ -1,0 +1,246 @@
+use std::collections::HashMap;
+use errors::Result;
+
+// A Service entry Key in a ServiceRegistry
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+struct ServiceKey(u32);
+
+// A list of services that a client is interested in and the last change seen
+#[derive(Debug)]
+pub struct ClientServices {
+    service_list: Vec<ServiceKey>,
+    change_count: u32,
+}
+
+#[derive(Debug)]
+struct ServiceRegistry {
+    lookup: HashMap<String, u32>,
+    table: Vec<String>,
+}
+
+// Return result of the first delta call for a client given a full list of service id's and versions
+#[derive(Debug)]
+pub struct ServiceClientInit {
+    pub service_list: ClientServices,
+    pub delta: Vec<Service>,
+}
+
+impl ServiceRegistry {
+    fn new() -> ServiceRegistry {
+        ServiceRegistry {
+            lookup: HashMap::new(),
+            table: Vec::new(),
+        }
+    }
+
+    // Add's a new service to the lookup table, returns the existing key if the service already
+    // exists
+    fn add_service(&mut self, service_id: String) -> ServiceKey {
+        if let Some(v) = self.lookup.get(&service_id) {
+            return ServiceKey(v.clone());
+        }
+        let i = self.table.len();
+        self.table.push(service_id.clone());
+        self.lookup.insert(service_id, i as u32);
+        ServiceKey(i as u32)
+    }
+
+    fn lookup_id(&self, key: ServiceKey) -> Option<String> {
+        self.table.get(key.0 as usize).cloned()
+    }
+
+    fn lookup_key(&self, service_id: String) -> Option<ServiceKey> {
+        self.lookup.get(&service_id).cloned().map(ServiceKey)
+    }
+}
+
+// An individual service and the current change count
+#[derive(Debug)]
+struct ServiceRevision {
+    change_count: u32,
+    service: ServiceKey,
+}
+
+// A provided Service/Version used for `ChangeList` initialization, client comparisons, and
+// outgoing deltas
+#[derive(Clone, Debug, PartialEq)]
+pub struct Service {
+    service_id: String,
+    version: String,
+}
+
+// ServiceChangeTracker tracks the services, their change_count, and the service lookup registry
+#[derive(Debug)]
+pub struct ServiceChangeTracker {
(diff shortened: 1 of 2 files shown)


autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,253 @@
...
+            return ServiceKey(v.clone());
+        }
+        let i = self.table.len();
+        self.table.push(service_id.clone());
+        self.lookup.insert(service_id, i as u32);
+        ServiceKey(i as u32)
+    }
+
+    fn lookup_id(&self, key: ServiceKey) -> Option<String> {
+        if key.0 >= (self.table.len() as u32) {
```
alexcrichton: This may be able to be simplified with:

```rust
self.table.get(key.0 as usize).cloned()
```

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,253 @@
...
+            None
+        } else {
+            Some(self.table[key.0 as usize].clone())
+        }
+    }
+
+    fn lookup_key(&self, service_id: String) -> Option<ServiceKey> {
+        self.lookup.get(&service_id).and_then(|v| Some(ServiceKey(v.clone())))
```
alexcrichton: `.and_then(|x| Some(...))` should be the same as `map(|x| ...)`

alexcrichton: although like above this could be `self.lookup.get(&service_id).cloned().map(ServiceKey)` (depending on preferences of course)

autopush_rs/src/util/megaphone.rs
bbangert: Thanks, I was thinking there would be a func like that and missed it somehow!

autopush_rs/src/util/megaphone.rs
bbangert: I like that last one!

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,249 @@
...
+        let mut svc_delta = Vec::new();
+        while let Some(svc) = svc_iter.next() {
+            if svc.change_count <= client_set.change_count {
+                break;
+            }
+            if let Some(ver) = self.service_versions.get(&svc.service) {
+                if let Some(svc_id) = self.service_registry.lookup_id(svc.service) {
+                    svc_delta.push(Service {
+                        service_id: svc_id,
+                        version: (*ver).clone(),
```
pjenvey: Could probably kill some of the clone()s in this module, but I don't care for now. This is the most important one

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,249 @@
...
+
+#[derive(Debug)]
+struct ServiceRegistry {
+    lookup: HashMap<String, u32>,
+    table: Vec<String>,
+}
+
+// Return result of the first delta call for a client given a full list of service id's and versions
+#[derive(Debug)]
+pub struct ServiceClientInit {
```
pjenvey: nitpick: tempted to kill this for a tuple struct since it's only used as one return type

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,249 @@
...
+                return Ok(self.change_count);
+            }
+            *ver = service.version;
+        } else {
+            return Err("Service not found".into());
+        }
+
+        // Check to see if this service has been updated since initialization
+        let mut svc_index: Option<usize> = None;
+        for (i, svc) in self.service_list.iter().enumerate() {
```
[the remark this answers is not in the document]
bbangert: Won't this iterate through the entire thing even if it finds a match early?

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,249 @@
...
+            if let Some(ver) = self.service_versions.get(&svc.service) {
+                if let Some(svc_id) = self.service_registry.lookup_id(svc.service) {
+                    svc_delta.push(Service {
+                        service_id: svc_id,
+                        version: (*ver).clone(),
+                    });
+                }
+            }
+        }
+        client_set.change_count = self.change_count;
```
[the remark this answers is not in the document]
bbangert: I was trying to remove knowledge of internals from the caller, so that it'd just have a 'thing' that can be checked/updated without know the internal details.

autopush_rs/src/util/megaphone.rs
bbangert: These will need to be clone'd since they're going to have to be passed back out to serde at some point, which will exceed the scope of which a borrow would work from the caller.

autopush_rs/src/util/megaphone.rs
```diff
@@ -0,0 +1,249 @@
+use std::collections::HashMap;
+use errors::Result;
+
+// A Service entry Key in a ServiceRegistry
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+struct ServiceKey(u32);
+
+// A list of services that a client is interested in and the last change seen
+#[derive(Debug)]
+pub struct ServiceList {
```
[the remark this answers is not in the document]
bbangert: how about ClientServices to indicate its for clients services it tracks?

autopush_rs/src/util/megaphone.rs
bbangert: In the case here, removing clone() annoys the compiler cause *ver cannot be moved out of borrowed context.

bbangert:
Since this is going to be a multi-PR issue, closing this in favor of a different branch name.

PR #1148, opened 2018-03-06
bug: fix _expiry() function to use seconds

### LINKED ISSUE #1073: Switch db.py boto3 get and query calls to use FilterExpression for `expiry`
see comments in db.py 
`TODO: once expiry is properly integrated, use FilterExpression`

Blocked until JAN 2018 to allow for sufficient numbers of channels to expire from rollover.

jrconlin:
Also enforce expiry on router records

Closes #1073

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -322,7 +322,7 @@ def _make_table(
 
 
 def _expiry(ttl):
-    return int(time.time() + ttl)
+    return int(time.time()) + ttl
 
 
 def get_router_table(tablename="router", read_throughput=5,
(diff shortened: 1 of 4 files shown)


jrconlin:
Please refer to https://github.com/mozilla-services/autopush/issues/1073#issuecomment-370939896 for why `filterExpression` is not being used here.

FWIW, the core problem is that if the key identified by `ExclusiveStartKey` can't be found (because, say the record expired and was dropped by the database), the query fails with a Validation error and the backfill fails.

PR #1152, opened 2018-03-13
feat: add megaphone broadcast handling, ping check, server tracking

bbangert:
Handles broadcast subscriptions per client, holds a single broadcast
change set tracker, and sends out broadcasts at the ping interval to
clients that have new broadcast deltas to send.

Issue #1129 

Note: Only remaining work is to add the megaphone polling portion that grabs the broadcast master list and keeps it up to date.

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -5,6 +5,7 @@
 //! of connected clients. Note that it's expected there'll be a lot of connected
 //! clients, so this may appears relatively heavily optimized!
 
+use std::collections::HashMap;
 use std::rc::Rc;
 
 use cadence::prelude::*;
@@ -23,6 +24,7 @@ use errors::*;
 use protocol::{ClientAck, ClientMessage, ServerMessage, ServerNotification, Notification};
 use server::Server;
 use util::parse_user_agent;
+use util::megaphone::{ClientServices, Service, ServiceClientInit};
 
 pub struct RegisteredClient {
     pub uaid: Uuid,
@@ -68,6 +70,7 @@ pub struct WebPushClient {
     uaid: Uuid,
     rx: mpsc::UnboundedReceiver<ServerNotification>,
     flags: ClientFlags,
+    broadcast_services: ClientServices,
     message_month: String,
     unacked_direct_notifs: Vec<Notification>,
     unacked_stored_notifs: Vec<Notification>,
@@ -120,7 +123,7 @@ impl ClientFlags {
 
 pub enum ClientState {
     WaitingForHello(Timeout),
-    WaitingForProcessHello(MyFuture<call::HelloResponse>),
+    WaitingForProcessHello(MyFuture<call::HelloResponse>, Vec<Service>),
     WaitingForRegister(Uuid, MyFuture<call::RegisterResponse>),
     WaitingForUnRegister(Uuid, MyFuture<call::UnRegisterResponse>),
     WaitingForCheckStorage(MyFuture<call::CheckStorageResponse>),
@@ -185,6 +188,14 @@ where
 
     pub fn shutdown(&mut self) {
         self.data.shutdown();
+    }
+
+    pub fn broadcast_delta(&mut self) -> Option<Vec<Service>> {
+        if let Some(ref mut webpush) = self.data.webpush {
+            self.data.srv.broadcast_delta(&mut webpush.broadcast_services)
+        } else {
+            None
+        }
     }
 
     fn transition(&mut self) -> Poll<ClientState, Error> {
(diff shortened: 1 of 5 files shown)


autopush_rs/src/client.rs
```diff
@@ -570,15 +615,18 @@ where
         rotate_message_table: bool,
         check_storage: bool,
         connected_at: u64,
+        services: &Vec<Service>,
```
alexcrichton: FWIW `&[T]` tends to be more idiomatic rather than `&Vec<T>`

autopush_rs/src/client.rs
```diff
@@ -479,7 +524,7 @@ where
                     Either::A(ClientMessage::Nack { .. }) => {
                         self.data.srv.metrics.incr("ua.command.nack").ok();
                         self.data.webpush.as_mut().unwrap().stats.nacks += 1;
-                        ClientState::WaitingForAcks
+                        ClientState::Await
```
bbangert: This was a  bug I noticed and fixed here, as its supposed to return to the prior state.

autopush_rs/src/client.rs
bbangert: Ah, is that used only when its a reference, or just in general as its shorter than Vec<T>?

autopush_rs/src/client.rs
```diff
@@ -422,6 +439,20 @@ where
                     return Ok(next_state.into());
                 }
                 match try_ready!(self.data.input_or_notif()) {
+                    Either::A(ClientMessage::BroadcastSubscribe { broadcasts }) => {
```
jrconlin: nit: "broadcasts" is being used as a parameter, local variable and key here. It's a bit confusing.

bbangert: It's used to destructure, then its re-bound locally as the Vec, and passed into a function. I can remove an intermediary reassignment though and pass the resulting Vec<Service> directly to the single use of it in a function, that might clear it up some.

autopush_rs/src/server/mod.rs
```diff
@@ -659,8 +702,20 @@ impl<T> WebpushSocket<T> {
         Error: From<T::SinkError>,
     {
         if self.ping {
-            debug!("sending a ping");
-            match self.inner.start_send(Message::Ping(Vec::new()))? {
+            let mut msg = Message::Ping(Vec::new());
```
pjenvey: easy mut kill here: let msg = if let etc else Message::Ping(Vec::new())

autopush_rs/src/server/mod.rs
```diff
@@ -546,10 +569,28 @@ impl Future for PingManager {
...
-                    self.timeout.reset(at);
-                    self.waiting = WaitingFor::Pong;
+                    // If we just sent a broadcast, reset the ping interval and clear the delta
+                    if socket.broadcast_delta.is_some() {
+                        let at = Instant::now() + self.srv.opts.auto_ping_interval;
+                        self.timeout.reset(at);
+                        self.waiting = WaitingFor::SendPing;
+                        socket.broadcast_delta = None;
+                    } else {
+                        let at = Instant::now() + self.srv.opts.auto_ping_timeout;
```
pjenvey: these 2 lines are duplicated in both cases

autopush_rs/src/util/megaphone.rs
```diff
@@ -60,12 +60,35 @@ struct ServiceRevision {
...
+        (svc.service_id, svc.version)
+    }
+}
+
+impl Service {
+    pub fn from_hashmap(val: HashMap<String, String>) -> Vec<Service> {
+        val.into_iter().map(|v| v.into()).collect()
+    }
+
+    pub fn to_hashmap(service_vec: Vec<Service>) -> HashMap<String, String> {
```
pjenvey: this function consumes service_vec so one of the rust naming conventions prefers it called into_hashmap https://deterministic.space/elegant-apis-in-rust.html#more-method-name-conventions

autopush_rs/src/client.rs
pjenvey: He just meant for &Vec<T>. If you're not doing anything too Vec specific, more types are able to deref into &[T] so it's good practice to prefer the slice type. I just linked this doc so, here's its further explanation https://deterministic.space/elegant-apis-in-rust.html#use-conversion-traits

autopush_rs/src/util/megaphone.rs
```diff
@@ -198,6 +224,31 @@ impl ServiceChangeTracker {
             svc_delta,
         )
     }
+
+    /// Update a `ClientServices` to account for a new service.
+    ///
+    /// Returns services that have changed.
+    pub fn client_service_add_service(&self, client_service: &mut ClientServices, services: &[Service]) -> Option<Vec<Service>> {
```
pjenvey: long line! (we could use cargo fmt run, separate commit though)

autopush_rs/src/util/megaphone.rs
```diff
@@ -198,6 +224,31 @@ impl ServiceChangeTracker {
...
+        let mut svc_delta = self.change_count_delta(client_service).unwrap_or(Vec::new());
+        for svc in services.iter() {
```
pjenvey: off hand i don't know if it's easily refactorable/doable but: service_delta has a very similar loop here

bbangert: Indeed, they both need to do slightly different things though, so I didn't see any way to abstract it out.

PR #1156, opened 2018-03-14
feat: add megaphone API query and repeating service updating

bbangert:
Megaphone API is now queried on startup and at every poll interval to
update the change state tracker.

Issue #1129 

Remaining work after this is going to be a fairly large set of integration tests to close out the issue.

### DIFF
--- a/autopush_rs/src/server/mod.rs
+++ b/autopush_rs/src/server/mod.rs
@@ -20,6 +20,7 @@ use hyper;
 use hyper::server::Http;
 use libc::c_char;
 use openssl::ssl::SslAcceptor;
+use reqwest;
 use sentry;
 use serde_json;
 use time;
@@ -42,7 +43,7 @@ use server::dispatch::{Dispatch, RequestType};
 use server::metrics::metrics_from_opts;
 use server::webpush_io::WebpushIo;
 use util::{self, RcObject, timeout};
-use util::megaphone::{ClientServices, Service, ServiceClientInit, ServiceChangeTracker};
+use util::megaphone::{ClientServices, MegaphoneAPIResponse, Service, ServiceClientInit, ServiceChangeTracker};
 
 mod dispatch;
 mod metrics;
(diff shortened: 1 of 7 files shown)


autopush_rs/src/util/megaphone.rs
```diff
@@ -114,11 +122,27 @@ impl ServiceChangeTracker {
         svc_change_tracker
     }
 
+    /// Creates a new `ServiceChangeTracker` initialized from a Megaphone API server version set
+    /// as provided as the fetch URL.
+    ///
+    /// This method uses a synchronous HTTP call.
+    pub fn with_api_services(url: &str) -> reqwest::Result<ServiceChangeTracker> {
+        let client = reqwest::Client::builder()
```
alexcrichton: I wonder if perhaps the `Client` may be best stored in the ambient server configuration rather than created on each iteration?

autopush_rs/src/server/mod.rs
```diff
@@ -510,6 +539,76 @@ impl Drop for Server {
...
+                            debug!("Fetched broadcasts: {:?}", broadcasts);
+                            let mut broadcaster = self.srv.broadcaster.borrow_mut();
+                            for srv in Service::from_hashmap(broadcasts) {
+                                broadcaster.add_service(srv);
+                            }
+                        }
+                        Ok(Async::NotReady) => return Ok(Async::NotReady),
+                        Err(_) => {
+                            // TODO: Flag sentry that we can't poll megaphone API
+                            debug!("Failed to get response, queue again");
```
alexcrichton: This may want to log the error as well just in case (same w/ the `map_err` above and dropping the error)

autopush_rs/src/util/megaphone.rs
bbangert: This one is a one-off startup request, so I didn't store the client. The recurring updater does store the client between requests though.

autopush_rs/src/server/mod.rs
```diff
@@ -510,6 +539,76 @@ impl Drop for Server {
...
+
+impl MegaphoneUpdater {
+    fn new(uri: &str, poll_interval: Duration, srv: &Rc<Server>) -> io::Result<MegaphoneUpdater> {
+        let client = reqwest::unstable::async::Client::builder()
+            .timeout(Duration::from_secs(1))
+            .build(&srv.handle)
+            .expect("Unable to build reqwest client".into());
+        Ok(MegaphoneUpdater {
+            srv: srv.clone(),
+            api_url: String::from(uri),
```
pjenvey: prefer uri.to_string() because we use it everywhere else

autopush_rs/src/server/mod.rs
```diff
@@ -510,6 +539,76 @@ impl Drop for Server {
     }
 }
 
+enum MegaphoneState {
```
pjenvey: pretty tempted to have these additions live elsewhere but it's fine for now, sitting alongside PingManager. I think we'll probably be shuffling around various things later

bbangert: Indeed, though they are so bound up in the caller implementation with futures and the server that this seemed the most appropriate place.

PR #1158, opened 2018-03-16
bug: add missing setup_module() function to test_db

### LINKED ISSUE #1157: Add setup_module() call for pytest
test_db.py is missing a `setup_module()` call. This should call the existing `setUp()` module setup function.

jrconlin:
Closes #1157

### DIFF
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -37,7 +37,7 @@ dummy_chid = str(uuid.UUID("deadbeef00000000decafbad00000000"))
 test_router = None
 
 
-def setUp():
+def setup_module():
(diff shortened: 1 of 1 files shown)


autopush/tests/test_db.py
```diff
@@ -37,6 +37,11 @@
 test_router = None
 
 
+def setup_module():
+    """Set up pytest module"""
+    setUp()
+
+
 def setUp():
```
pjenvey: you can rename setUp -> setup_module instead, nose recognizes both.

jrconlin: Oh, awesome. better fix.

PR #1159, opened 2018-03-16
feat: add megaphone integration tests

### LINKED ISSUE #1129: Create megaphone integration

bbangert:
Closes #1129

### DIFF
--- a/autopush/main.py
+++ b/autopush/main.py
@@ -332,7 +332,7 @@ class RustConnectionApplication(AutopushMultiService):
         self.push_server = WebPushServer(
             self.conf,
             self.db,
-            num_threads=num_threads
+            num_threads=num_threads,
         )
 
     def run(self):  # pragma: nocover
@@ -381,5 +381,6 @@ class RustConnectionApplication(AutopushMultiService):
             close_handshake_timeout=ns.close_handshake_timeout,
             aws_ddb_endpoint=ns.aws_ddb_endpoint,
             megaphone_api_url=ns.megaphone_api_url,
+            megaphone_poll_interval=ns.megaphone_poll_interval,
             resource=resource
         )
--- a/autopush/tests/test_rs_integration.py
+++ b/autopush/tests/test_rs_integration.py
@@ -7,16 +7,22 @@ Differences from original integration test:
    last message as production currently uses.
 
 """
+import json
 import logging
 import os
+import re
+import socket
 import time
 import uuid
 from contextlib import contextmanager
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from httplib import HTTPResponse  # noqa
 from mock import Mock, call
+from threading import Thread, Event
 from unittest.case import SkipTest
 
 import ecdsa
+import requests
 import twisted.internet.base
 from cryptography.fernet import Fernet
 from typing import Optional  # noqa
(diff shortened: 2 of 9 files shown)


autopush/tests/test_rs_integration.py
```diff
@@ -45,12 +51,38 @@
...
+    services = {}
+    polled = Event()
+
+    def do_GET(self):
+        if re.search(self.API_PATTERN, self.path):
+            self.send_response(requests.codes.ok)
+            self.send_header('Content-Type', 'application/json; charset=utf-8')
+            self.end_headers()
+            response_content = json.dumps(
+                {"broadcasts": MockMegaphoneRequestHandler.services}
```
jrconlin: not `self.services`?

autopush/tests/test_rs_integration.py
```diff
@@ -81,21 +113,38 @@ class TestRustWebPush(unittest.TestCase):
     )
 
     def setUp(self):
```
jrconlin: to avoid some pytest issues, might want to rename this to `setup_class` 

see https://docs.pytest.org/en/2.7.3/xunit_setup.html#class-level-setup-teardown and http://nose.readthedocs.io/en/latest/writing_tests.html

autopush/tests/test_rs_integration.py
bbangert: I was having a difficult time remembering if that would get a new local instance copy vs. the class copy, self.services should work since it is never assigned to in the instance I think?

autopush/tests/test_rs_integration.py
bbangert: Hmm, we're using setUp everywhere though?

pjenvey:
> Each one of these adds about 2-3 seconds to run-time

I forgot that init_logging would be called on every test run so get_ec2_instance_id is probably the cause of one of those seconds

autopush/main.py
```diff
@@ -323,7 +323,7 @@ class RustConnectionApplication(AutopushMultiService):
     logger_name = "AutopushRust"
     push_server = None
 
-    def setup(self, rotate_tables=True, num_threads=10):
+    def setup(self, rotate_tables=True, num_threads=10, json_logging=True):
```
pjenvey: you should put the argparse'd human_logs setting into AutopushConfig. Then AutopushServer can set its cfg.json_logging from it. Then you won't need the various json_logging arguments/passthrus

autopush_rs/__init__.py
```diff
@@ -19,7 +19,7 @@ def free(obj, free_fn):
 
 
 class AutopushServer(object):
-    def __init__(self, conf, queue):
+    def __init__(self, conf, queue, json_logging=True):
```
pjenvey: kill :feelsgood:

autopush/main.py
```diff
@@ -332,7 +332,7 @@ def setup(self, rotate_tables=True, num_threads=10):
         self.push_server = WebPushServer(
             self.conf,
             self.db,
-            num_threads=num_threads
+            num_threads=num_threads,
```
jrconlin: not sure why you added this, but not going to ding either.

autopush/tests/test_rs_integration.py
pjenvey: FYI this is a twisted trial TestCase, which is calling the setUp, regardless of test runner

autopush/tests/test_rs_integration.py
```diff
@@ -111,7 +144,7 @@ def setUp(self):
         # Websocket server
         self.conn = conn = RustConnectionApplication(
             conn_conf,
-            resource=autopush.tests.boto_resource
+            resource=autopush.tests.boto_resource,
```
jrconlin: likewise, not sure why you added the extra ','

PR #1166, opened 2018-03-27
feat: add megaphone api token auth

### LINKED ISSUE #1164: Add authentication token for reading megaphone API
Megaphone API uses a basic bearer token in the Authentication header. The connection node needs to include this from config to access the API.

bbangert:
Closes #1164

### DIFF
--- a/autopush/main_argparse.py
+++ b/autopush/main_argparse.py
@@ -231,6 +231,9 @@ def parse_connection(config_files, args):
     parser.add_argument('--megaphone_api_url',
                         help="The megaphone API URL to query for updates",
                         default=None, type=str, env_var="MEGAPHONE_API_URL")
+    parser.add_argument('--megaphone_api_token',
+                        help="The megaphone API token",
+                        default=None, type=str, env_var="MEGAPHONE_API_TOKEN")
     parser.add_argument('--megaphone_poll_interval',
                         help="The megaphone API polling interval",
                         default=30, type=int,
--- a/autopush/tests/test_rs_integration.py
+++ b/autopush/tests/test_rs_integration.py
@@ -69,9 +69,11 @@ class MockMegaphoneRequestHandler(BaseHTTPRequestHandler):
     API_PATTERN = re.compile(r'/v1/broadcasts')
     services = {}
     polled = Event()
+    token = "Bearer {}".format(uuid.uuid4().hex)
 
     def do_GET(self):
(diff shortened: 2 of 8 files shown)


autopush/main_argparse.py
```diff
@@ -231,6 +231,9 @@ def parse_connection(config_files, args):
     parser.add_argument('--megaphone_api_url',
                         help="The megaphone API URL to query for updates",
                         default=None, type=str, env_var="MEGAPHONE_API_URL")
+    parser.add_argument('--megaphone_api_token',
+                        help="The megaphone API URL to query for updates",
```
pjenvey: copy pasta

autopush_rs/src/server/mod.rs
```diff
@@ -326,7 +330,10 @@ impl Server {
     fn new(opts: &Arc<ServerOptions>, tx: queue::Sender) -> Result<(Rc<Server>, Core)> {
         let core = Core::new()?;
         let broadcaster = if let Some(ref megaphone_url) = opts.megaphone_api_url {
-            ServiceChangeTracker::with_api_services(megaphone_url)
+            let megaphone_token = opts.megaphone_api_token.as_ref().expect(
+                "Megaphone API requires a Megaphone API Token to be set".into()
```
pjenvey: nitpick .into() isn't necessary in these 2 expects. I guess we have them elsewhere but they aren't needed

autopush/tests/test_rs_integration.py
```diff
@@ -69,9 +69,11 @@ class MockMegaphoneRequestHandler(BaseHTTPRequestHandler):
     API_PATTERN = re.compile(r'/v1/broadcasts')
     services = {}
     polled = Event()
+    token = uuid.uuid4().hex
```
pjenvey: Note that the auth header should be "Bearer sometoken". This patch requires "Bearer " to be specified along side the token in the config.

megaphone sort of has to hardcode "Bearer" in its code I think. I'd lean towards these 2 configs matching. If not, at the very least I'd call this config value a "header" vs a "token" and document it w/ an example. and probably include a "Bearer " on this line as well

(autopush_connection.ini could also probably use an example)

jrconlin:
Well, if you want to get SUPER Picky, the schema + token bit are "credentials" according to the RFCs. 

Please don't be super picky.

PR #1169, opened 2018-03-31
feat: refactor client to remove duplicate logic

### LINKED ISSUE #1165: Refactor out one of the waiting states
The rust connection node currently has two waiting states with almost identical source code. They should be combined into a single command waiting state, that does the ack check up front, and appropriately ignores ack's when they shouldn't be waiting. This would remove the duplicate code between the Await and WaitingForAcks states.

bbangert:
WaitingForAck's was essentially identical to Await, except it allowed
for Ack's to come in. Combining these reduces duplication and errors
that may occur as a result while simplifying command handling.

Closes #1165

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -135,7 +135,6 @@ pub enum ClientState {
     SendMessages(Option<Vec<Notification>>),
     CheckStorage,
     IncrementStorage,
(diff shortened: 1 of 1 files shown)


bbangert:
I'll admit I'm puzzled by the coverage change.

PR #1171, opened 2018-04-04
feat: update dependencies

### LINKED ISSUE #1170: Update dependencies [Q2]

pjenvey:
both python and rust side

hold off on a woothee-rust update for now in case the API changes back

Closes #1170

### DIFF
--- a/autopush_rs/Cargo.lock
+++ b/autopush_rs/Cargo.lock
@@ -12,11 +12,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.6"
+name = "arrayvec"
+version = "0.4.7"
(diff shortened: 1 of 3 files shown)


pjenvey:
> I wonder if we should drop Cargo.lock from the repo. It's a work file, or do we need it for CI?

We definitely want it in for a reproducible build. Cargo recommends leaving it out for pure libraries but not for binaries/apps: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

pjenvey:
I'll also note we don't really need specific minor versions included in Cargo.toml because they're locked already by Cargo.lock.

I switched our Cargo.toml to use minor versions only because cargo upgrade does it that way. I think there may have been some talk on its issue tracker about leaving the minor version off

PR #1178, opened 2018-04-15
feat: fix asserts on client dis/connect and convert inc call to rust

### LINKED ISSUE #1177: Avoid asserts connecting/disconnecting clients
Connecting clients uses an assert, and we should disconnect the client if its the same uaid.

bbangert:
Removes the asserts when adding the uaid to the uaids hash, and issue
proper disconnect call to existing client. Ensure uaid removal removes
the right client indicated by a separate UUID per client object.

Convert IncrementStorage call to pure Rust async and remove Python
implementation.

Closes #1177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
 - pip install tox ${CODECOV:+codecov}
 - if [ ${WITH_RUST:-true} != "false" ]; then curl https://sh.rustup.rs | sh -s -- -y || travis_terminate 1; fi
 - export PATH=$PATH:$HOME/.cargo/bin
-- export AWS_SHARED_CREDENTIALS_FILE=./automock/credentials.cfg
+- export AWS_SHARED_CREDENTIALS_FILE=./automock/boto.cfg
 - export BOTO_CONFIG=./automock/boto.cfg
 script:
 - tox -- ${CODECOV:+--with-coverage --cover-xml --cover-package=autopush}
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -14,6 +14,9 @@ use futures::future::Either;
 use futures::sync::mpsc;
 use futures::sync::oneshot::Receiver;
 use futures::{Stream, Sink, Future, Poll, Async};
+use futures_backoff::retry_if;
+use rusoto_dynamodb::{AttributeValue, DynamoDb};
+use rusoto_dynamodb::{UpdateItemInput, UpdateItemOutput, UpdateItemError};
 use tokio_core::reactor::Timeout;
 use time;
 use uuid::Uuid;
@@ -26,8 +29,11 @@ use server::Server;
 use util::parse_user_agent;
 use util::megaphone::{ClientServices, Service, ServiceClientInit};
 
+const MAX_EXPIRY: u64 = 2592000;
+
 pub struct RegisteredClient {
     pub uaid: Uuid,
+    pub uid: Uuid,
     pub tx: mpsc::UnboundedSender<ServerNotification>,
 }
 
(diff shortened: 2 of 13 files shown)


autopush_rs/src/server/mod.rs
```diff
@@ -338,9 +342,19 @@ impl Server {
         } else {
             ServiceChangeTracker::new(Vec::new())
         };
+        let region = if env::var("AWS_PROFILE").unwrap_or("default".to_string()) == "test" {
```
jrconlin: use `if cfg!(test)` here?

autopush_rs/src/client.rs
```diff
@@ -642,6 +649,42 @@ where
         ClientState::WaitingForUnRegister(channel_id, fut)
     }
 
+    fn process_increment_storage(&mut self) -> ClientState {
+        // Let the variable copies begin, so that nothing is left dangling in the future
+        // we then assemble because Lifetimes.
+        let webpush = self.webpush.as_ref().unwrap();
```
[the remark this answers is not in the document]
bbangert: Logically to get to the point where we can do state transitions, we need a webpush here, same with how the state machine guarantee's we'll have unacked by this point or we wouldn't need to increment storage.

autopush_rs/src/server/mod.rs
bbangert: Do we set what ever that checks somewhere?

jrconlin: `cfg!()` can be called at run time to see what configuration you happen to be in. It's the inline equivalent of "#[cfg(test)]". 

It might be more useful than checking the AWS_PROFILE for a test configuration value.

bbangert: I mean, how do we flag in the environment that its in test mode? It's not running a Rust based test suite that would flag this afaik.

pjenvey: Right, the #cfg test toggle is via a compiler flag, so unfortunately we can't use it here.

The python code keys on AWS_LOCAL_DYNAMODB, so I'd rather this key on that as well. Then I don't think the changes to .travis.yml/boto.cfg/tox.ini are necessary (and local testing should continue to work as normal)

autopush_rs/src/server/mod.rs
```diff
@@ -338,9 +342,19 @@ impl Server {
...
+            Region::Custom {
+                endpoint: "http://localhost:8000".to_string(),
+                name: "local_dynamodb".to_string(),
+            }
+        } else {
+            Region::UsEast1
```
pjenvey: don't hardcode, you want Region::default() so it's read from the AWS_DEFAULT_REGION env var

autopush_rs/src/server/mod.rs
```diff
@@ -338,9 +342,17 @@ impl Server {
         } else {
             ServiceChangeTracker::new(Vec::new())
         };
+        let region = env::var("AWS_LOCAL_DYNAMODB").map(|endpoint| {
```
pjenvey: nitpick: unwrap_or_else would be preferred here (don't care much because this is for tests)

.travis.yml
```diff
@@ -28,7 +28,7 @@ install:
 - pip install tox ${CODECOV:+codecov}
 - if [ ${WITH_RUST:-true} != "false" ]; then curl https://sh.rustup.rs | sh -s -- -y || travis_terminate 1; fi
 - export PATH=$PATH:$HOME/.cargo/bin
-- export AWS_SHARED_CREDENTIALS_FILE=./automock/credentials.cfg
+- export AWS_SHARED_CREDENTIALS_FILE=./automock/boto.cfg
```
pjenvey: is this still necessary? (boto.cfg should probably be killed at this point if anything)

bbangert: It is, because rusoto wants to find itself some aws access/secret keys, and if it can't find them in a file it expects, it starts issuing HTTP calls looking for IAM keys and such which messes things up. This is also why I added random access/secret keys to the boto.cfg.

autopush_rs/src/server/mod.rs
bbangert: But it needs to construct the Region object with the value?

.travis.yml
bbangert: What credentials.cfg? https://github.com/mozilla-services/autopush/tree/master/automock

PR #1179, opened 2018-04-16
feat: add flag to stop table rotation

### LINKED ISSUE #1172: Halt table rotation for TTL for Autopush
- [ ] Add config to halt table rotation.
- [ ] Add code to prevent month date updates, can use last table set.

jrconlin:
Closes #1172


bbangert:
This has a conflict with master branch.

autopush/webpush_server.py
```diff
@@ -466,7 +471,7 @@ def process(self, command):
     def _check_storage(self, command):
         timestamp = None
         messages = []
-        message = Message(command.message_month,
+        message = Message(self.get_month(command),
```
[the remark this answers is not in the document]
jrconlin: good call on caching the latest message tablename, That can probably go into the DynamoDBResource.

as for the prefix, we already specify "--message_tablename", this is stowed as `self._message_conf.tablename` and is used as the `tablename_base_string=` in `DatabaseManager.setup_tables()`. I'm not sure of a better parameter to use to specify it, but I'm happy to switch to something better.

autopush/db.py
```diff
@@ -466,6 +485,17 @@ def __init__(self, **kwargs):
     def __getattr__(self, name):
         return getattr(self._resource, name)
 
+    def get_latest_message_tablename(self, prefix="message"):
+        # type: (Optional[str]) -> str  # noqa
+        """Fetches the name of the last message table"""
+        response = filter(
+            lambda name: name.lower().startswith(prefix),
+            self._resource.meta.client.list_tables().get("TableNames"))
```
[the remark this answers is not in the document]
jrconlin: It's my understanding that list_tables will return a maximum of 100 entries per call. We rotate tables each month. Do we feel that we will be doing table rotation for 8 years?

autopush/db.py
```diff
@@ -1078,6 +1117,21 @@ def create_initial_message_tables(self):
         an entry for tomorrow, if tomorrow is a new month.
 
         """
+        if not self.allow_table_rotation:
+            tablename = self.resource.get_latest_message_tablename(
```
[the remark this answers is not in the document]
jrconlin: If I understand correctly, `self._message_conf.tablename` is the config option to specify the root name for the message table. Here, it's used as a mask to determine the message table from the returned list of tables.

the following table creation methods will probably not be called outside of test routines, since the table will have already been created.

autopush/db.py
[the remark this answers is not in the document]
pjenvey: it lists all tables for "the current account and endpoint" which is *all* the tables. there's > 100 on dev and stage. I've actually ran into this on dev at one point and fixed all our cases of it, before we got unlucky on stage/prod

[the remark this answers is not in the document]
jrconlin: Oh. awesome. Yeah, that makes sense.

jrconlin:
The presumption is that they're kept at the last table that we rotate to. Once the flag is flipped, no further rotation happens. 

When we decide we no longer need the flag, all the rotation stuff can be stripped out, and we can even config the message table name. The flag tries to maintain as much of the rotation code as possible so that the switch can be maintained or quickly reverted in case of unexpected issues.

autopush/db.py
```diff
@@ -466,6 +486,23 @@ def __init__(self, **kwargs):
...
 
+    def get_latest_message_tablename(self, prefix="message"):
+        # type: (Optional[str]) -> str  # noqa
+        """Fetches the name of the last message table"""
+        if not self._latest_message_table:
+            client = self._resource.meta.client
+            paginator = client.get_paginator("list_tables")
+            tables = []
+            for table in paginator.paginate().search(
+                    "TableNames[?contains(@,'{}')==`true`]|sort(@)[-1]".format(
```
bbangert: Contains? Is this going to require it as the prefix or will it find the prefix anywhere in the table name?

bbangert: Right now our table prpefix's look like this "{dev/stage/prod}.autopush.message", so it should work ok for us as Contains, but it'd prolly be safer if it required the table to start with the prefix.

autopush/db.py
```diff
@@ -478,8 +515,9 @@ def __getattr__(self, name):
 
 class Message(object):
     """Create a Message table abstraction on top of a DynamoDB Table object"""
-    def __init__(self, tablename, metrics=None, boto_resource=None,
-                 max_ttl=MAX_EXPIRY):
+    def __init__(self, tablename=None, metrics=None, boto_resource=None,
+                 max_ttl=MAX_EXPIRY,
+                 table_base_string="message_"):
```
bbangert: How is the table_base_string different from the prefix?

autopush/db.py
```diff
@@ -995,19 +1036,23 @@ class DatabaseManager(object):
...
         """Initialize sane defaults"""
-        today = datetime.date.today()
-        self.current_month = today.month
-        self.current_msg_month = make_rotating_tablename(
-            self._message_conf.tablename,
-            date=today
-        )
+        if self.allow_table_rotation:
+            today = datetime.date.today()
+            self.current_month = today.month
```
bbangert: If this is only assigned when allow_table_rotation is True, are we 100% sure that nothing anywhere when its off might try and access the current_month attribute?

autopush/webpush_server.py
```diff
@@ -519,7 +524,9 @@ class MigrateUserCommand(ProcessorCommand):
     def process(self, command):
         # type: (MigrateUser) -> MigrateUserResponse
         # Get the current channels for this month
-        message = Message(command.message_month,
+        if not self.conf.allow_table_rotation:
+            return MigrateUserResponse(message_month=None)
```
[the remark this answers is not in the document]
jrconlin: Added https://github.com/mozilla-services/autopush/pull/1179/files#diff-f37e66b6947f8539b0024d6672bd5ce1R189.

The rust side still calls to python for data operations, so it should never see a TableName of None.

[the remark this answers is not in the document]
pjenvey: I ported over test_webpush_monthly_rotation to test_rs_integration, which will trigger this error:

https://pastebin.mozilla.org/9083770

`
{"Severity":5,"Timestamp":1524526849884206435,"Logger":"AutopushRust-0.1.0","EnvVersion":"2.0","Type":"autopush_rs:log","Hostname":"ubuntu","Fields":{"message":"Error for shutdown of 127.0.0.1: invalid type: null, expected a string at line 1 column 74"}}
`

Of course after all that whats occurs to me: do we even want to be in this code in the first place when allow_table_rotation is False? If not, then we should be disabling that flags["rotate_message_table"] = True in the HelloCommand above

autopush/db.py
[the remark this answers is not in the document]
pjenvey: my fault, I misunderstood that we'd be going w/ the final dated table after toggling allow_table_rotation as False

configs/autopush_shared.ini.sample
```diff
@@ -132,3 +132,6 @@ endpoint_port = 8082
 ; e.g {"firefox":{"cert":"certs/main.cert","key":"certs/main.key","topic":"com.mozilla.org.Firefox","max_retry":2},"beta":{"cert":"certs/beta.cert","key":"certs/beta.key","topic":"com.mozilla.org.FirefoxBeta"}}
 #apns_creds =
 
+; With TTL implemented, message table rotation is no longer required.
+; This flag determines if table rotation should be allowed to continue:
+#allow_table_rotation = true
```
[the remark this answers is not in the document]
jrconlin: https://github.com/mozilla-services/autopush/pull/1179/files#diff-05700f34cfdcdebb78733625f5c8df92R108

autopush/db.py
jrconlin: There's not a huge difference between the two, aside from `prefix` is used mostly in table rotation, where `table_base_string` is not.

autopush/db.py
```diff
@@ -995,19 +1036,23 @@ class DatabaseManager(object):
 
     router = attrib(default=None)                   # type: Optional[Router]
     message_tables = attrib(default=Factory(list))  # type: List[str]
-    current_msg_month = attrib(init=False)          # type: Optional[str]
+    current_msg_month = attrib(default=None,
```
pjenvey: this becomes None w/ allow_table_rotation = False. Instead it should be set to the latest pegged tablename (from get_latest_message_tablename). This should really be where we cache that value.

This shouldn't be None because it propagates everywhere (through all the rust Command/Responses, which right now assumes it's a String vs Option<String>. Also the "current_month" in the router table)

autopush/webpush_server.py
```diff
@@ -408,15 +413,16 @@ def lookup_user(self, hello):
         flags["check_storage"] = True
 
         # Determine if message table rotation is needed
-        flags["message_month"] = record["current_month"]
```
pjenvey: we still need to continue migrating to the new pegged table, so revert this. my bad!

autopush/webpush_server.py
```diff
@@ -348,6 +348,11 @@ def metrics(self):
     def process(self, command):
         raise NotImplementedError()
 
+    def get_month(self, command):
```
pjenvey: I think we'll just be passing through the db.current_msg_month so I think this should go away completely? update_rotating_tables will never change it

autopush/websocket.py
```diff
@@ -786,10 +786,11 @@ def _verify_user_record(self):
...
         # Include and update last_connect if needed, otherwise exclude
-        if has_connected_this_month(record):
-            del record["last_connect"]
-        else:
-            record["last_connect"] = generate_last_connect()
+        if self.conf.allow_table_rotation:
+            if has_connected_this_month(record):
+                del record["last_connect"]
+            else:
+                record["last_connect"] = generate_last_connect()
```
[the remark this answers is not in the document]
pjenvey: this should be reverted along w/ the similar webpush_server.py change and its accompanying tweak to trigger the table migration (my bad)

autopush/db.py
```diff
@@ -466,6 +486,31 @@ def __init__(self, **kwargs):
...
+            tables.append(table)
+        if not len(tables) or tables[0] is None:
+            return [prefix]
+        tables.sort()
+        return tables[0-previous:]
+
+    def get_latest_message_tablename(self, prefix="message"):
+        # type: (Optional[str]) -> str  # noqa
+        """Fetches the name of the last message table"""
+        if not self._latest_message_table:
```
pjenvey: we only call these during initialization now, so let's not cache them here

autopush/main_argparse.py
```diff
@@ -105,6 +105,10 @@ def add_shared_args(parser):
                         help="AWS DynamoDB endpoint override",
                         type=str, default=None,
                         env_var="AWS_LOCAL_DYNAMODB")
+    parser.add_argument('--allow_table_rotation',
+                        help="Allow monthly message table rotation",
+                        action="store_true", default=False,
```
pjenvey: default to False here?

Also this isn't being passed along to AutopushConfig. Annoyingly enough, this still needs a line in AutopushConfig::from_argparse

jrconlin: crap, yeah, forgot that.

jrconlin:
changed the name (and a bit of logic) to the argument.

the option `--no_table_rotation` will set the config `allow_table_rotation=False`. This will allow the app to continue to work as expected UNTIL the `--no_table_rotation` flag is set.

(it also makes the code a bit more readable rather than `not config.no_table_rotation` everywhere)

autopush/config.py
```diff
@@ -297,6 +299,9 @@ def from_argparse(cls, ns, **kwargs):
         if not ns.no_aws:
             ami_id = get_amid() or "Unknown"
 
+        import pdb;pdb.set_trace()
```
pjenvey: woops

autopush/db.py
```diff
@@ -458,6 +477,7 @@ def __init__(self, **kwargs):
         if "region_name" in conf:
             del(conf["region_name"])
         self.conf = conf
+        self._latest_message_table = None
```
pjenvey: kill

autopush/tests/test_db.py
```diff
@@ -84,6 +84,47 @@ def test_init_with_resources(self):
...
+            )
+
+    def test_init_with_no_rotate_create_table(self):
+        fake_conf = Mock()
+        fake_conf.allow_table_rotation = False
+        fake_conf.message_table = Mock()
+        fake_conf.message_table.tablename = "message_bogus"
+        fake_conf.message_table.read_throughput = 5
+        fake_conf.message_table.write_throughput = 5
+        safe = autopush.tests.boto_resource._latest_message_table
```
pjenvey: kill

autopush/tests/test_db.py
```diff
@@ -84,6 +84,47 @@ def test_init_with_resources(self):
...
+        dm = DatabaseManager.from_config(
+            fake_conf,
+            resource=autopush.tests.boto_resource)
+        try:
+            autopush.tests.boto_resource._latest_message_table = None
```
pjenvey: kill

autopush/config.py
jrconlin: also helps if I merge and push changes instead of wandering off to get a sandwich. ._.

PR #1191, opened 2018-04-27
feat: use chrono epoch times instead of monotonic time

### LINKED ISSUE #1180: Use chrono functions for time
We use an old time/chrono crate on occasion, we should use std::time now.

bbangert:
Closes #1180

### DIFF
--- a/autopush_rs/src/util/mod.rs
+++ b/autopush_rs/src/util/mod.rs
@@ -2,6 +2,7 @@
 use std::io;
 use std::time::Duration;
 
+use chrono::Utc;
 use hostname::get_hostname;
 use futures::future::{Either, Future, IntoFuture};
 use slog;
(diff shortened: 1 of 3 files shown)


autopush_rs/src/util/mod.rs
```diff
@@ -92,6 +93,24 @@ pub fn init_logging(json: bool) {
     slog_stdlog::init().ok();
 }
 
+/// Get the time since the UNIX epoch in seconds
+pub fn sec_since_epoch() -> u64 {
+    Utc::now().timestamp() as u64
+}
+
+/// Get the time since the UNIX epoch in seconds
```
pjenvey: s/seconds/milliseconds

autopush_rs/src/util/mod.rs
```diff
@@ -92,6 +93,24 @@ pub fn init_logging(json: bool) {
...
+pub fn ms_since_epoch() -> u64 {
+    let now = Utc::now();
+    (now.timestamp() as u64) * 1_000 + (now.timestamp_millis() as u64)
```
pjenvey: you meant timestamp_subsec_millis here, similar to us_since_epoch's impl. However timestamp_millis does all of this for us, so this whole fn should simply be: `Utc::now().timestamp_millis() as u64`

PR #1192, opened 2018-04-27
Simplify some code

alex:


alex:
Nevermind. For some reason I thought this proejct was py3 only.

PR #1210, opened 2018-05-04
bug: Double channel regsitration ID expiration

### LINKED ISSUE #1199: Registering a channel should use 2x MAX_EXPIRY
To better emulate the current system, we should have `register_channel` use 2 x MAX_EXPIRY.

jrconlin:
Closes #1199

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -544,7 +544,7 @@ class Message(object):
         """Register a channel for a given uaid"""
         # Generate our update expression
         if ttl is None:
-            ttl = self._max_ttl
+            ttl = self._max_ttl * 2
         expr_values = {
             ":channel_id": set([normalize_id(channel_id)]),
             ":expiry": _expiry(ttl)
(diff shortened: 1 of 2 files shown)


autopush/tests/test_db.py
```diff
@@ -528,8 +549,6 @@ def test_no_uaid_found(self):
     def test_uaid_provision_failed(self):
         router = Router(self.table_conf,  SinkMetrics(),
                         resource=self.resource)
-        router.table = Mock()
```
jrconlin: removing this because we set the router.table to a new mock below. It's a noop.

autopush/db.py
```diff
@@ -841,8 +841,13 @@ def register_user(self, data):
...
             raise AutopushException("data is missing router_type "
                                     "or connected_at")
+        # Mobile users set this value to "0" which means that they are
+        # never dropped from DynamoDB (DDB only drops values that have a
+        # "reasonable" expiration date within 2 years or so of present.)
+        # We double the expiration to match the previous behavior of
+        # idle registrations lasting about 2 rotations.
         if "expiry" not in data:
-            data["expiry"] = _expiry(MAX_EXPIRY)
+            data["expiry"] = _expiry(MAX_EXPIRY * 2)
```
pjenvey: Probably should have asked this during #1162, but now that we don't enforce the expiry on the router table, do we even need the field at all? I don't see anything actually being done w/ it..

jrconlin: Excellent point. It should be removed so that it's not confusing folks. 

Thank you for pointing out the blindingly obvious. 

/me sighs heavily and goes about doing just that.

PR #1214, opened 2018-05-04
feat: hardcode autopush_rs's AWS creds in local dynamodb mode

### LINKED ISSUE #1201: hardcode autopush_rs tests' credentials
autopush_rs shouldn't need to specify any special env vars such as AWS_SHARED_CREDENTIALS_FILE for the tests to run, like the python side.

pjenvey:
and kill now unneeded automock / boto.cfg files and the old
Dockerfile.automock that nobody uses any more

Closes #1201


pjenvey:
> Why'd it have to go in the Box?

The if else block causes the DynamoDbClient struct to be different enough (its associated types) that we have to refer to it via its trait now (DynamoDb). So without impl trait we have to box it to return it.

PR #1220, opened 2018-05-08
feat: port hello command to Rust

### LINKED ISSUE #1188: Convert hello command from Python to Rust

bbangert:
Co-authored by: Phil Jenvey <pjenvey@underboss.org>

Closes #1188

### DIFF
--- a/autopush_rs/src/util/ddb_helpers.rs
+++ b/autopush_rs/src/util/ddb_helpers.rs
@@ -2,24 +2,34 @@
 use std::env;
 use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
+use std::result::Result as StdResult;
 use uuid::Uuid;
 
+use cadence::{Counted, StatsdClient};
+use chrono::Utc;
 use futures::Future;
 use futures::future;
 use futures_backoff::retry_if;
+use rand;
+use rand::distributions::{IndependentSample, Range};
 use regex::RegexSet;
 use rusoto_core::Region;
 use rusoto_core::reactor::RequestDispatcher;
 use rusoto_credential::StaticProvider;
-use rusoto_dynamodb::{AttributeValue, DynamoDb, DynamoDbClient, QueryError, QueryInput,
+use rusoto_dynamodb::{AttributeValue, DeleteItemError, DeleteItemInput, DeleteItemOutput,
+                      DynamoDb, DynamoDbClient, GetItemError, GetItemInput, GetItemOutput,
+                      PutItemError, PutItemInput, PutItemOutput, QueryError, QueryInput,
                       UpdateItemError, UpdateItemInput, UpdateItemOutput};
+use serde::Serializer;
 use serde_dynamodb;
 use time;
 
 use protocol::Notification;
 use errors::*;
+use util::timing::ms_since_epoch;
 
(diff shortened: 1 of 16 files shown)


autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -501,6 +566,241 @@ impl DynamoStorage {
...
+                ddb.delete_item(&DeleteItemInput {
+                    table_name: table_name.clone(),
+                    key: ddb_item! {
+                        uaid: s => uaid.clone()
+                    },
+                    ..Default::default()
+                })
+            },
+            |err: &DeleteItemError| {
+                matches!(err, &DeleteItemError::ProvisionedThroughputExceeded(_))
```
alexcrichton: This retry loop I've seen in a number of places, it may be a little more ergonomic to have something like:

```rust
trait Retry {
    fn retry(&self) -> bool;
}

pub fn retrying<F>(f: F) -> impl Future<Item = F::Item, Error = F::Error> 
    where F: Future, F::Error: Retry
{
    retry_if(...)
}

impl Retry for DeleteItemError { ... }
impl Retry forGetItemError { ... }
```

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -501,6 +566,241 @@ impl DynamoStorage {
...
+        let table_name = router_table_name.to_string();
+        let response = retry_if(
+            move || {
+                ddb.delete_item(&DeleteItemInput {
+                    table_name: table_name.clone(),
+                    key: ddb_item! {
+                        uaid: s => uaid.clone()
+                    },
+                    ..Default::default()
+                })
```
alexcrichton: You can probably avoid some explicit captures outside this closure by creating the `DeleteItemInput` outside the closure and then moving it into the closure, although that's just a stylistic choice!

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -501,6 +566,241 @@ impl DynamoStorage {
...
+            },
+            |err: &DeleteItemError| {
+                matches!(err, &DeleteItemError::ProvisionedThroughputExceeded(_))
+            },
+        ).map_err(|_| "Error fetching user".into());
```
alexcrichton: I'd recommend not losing the original error context here as it can often make things more difficult to debug. I think with error-chain there's a `chain_err` combinator for futures, right?

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -501,6 +566,241 @@ impl DynamoStorage {
...
+                            attribute_not_exists(node_id) or
+                            (connected_at < :connected_at)
+                        )"#.to_string(),
+                        ),
+                        return_values: Some("ALL_OLD".to_string()),
+                        ..Default::default()
+                    })
+                },
+                |err: &PutItemError| matches!(err, &PutItemError::ProvisionedThroughputExceeded(_)),
+            ).map_err(|e| format!("Error storing user record: {:?}", e).into());
```
alexcrichton: Similar to above using `chain_err` allows you to losslessly retain the original error context (this implementation may lose some info)

PR #1222, opened 2018-05-10
feat: port register command to Rust

### LINKED ISSUE #1190: Convert register command from Python to Rust

bbangert:
Closes #1190

### DIFF
--- a/autopush/webpush_server.py
+++ b/autopush/webpush_server.py
@@ -266,14 +266,12 @@ class CommandProcessor(object):
         self.delete_message_processor = DeleteMessageCommand(conf, db)
         self.drop_user_processor = DropUserCommand(conf, db)
         self.migrate_user_proocessor = MigrateUserCommand(conf, db)
-        self.register_process = RegisterCommand(conf, db)
         self.unregister_process = UnregisterCommand(conf, db)
         self.store_messages_process = StoreMessagesUserCommand(conf, db)
         self.deserialize = dict(
             delete_message=DeleteMessage,
             drop_user=DropUser,
             migrate_user=MigrateUser,
-            register=Register,
             unregister=Unregister,
             store_messages=StoreMessages,
         )
@@ -281,7 +279,6 @@ class CommandProcessor(object):
             delete_message=self.delete_message_processor,
             drop_user=self.drop_user_processor,
             migrate_user=self.migrate_user_proocessor,
-            register=self.register_process,
             unregister=self.unregister_process,
             store_messages=self.store_messages_process,
         )  # type: Dict[str, ProcessorCommand]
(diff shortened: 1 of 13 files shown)


autopush_rs/src/client.rs
```diff
@@ -733,17 +733,15 @@ where
                 }
             }
             Either::A(ClientMessage::Register { channel_id, key }) => {
+                data.srv.metrics.incr("ua.command.register").ok();
```
pjenvey: there was also an associated log.info here in python

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -631,6 +648,37 @@ impl DynamoStorage {
...
     }
 
+    fn register_channel_id(
+        ddb: Rc<Box<DynamoDb>>,
+        uaid: &Uuid,
+        channel_id: &Uuid,
+        message_table_name: &str,
+    ) -> MyFuture<UpdateItemOutput> {
+        let chid = channel_id.hyphenated().to_string();
+        let expiry = (time::get_time().sec as u64) + MAX_EXPIRY;
```
pjenvey: prefer util::sec_since_epoch

autopush_rs/src/server/mod.rs
```diff
@@ -485,6 +504,29 @@ impl Server {
...
+        let root = format!("{}/wpush/", self.opts.endpoint_url);
+        let mut base = hex::decode(uaid.simple().to_string()).chain_err(|| "Error decoding")?;
+        base.extend(hex::decode(chid.simple().to_string()).chain_err(|| "Error decoding")?);
+        if let Some(k) = key {
+            let raw_key = base64::decode_config(&k, base64::URL_SAFE)
+                .chain_err(|| "Error encrypting payload")?;
+            let key_digest = hash::hash(hash::MessageDigest::sha256(), &raw_key)
+                .chain_err(|| "Error creating message digest for key")?;
+            base.extend(key_digest.iter());
+            let encrypted = self.opts.fernet.encrypt(&base);
```
pjenvey: let's trim_matches('=') the result of fernet.encrypt here and below (like the python code), just to be super safe

autopush/webpush_server.py
```diff
@@ -434,58 +431,6 @@ def _validate_chid(chid):
...
-    error_msg = attrib()  # type: str
-    error = attrib(default=True)  # type: bool
-    status = attrib(default=401)  # type: int
-
-
-class RegisterCommand(ProcessorCommand):
-
-    def process(self, command):
-        # type: (Register) -> Union[RegisterResponse, RegisterErrorResponse]
-        valid, msg = _validate_chid(command.channel_id)
```
pjenvey: We have _validate_chid here to mimick websocket.py's validation of the channel_id sent in the register command is hyphenated.

Now I'm wondering if this check is bogus: because the rust side parses the register command now. It's deserializing it via serde into ClientMessage::Register channel_id: Uuid. Does serde's deserialization allow non hyphenated uuids to pass through unnoticed?

autopush_rs/src/client.rs
bbangert: Yup, we never used it and I don't know why we bothered logging it.

autopush_rs/src/server/mod.rs
bbangert: The Python one only strips the encrypted value once.

autopush/webpush_server.py
bbangert: Yes, serde will allow hypenated or not to pass through.

autopush_rs/src/server/mod.rs
pjenvey: I meant below as in the else block

PR #1223, opened 2018-05-10
feat: port drop user command to Rust

### LINKED ISSUE #1207: Convert drop user command from Python to Rust

bbangert:
Closes #1207

### DIFF
--- a/autopush_rs/src/client.rs
+++ b/autopush_rs/src/client.rs
@@ -592,7 +592,7 @@ where
 
     #[state_machine_future(transitions(AuthDone))]
     AwaitDropUser {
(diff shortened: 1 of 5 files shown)


autopush_rs/src/client.rs
```diff
@@ -592,7 +592,7 @@ where
 
     #[state_machine_future(transitions(AuthDone))]
     AwaitDropUser {
-        response: MyFuture<call::DropUserResponse>,
+        response: MyFuture<bool>,
```
pjenvey: we don't look at the result anyway, so use nil () instead

PR #1224, opened 2018-05-10
Feat/issue 1205

bbangert:
The register command PR should be merged and this rebased on top of master before this is reviewed.


autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -869,6 +885,27 @@ impl DynamoStorage {
...
+        channel_id: &Uuid,
+        message_month: &str,
+        code: u32,
+    ) -> MyFuture<bool> {
+        let ddb = self.ddb.clone();
+        let response = DynamoStorage::unregister_channel_id(ddb, uaid, channel_id, message_month)
+            .and_then(move |_| -> MyFuture<_> { Box::new(future::ok(true)) })
+            .or_else(move |_| -> MyFuture<_> { Box::new(future::ok(false)) });
+        if code != 200 {
+            info!("Unregister";
```
pjenvey: did we want the info! for register then? You said we didn't need it in the other PR

autopush_rs/src/client.rs
```diff
@@ -990,22 +994,26 @@ where
     ) -> Poll<AfterAwaitUnregister<T>, Error> {
         debug!("State: AwaitUnRegister");
         let msg = match try_ready!(await_unregister.response.poll()) {
-            call::UnRegisterResponse::Success { success } => {
+            true => {
```
pjenvey: since this is a bool now, I think you'd save a few lines of code making this an if/else instead of a match. especially since both blocks return ServerMessage::Unregister

autopush_rs/src/util/ddb_helpers.rs
bbangert: I don't think we care about successful registers (they all are), but unsubscribes are different cause they can happen for non-user-driven reasons like quota expiration, etc. The Python code always logged the unregisters, which like the registers is a bit pointless. But we definitely want to know about the codes involved in non-user-driven unregisters. Though, I think we want a tagged metric now that you mention it, not a log statement.

PR #1226, opened 2018-05-11
feat: switch from snaek to milksnake

### LINKED ISSUE #1045: Migrate from snaek to milksnake
snaek was recently deprecated for milksnake. It continues to function for us in its current form but it's unlikely to be maintained going forward. At some point we should migrate to milksnake -- it is a lower level solution than snaek though, so it appears to it'll require some tweaking.

Some more information on milksnake: https://www.reddit.com/r/rust/comments/744mjw/milksnake_a_new_approach_for_rust_and_python/

pjenvey:
requires a couple workarounds:

- rename queue::Sender -> AutopushSender to avoid cbindgen bug w/ type aliases:
  https://github.com/eqrion/cbindgen/issues/131
- make structs not representable in C opaque (remove #[repr(C)])
- have to pin to an older serde_derive for now because cbindgen does

Closes #1045

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,14 @@ language: python
 cache:
   directories:
     - $HOME/.cargo
-    - cargo
     - autopush_rs/target
     - $HOME/.cache/pip
 sudo: required
 dist: trusty
+
+env:
+  global:
+    - WITH_RUST=release
 
 matrix:
   include:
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -2,12 +2,15 @@ language: python
...
     - $HOME/.cargo
-    - cargo
     - autopush_rs/target
     - $HOME/.cache/pip
 sudo: required
 dist: trusty
 
+env:
+  global:
+    - WITH_RUST=release
```
pjenvey: I was hoping to switch travis to rustc debug mode but that enables slog's verbose logging that floods travis w/ 4mb of logs, causing tests to fail (we can revisit this later)

PR #1227, opened 2018-05-11
feat: port delete command to Rust

### LINKED ISSUE #1204: Convert delete message command from Python to Rust

bbangert:
Closes #1204

### DIFF
--- a/autopush_rs/src/protocol.rs
+++ b/autopush_rs/src/protocol.rs
@@ -10,6 +10,8 @@
 use std::collections::HashMap;
 use uuid::Uuid;
 
+use util::ms_since_epoch;
+
 // Used for the server to flag a webpush client to deliver a Notification or Check storage
 pub enum ServerNotification {
     CheckStorage,
@@ -103,11 +105,12 @@ pub enum ServerMessage {
 
 #[derive(Serialize, Default, Deserialize, Clone, Debug)]
 pub struct Notification {
+    #[serde(skip_serializing)]
     pub uaid: Option<String>,
     #[serde(rename = "channelID")]
     pub channel_id: Uuid,
     pub version: String,
-    #[serde(default = "default_ttl")]
+    #[serde(default = "default_ttl", skip_serializing)]
     pub ttl: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub topic: Option<String>,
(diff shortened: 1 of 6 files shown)


autopush_rs/src/protocol.rs
```diff
@@ -103,6 +105,7 @@ pub enum ServerMessage {
 
 #[derive(Serialize, Default, Deserialize, Clone, Debug)]
 pub struct Notification {
+    #[serde(skip_serializing)]
```
pjenvey: doesn't look like we send ttl either

autopush_rs/src/protocol.rs
```diff
@@ -120,6 +123,25 @@ pub struct Notification {
     pub headers: Option<HashMap<String, String>>,
 }
 
+impl Notification {
+    /// Generate a sort key suitable for use as the chidmessageid
+    pub fn sort_key(&self) -> String {
+        if let Some(ref topic) = self.topic {
+            format!("01:{}:{}", self.channel_id.hyphenated(), topic)
+        } else if let Some(ref sortkey_timestamp) = self.sortkey_timestamp {
```
pjenvey: u64 is Copy so this could be a lot simpler. probably like

```rust
if let Some(sortkey_timestamp) = ... {
    format!(
        "..",
        if sortkey_timestamp == 0 { ms_since_epoch } else { sortkey_timestamp },
        ...
    )
}
```

autopush_rs/src/protocol.rs
```diff
@@ -120,6 +123,25 @@ pub struct Notification {
     pub headers: Option<HashMap<String, String>>,
 }
 
+impl Notification {
+    /// Generate a sort key suitable for use as the chidmessageid
```
pjenvey: the big docstring in the python version was nice, let's bring it over

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -905,6 +908,43 @@ impl DynamoStorage {
         Box::new(response)
     }
 
+    /// Delete a given notification from the database
+    ///
+    /// No checks are done to see that this message came from the database or has
+    /// sufficient properties for a delete as that is expectecd to have been done
```
pjenvey: expectecd

autopush_rs/src/protocol.rs
```diff
@@ -120,6 +123,25 @@ pub struct Notification {
...
+    pub fn sort_key(&self) -> String {
+        if let Some(ref topic) = self.topic {
+            format!("01:{}:{}", self.channel_id.hyphenated(), topic)
+        } else if let Some(ref sortkey_timestamp) = self.sortkey_timestamp {
+            let sortkey_timestamp = if *sortkey_timestamp == 0 {
+                ms_since_epoch()
+            } else {
+                sortkey_timestamp.clone()
+            };
+            format!("02:{}:{}", sortkey_timestamp, self.channel_id.simple())
```
pjenvey: self.channel_id.hyphenated() :rage4:

pjenvey: hyphenate it once in the first line of the fn

autopush_rs/src/protocol.rs
```diff
@@ -120,6 +123,33 @@ pub struct Notification {
...
+    /// For topic messages:
+    ///     01:{chid}:{topic}
+    ///
+    /// Old format for non-topic messages that is no longer returned:
+    ///     {chid}:{message_id}
+    pub fn sort_key(&self) -> String {
+        let chid = self.channel_id.hyphenated();
+        if let Some(ref topic) = self.topic {
+            format!("01:{}:{}", chid, topic)
+        } else if let Some(ref sortkey_timestamp) = self.sortkey_timestamp {
```
pjenvey: you don't even need the ref or derefing, you'll get a copy

bbangert: I only put that there cause Rust complained:
```
139 |         if let Some(topic) = self.topic {
    |                     -----    ^^^^ cannot move out of borrowed content
    |                     |
    |                     hint: to prevent move, use `ref topic` or `ref mut topic`
```

PR #1229, opened 2018-05-11
fix: re-rustfmt it all and fix the register expiry

bbangert:


autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -654,7 +651,7 @@ impl DynamoStorage {
                 uaid: s => uaid.simple().to_string(),
                 chidmessageid: s => " ".to_string()
             },
-            update_expression: Some("ADD chids :channel_id, expiry :expiry".to_string()),
+            update_expression: Some("ADD chids :channel_id SET expiry=:expiry".to_string()),
```
bbangert: This bit is important.

PR #1233, opened 2018-05-14
fix: properly set record expiration

jrconlin:
Closes bug/1231


autopush/db.py
```diff
@@ -537,25 +537,42 @@ def __init__(self, tablename, boto_resource=None,
...
                 'chidmessageid': ' ',
             },
-            UpdateExpression='ADD chids :channel_id, expiry :expiry',
-            ExpressionAttributeValues=expr_values,
+            UpdateExpression='ADD chids :channel_id',
+            ExpressionAttributeValues={
+                ":channel_id": set([normalize_id(channel_id)]),
+            },
         )
+        self.touch_record(uaid)
```
bbangert: The chids and expiry can be changed in the same command, the Rust version does this, the update expression looks like this:
```rust
            update_expression: Some("ADD chids :channel_id SET expiry=:expiry".to_string()),
```

autopush/websocket.py
```diff
@@ -989,6 +989,7 @@ def finish_webpush_notifications(self, result):
         now = int(time.time())
         messages_sent = False
         message = self.db.message_table(self.ps.message_month)
+        message.touch_record(self.ps.uaid)
```
pjenvey: Is this the best place for this? process_notifications/finish_webpush_notifications isn't called unoften, right?

Also calling it from here, it should really be called via a thread

jrconlin: I suppose it makes sense if we ever expect that a websocket connection will live longer than the max expiration and only does checks. I'll grant that it's an edge case.

autopush/db.py
```diff
@@ -537,42 +537,59 @@ def __init__(self, tablename, boto_resource=None,
     def table_status(self):
         return self.table.table_status
 
+    @track_provisioned
+    def touch_record(self, uaid, ttl=None):
```
pjenvey: should we kill this, then? it's now unused

jrconlin: yep

PR #1234, opened 2018-05-15
feat: port store messages command to Rust

### LINKED ISSUE #1208: Convert store message command from Python to Rust

bbangert:
Add's Rust integration test to verify the direct message is stored that
was not present before.

Closes #1208

### DIFF
--- a/autopush_rs/src/util/ddb_helpers.rs
+++ b/autopush_rs/src/util/ddb_helpers.rs
@@ -1,3 +1,4 @@
+use std::cmp::min;
 /// DynamoDB Client helpers
 use std::collections::{HashMap, HashSet};
 use std::env;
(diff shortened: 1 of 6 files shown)


autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -182,6 +184,17 @@ impl From<NotificationHeaders> for HashMap<String, String> {
     }
 }
 
+impl From<HashMap<String, String>> for NotificationHeaders {
+    fn from(val: HashMap<String, String>) -> NotificationHeaders {
+        NotificationHeaders {
+            crypto_key: val.get("crypto_key").map(|v| v.to_string()),
+            encryption: val.get("encrypption").map(|v| v.to_string()),
```
jrconlin: "encryption"

jrconlin: Also, not sure what the impact of setting these to the string "None" might be. Might be better to set them to an empty string?

bbangert: They won't be included if they're None, as the serde tag says to skip serializing them in that case.

autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -899,6 +930,48 @@ impl DynamoStorage {
...
+                DynamoDbNotification::from_notif(n)
+                    .map(|notif| serde_dynamodb::to_hashmap(&notif).ok())
+                    .unwrap_or_default()
+            })
+            .map(|hm| WriteRequest {
+                put_request: Some(PutRequest { item: hm }),
+                delete_request: None,
+            })
+            .collect();
+        let mut request_items = HashMap::new();
```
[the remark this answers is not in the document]
bbangert: I can? hashmap! makes a different structure than I want here though.

PR #1237, opened 2018-05-15
feat: port migrate user to Rust, remove Python calling

### LINKED ISSUE #1206: Convert migrate user command from Python to Rust

bbangert:
Closes #1206


autopush_rs/src/util/ddb_helpers.rs
```diff
@@ -929,6 +998,44 @@ impl DynamoStorage {
...
+        router_table_name: &str,
+    ) -> MyFuture<String> {
+        let ddb = self.ddb.clone();
+        let ddb1 = self.ddb.clone();
+        let ddb2 = self.ddb.clone();
+        let uaid = uaid.clone();
+        let cur_month = current_message_month.to_string();
+        let cur_month1 = cur_month.clone();
+        let cur_month2 = cur_month.clone();
+        let cur_month3 = cur_month.clone();
```
bbangert: Yes I hate all these clones, the joy of static closures.

autopush_rs/src/client.rs
```diff
@@ -690,9 +689,12 @@ where
...
             transition!(CheckStorage { data });
         } else if all_acked && webpush.flags.rotate_message_table {
-            let response = data.srv.migrate_user(
-                webpush.uaid.simple().to_string(),
-                webpush.message_month.clone(),
+            debug!("Triggering migration");
+            let response = data.srv.ddb.migrate_user(
+                &webpush.uaid,
+                &webpush.message_month,
+                &data.srv.opts.current_message_month,
```
pjenvey: so opts.current_message_month is essentially hardcoded at startup in autopush_rs.

ddb.migrate_user doesn't need to return it, then. await_migrate_user can assign `webpush.message_month = data.srv.opts.current_message_month.clone();` instead

autopush_rs/src/server/mod.rs
```diff
@@ -223,6 +223,10 @@ pub extern "C" fn autopush_server_new(
                 .expect("poll interval cannot be 0"),
         };
         opts.message_table_names.sort_unstable();
+        opts.current_message_month = opts.message_table_names
```
pjenvey: a number of the db methods do this same call to get current_message_month. we can improve that later though since they're getting refactored

autopush_rs/src/client.rs
bbangert: That's true.

autopush_rs/src/server/mod.rs
bbangert: Yes, and they do weird things since its not 'for-sure' that there's a last() on the vec, while now its guaranteed there'll be a current month.

PR #1244, opened 2018-05-18
feat: add Rust + Python mixed integration tests

### LINKED ISSUE #1221: Create integrated mixed tests
For better integrated verification the Rust node is compatible, we should have an integrated test that spins up both the Rust and Python connection nodes on different ports and try:

- Connect/reg to Python node, disconnect, send messages in, connect to Rust and verify they come down
- Connect/reg to Rust node, disconnect, etc.
- etc.

bbangert:
Closes #1221

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -100,8 +100,11 @@ keyid="http://example.org/bob/keys/123;salt="XZwpw6o37R-6qoZjw6KwAw"\
         else:
             return f
 
-    def connect(self):
-        self.ws = websocket.create_connection(self.url)
+    def connect(self, connection_port=None):
+        url = self.url
+        if connection_port:
+            url = "ws://localhost:{}/".format(connection_port)
+        self.ws = websocket.create_connection(url)
         return self.ws.connected
 
     def hello(self, uaid=None, services=None):
--- a/autopush/tests/test_rs_integration.py
+++ b/autopush/tests/test_rs_integration.py
@@ -38,7 +38,11 @@ from autopush.db import (
     Message,
 )
 from autopush.logging import begin_or_register
-from autopush.main import EndpointApplication, RustConnectionApplication
+from autopush.main import (
+    ConnectionApplication,
+    EndpointApplication,
+    RustConnectionApplication,
+)
 from autopush.utils import base64url_encode
 from autopush.tests.support import TestingLogObserver
 from autopush.tests.test_integration import (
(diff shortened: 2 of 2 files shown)


autopush/tests/test_rs_integration.py
```diff
@@ -740,130 +709,278 @@ def test_delete_saved_notification(self):
...
-                self.conn.db.router.get_uaid,
-                client.uaid)
-            if c["current_month"] == self.conn.db.current_msg_month:
-                break
-            else:
-                yield deferToThread(time.sleep, 0.2)
+class TestRustWebPushBroadcast(unittest.TestCase):
+    connection_port = 9052
+    endpoint_port = 9060
+    router_port = 9072
```
jrconlin: nit: Base these as offsets from the globals? (looks like endpoint_port stays the same anyway)

autopush/tests/test_rs_integration.py
```diff
@@ -950,102 +1086,159 @@ def test_webpush_monthly_rotation_prior_record_exists(self):
...
         yield self.shut_down(client)
 
 
-class TestRustWebPushBroadcast(unittest.TestCase):
-    connection_port = 9050
+class TestRustAndPythonWebPush(unittest.TestCase):
+    connection_port = 9052
     endpoint_port = 9060
-    router_port = 9070
+    router_port = 9072
```
jrconlin: nit: same as above, offset these?

bbangert: Either way one needs to look at the ports at the top to determine if any collide, and the tests won't pass if they collide since the server will fail to start. Is there some reason for using offsets I'm missing?

autopush/tests/test_integration.py
```diff
@@ -100,8 +100,11 @@ def __getattribute__(self, name):
         else:
             return f
 
-    def connect(self):
-        self.ws = websocket.create_connection(self.url)
+    def connect(self, connection_port=None):
+        url = self.url
+        if connection_port:
+            url = "ws://localhost:{}/".format(connection_port)
```
pjenvey: This is pretty meh, how about
```python
def reconnect(url):
    if self.ws.connected:
        self.disconnect()
    self.url = url
    return self.connect()
```

easier than disconnect(); connect(); and also works when you need to do something in between those 2

autopush/tests/test_rs_integration.py
```diff
@@ -53,6 +57,13 @@
 ROUTER_TABLE = os.environ.get("ROUTER_TABLE", "router_int_test")
 MESSAGE_TABLE = os.environ.get("MESSAGE_TABLE", "message_int_test")
 
+crypto_key = Fernet.generate_key()
```
pjenvey: upper case these constants

autopush/tests/test_integration.py
bbangert: That doesn't check that the current connection is to the given URL, not sure how easy it is to determine that.

pjenvey: where do you need that?

bbangert: It's also not a reconnect if its connecting to a different URL than what it was on, that's a new connect. But now its just semantics,sigh.

PR #1245, opened 2018-05-18
refactor: minor cleanup, then impl trait and box killing

pjenvey:
these are 2 separate commits

### DIFF
--- a/autopush_rs/src/db/mod.rs
+++ b/autopush_rs/src/db/mod.rs
@@ -103,13 +103,12 @@ impl DynamoStorage {
             table_name: table_name.to_string(),
             ..Default::default()
         };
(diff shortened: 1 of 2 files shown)


autopush_rs/src/db/mod.rs
```diff
@@ -223,8 +219,8 @@ impl DynamoStorage {
     ) -> MyFuture<bool> {
```
bbangert: Not able to impl trait here?

pjenvey: I'll merge this then try this out, moving more of the boxing into client.rs (i think it's mostly going to be moving around vs getting rid of boxes)

Issue #1232, opened 2018-05-14
Reduce our docker build size


pjenvey:
Our Docker image size has increased from 725mb to 1gb between 1.45.0 and 1.46.0. This is just too large, it takes circleci =~ 7 minutes to upload

We can easily reduce some of this size by not including the full rustup install/cargo/build artifacts in the final image.

PR #1248, opened 2018-05-22
fix: shrink the Docker build

pjenvey:
by nuking the rustup install & target dir after installation

Issue #1232


autopush/tests/test_rs_integration.py
```diff
@@ -584,7 +584,7 @@ def test_ttl_batch_expired_and_good_one(self):
             yield client.send_notification(data=data, ttl=1)
 
         yield client.send_notification(data=data2)
-        time.sleep(0.75)
+        time.sleep(1)
```
jrconlin: Heh, c'mon, luck of the clock not enough? ;)

Issue #1249, opened 2018-05-22
ensure strongly consistent reads for autopush_rs message fetching


pjenvey:
We missed enabling consistent_read on the two fetch_message functions while porting from Python

PR #1250, opened 2018-05-22
fix: ensure consistent_read on fetch messages fns

pjenvey:
lost this in the python port

Issue #1249


autopush_rs/src/db/commands.rs
```diff
@@ -125,10 +127,6 @@ pub fn fetch_timestamp_messages(
                     .filter_map(|ddb_notif: DynamoDbNotification| ddb_notif.into_notif().ok())
                     .collect()
             });
-            if messages.is_empty() {
-                return Ok(Default::default());
-            }
-
             let timestamp = messages.iter().filter_map(|m| m.sortkey_timestamp).max();
```
pjenvey: max() gives us None for free here on empty iters

PR #1252, opened 2018-05-23
chore: tag 1.47.1

pjenvey:


autopush/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '1.46.0'  # pragma: nocover
+__version__ = '1.47.1'  # pragma: nocover
```
bbangert: Does this mean 1.47.0 will not identify itself correctly?

PR #1255, opened 2018-05-23
feat: switch Rust startup to Rust only

### LINKED ISSUE #1243: Load config in Rust
The config crate allows TOML/env reading so that we should be able to load config in Rust directly and avoid Python for starting up the connection node.

bbangert:
Co-authored by: Phil Jenvey <pjenvey@underboss.org>

Closes #1243


alexcrichton:
🎉

autopush_rs/src/settings.rs
```diff
@@ -0,0 +1,130 @@
...
+            self.router_port
+        )
+    }
+
+    pub fn endpoint_url(&self) -> String {
+        let hostname = self.host_name();
+        format!(
+            "{}://{}:{}",
+            self.endpoint_scheme,
+            self.endpoint_hostname.as_ref().unwrap_or(&hostname),
```
pjenvey: for just the connection node it doesn't make sense to fallback to hostname

Dockerfile.python27
```diff
@@ -14,7 +14,7 @@ RUN \
     make clean && \
     pip install -r requirements.txt && \
     python setup.py develop && \
-    cd autopush_rs && \
+    cargo install && \
```
pjenvey: needs --path

autopush_rs/src/settings.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub router_ssl_cert: Option<String>,
+    pub router_ssl_dh_param: Option<String>,
+    pub auto_ping_interval: f64,
+    pub auto_ping_timeout: f64,
+    pub max_connections: u32,
+    pub close_handshake_timeout: u32,
+    pub endpoint_scheme: String,
+    pub endpoint_hostname: Option<String>,
+    pub endpoint_port: u16,
+    pub crypto_key: String,
```
pjenvey: this is an append (List[str]) in main_argparse. Can Config do something similar?

bbangert: Not if we want it to work with env vars.

autopush_rs/src/settings.rs
bbangert: It's used to generate the endpoint, it needs some value from somewhere.

pjenvey: Can we just start expecting it here then? The python's handling of hostnames is pretty confusing

bbangert: The tests assume the endpoint hostname will be set to the hostname. They'd all have to change too.

PR #1256, opened 2018-05-29
chore: upgrade Rust dependencies

bbangert:


autopush_rs/Cargo.toml
```diff
@@ -15,62 +15,62 @@ name = "autopush"
...
 tokio-io = "0.1.6"
 tokio-openssl = "0.2.0"
 tokio-service = "0.1.0"
 tokio-tungstenite = { version = "0.5.1", default-features = false }
 tungstenite = { version = "0.5.3", default-features = false }
-uuid = { version = "0.6.2", features = ["serde", "v4"] }
+uuid = { version = "0.6.5", features = ["serde", "v4"] }
 woothee = "0.7.3"
+
+[dependencies.config]
```
pjenvey: FYI this could also all go inline: config = { git = "..", rev = ".." }

jrconlin: Heh, FWIW, I'm a fan of keeping the information grouped and not inlined because it makes it clear there's special handling.

(Granted, me advocating for a style is probably a really clear sign you shouldn't do it, but that's beside the point.)

PR #1261, opened 2018-06-19
bug: doc fixup

### LINKED ISSUE #1260: Update running.rst documentation
The documentation for [running.rst](https://github.com/mozilla-services/autopush/blob/master/docs/running.rst) is old and difficult to understand (personal consideration).

1. From commit [6865ded](https://github.com/mozilla-services/autopush/commit/6865dede42b886073a927d7c88c74328b7399670#diff-cb5bbedac0c1db2213e4cd6f94ea9408) the boto-compose.cfg is no more available, specify in the docs how to handle with this change.
2. The link provided in Sample Configurations is wrong.

Currently i'm trying to run on Windows10 Enterprise 64bit. I had also the following problems:

1. `$ docker run -t -i ~/autopush autokey` didn't work, i had to change it to `$ docker run -t -i bbangert/autopush autokey`
2. Start Autopush command `$ CRYPTO_KEY=THEKEY docker-compose up` didn't work i had to change it to `$env:CRYPTO_KEY="THEKEY";docker-compose up`

jrconlin:
* Fix out of date documentation
* Add AWS_LOCAL_CREDS to pass local credential info (thanks boto3)
* Fix tests to use new local env var (like we do for endpoint)
* Add rust migration document
* Fix autokey to return something useful
* Point out that docker_compose does bad things to your expectations
  about CRYPTO_KEY

Closes #1260
Issue mozilla-services/autopush-rs#7


autopush/db.py
```diff
@@ -468,6 +468,10 @@ def __init__(self, **kwargs):
         conf = kwargs
         if not conf.get("endpoint_url"):
             conf["endpoint_url"] = os.getenv("AWS_LOCAL_DYNAMODB")
+        creds = os.getenv("AWS_LOCAL_CREDS")
```
pjenvey: autopush-rs hardcodes creds to bogus whenever AWS_LOCAL_DYNAMODB is set. I'm thinking we do the same here (killing AWS_LOCAL_CREDS), as I don't think anyone will ever need custom creds when pointing to a local ddb

jrconlin: I was debating that myself. The only potential reason I could come up with to keep them separate was the low percentile chance that they actually had some specific credentials that they wanted to use. 

Yeah, I have really dumb arguments with myself.

docker-compose.yml
```diff
@@ -1,12 +1,14 @@
...
   autopush:
-    image: bbangert/autopush
+    # image: bbangert/autopush
+    build: .
     hostname: autopush
     environment:
       - LOCAL_HOSTNAME=localhost
       - ROUTER_HOSTNAME=autopush
       - AWS_LOCAL_DYNAMODB=http://dynamodb:8000
+      - AWS_LOCAL_CREDS=Bogus:Bogus
```
pjenvey: one more thing, forgot to kill it here and below (possibly in the commit message too?)

docker-compose.yml
```diff
@@ -1,7 +1,8 @@
 version: '2'
 services:
   autopush:
-    image: bbangert/autopush
+    # image: bbangert/autopush
```
bbangert: Just delete this?

docs/index.rst
```diff
@@ -10,6 +10,11 @@ autopush
 
 Mozilla Push server and Push Endpoint utilizing PyPy, twisted, rust, and DynamoDB.
```
bbangert: Strike rust.

docker-compose.yml
jrconlin: I was actually debating if we wanted to make this use the image we generate when we tag  a release, rather than doing a local build.

docker-compose.yml
```diff
@@ -1,7 +1,12 @@
 version: '2'
 services:
   autopush:
-    image: bbangert/autopush
+    build: .
+    # If you do not wish to use a local build, comment out
+    # the previous 'build:' line and uncomment the following
+    # 'image:' line. This will use the latest production
+    # built image.
+    #image: mozilla/autopush
```
jrconlin: I'll note that the current mozilla/autopush image doesn't work with this version of docker-compose, thus the comment for now. Obviously, the most stable thing is to build locally, but if we want folks to just play, we can switch the defaults here.

PR #1282, opened 2018-08-07
feat: improve endpoint_diagnostic tool and documents

### LINKED ISSUE #1281: Improve diagnostic cli, add testing initialization suggestions
We don't often need the `endpoint_diagnostic` command, but when we do, we almost always forget we have it. Add some docs about how to use it as well as clean up some of the messaging and output.

jrconlin:
Closes #1281

### DIFF
--- a/autopush/diagnostic_cli.py
+++ b/autopush/diagnostic_cli.py
@@ -5,11 +5,32 @@ import re
 
 import configargparse
 from twisted.logger import Logger
+from cryptography.fernet import InvalidToken
 
 from autopush.config import AutopushConfig
 from autopush.db import DatabaseManager, Message
+from autopush.exceptions import ItemNotFound, InvalidTokenException
 from autopush.main import AutopushMultiService
 from autopush.main_argparse import add_shared_args
+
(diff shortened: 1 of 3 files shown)


autopush/diagnostic_cli.py
```diff
@@ -54,26 +75,35 @@ def run(self):
...
-        uaid, chid = parsed["uaid"], parsed["chid"]
+        try:
+            parsed = self._conf.parse_endpoint(
+                self.db.metrics,
+                token=token,
+                version=api_ver,
+            )
+            uaid, chid = parsed["uaid"], parsed["chid"]
+        except (InvalidTokenException, InvalidToken) as ex:
+            print("Token could not be deciphered. Are you using the correct"
```
pjenvey: is it worth including the exception in this message (I'm leaning towards sure, why not)?

jrconlin: Yeah, was kinda on the fence. Mostly because the bulk was "Invalid Token", but yeah, there are times it's useful to dump as well.

PR #1289, opened 2018-08-27
feat: upgrade to pypy6 for debug symbols

### LINKED ISSUE #1288: Upgrade to pypy 6
We recently had a need for debug symbols of the pypy binary. The 6.0.0 release began including them: let's upgrade now in case we need them again

pjenvey:
and include its new gc.get_stats in memusage

Closes #1288

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -139,7 +139,7 @@ class GCMRouter(object):
                                        self._base_tags,
                                        reason="timeout"))
             raise RouterException("Server error", status_code=502,
-                                  errno=902,
+                                  errno=903,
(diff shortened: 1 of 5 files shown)


autopush/router/gcm.py
```diff
@@ -139,7 +139,7 @@ def _route(self, notification, uaid_data):
                                        self._base_tags,
                                        reason="timeout"))
             raise RouterException("Server error", status_code=502,
-                                  errno=902,
+                                  errno=903,
```
jrconlin: Missing from https://autopush.readthedocs.io/en/latest/http.html

PR #1290, opened 2018-08-27
feat: add ADM support

### LINKED ISSUE #1275: Add Amazon Device Messaging support
For Fire TV/OS support

Part of [OKR 1.3](https://docs.google.com/spreadsheets/d/1Bt71RclMcgAI8ZSV7GDYMhLVAahqsWnJfTjuZ9jNFVY/edit#gid=2050679495)

See also: https://github.com/jrconlin/ADMPushDemo

Needs:
 - [x] client 
 - [x] Tests
 - [x] Docs

jrconlin:
Closes #1275

### DIFF
--- a/autopush/router/adm.py
+++ b/autopush/router/adm.py
@@ -1,0 +1,220 @@
+"""ADM Router"""
+import time
+import requests
+
+from typing import Any  # noqa
+
+from requests.exceptions import ConnectionError, Timeout
+from twisted.internet.threads import deferToThread
+from twisted.logger import Logger
+
+from autopush.exceptions import RouterException
+from autopush.metrics import make_tags
+from autopush.router.interface import RouterResponse
+from autopush.types import JSONDict  # noqa
+
+
+class ADMAuthError(Exception):
+    pass
+
+
+class ADMClient(object):
+    def __init__(self,
+                 credentials=None,
+                 logger=None,
+                 metrics=None,
+                 endpoint="api.amazon.com",
+                 timeout=2,
+                 **options
+                 ):
+
+        self._client_id = credentials["client_id"]
+        self._client_secret = credentials["client_secret"]
+        self._token_exp = 0
+        self._auth_token = None
+        self._aws_host = endpoint
+        self._logger = logger
+        self._metrics = metrics
+        self._request = requests
+        self._timeout = timeout
+
+    def refresh_key(self):
+        url = "https://{}/auth/O2/token".format(self._aws_host)
+        if self._auth_token is None or self._token_exp < time.time():
+            body = dict(
+                grant_type="client_credentials",
+                scope="messaging:push",
+                client_id=self._client_id,
+                client_secret=self._client_secret
+            )
+            headers = {
+                "content-type": "application/x-www-form-urlencoded"
+            }
+            resp = self._request.post(url, data=body, headers=headers,
+                                      timeout=self._timeout)
+            if resp.status_code != 200:
+                self._logger.error("Could not get ADM Auth token {}".format(
+                    resp.text
+                ))
+                raise ADMAuthError("Could not fetch auth token")
+            reply = resp.json()
+            self._auth_token = reply['access_token']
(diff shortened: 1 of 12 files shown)


autopush/tests/test_integration.py
```diff
@@ -1936,6 +1939,253 @@ def test_registration(self):
...
+        self._mock_reply.json.return_value = dict(access_token="access.123")
+
+        response, body = yield _agent(
+            'POST',
+            str(jbody['endpoint']),
+            headers=Headers({
+                "crypto-key": [crypto_key],
+                "encryption": [salt],
+                "ttl": ["0"],
+                ""
```
pjenvey: kill this stray line (it just str concats into "content-encoding")

autopush/router/adm.py
```diff
@@ -0,0 +1,206 @@
...
+            body = dict(
+                grant_type="client_credentials",
+                scope="messaging:push",
+                client_id=self._client_id,
+                client_secret=self._client_secret
+            )
+            headers = {
+                "content-type": "application/x-www-form-urlencoded"
+            }
+            resp = self._request.post(url, data=body, headers=headers)
```
pjenvey: we definitely want to specify some kind of timeout to requests calls, GCM bridge uses timeout=10. Then you have to catch the Timeout exception somewhere (probably _route)

autopush/router/adm.py
```diff
@@ -0,0 +1,206 @@
...
+        }
+        data = {}
+        if ttl:
+            data["expiresAfter"] = ttl
+        if collapseKey:
+            data["consolidationKey"] = collapseKey
+        data["data"] = payload
+        url = ("https://api.amazon.com/messaging/registrations"
+               "/{}/messages".format(reg_id))
+        resp = self._request.post(
```
pjenvey: same timeout here

autopush/tests/test_integration.py
```diff
@@ -1936,6 +1939,253 @@ def test_registration(self):
...
+        self._add_router()
+
+        # prevent the token refresh check
+        self.ep.routers["adm"].profiles["dev"]._auth_token = "SomeToken"
+        self.ep.routers["adm"].profiles["dev"]._token_exp = time.time() + 300
+        self._mock_reply.status_code = 400
+        self._mock_reply.text = "Test error"
+        self._mock_reply.content = "Test content"
+
+        # This will call RouterException, which may halt the test thread.
```
pjenvey: I don't get this comment, why would RouterException halt the test thread?

autopush/router/adm.py
bbangert: We can't use an async HTTP request? I know some of the router calls use third party tooling that we couldn't use the twisted http client with, but we should be able to here, right?

autopush/router/adm.py
jrconlin: agreed.

autopush/router/adm.py
jrconlin: I'd prefer to do the async requests as a separate issue, since that will span multiple files and fuzz exactly what this patch is supposed to address.

autopush/tests/test_integration.py
jrconlin: Heh, had that in there for readability but then realized that i was the only person that cared.

autopush/tests/test_integration.py
jrconlin: Since we'd be calling via the `_agent`, `RouterException` is passed up and not trapped, causing `_agent` to fail out and kill/fail the test. I was not able to trap for it. Thus calling the send directly with the raise trap. I can try to muck about more with `_agent` to try and handle that sort of thing, but seems a bit out of scope for this patch.

pjenvey: That might be a case of twisted's TestCase failing if it sees a log.failure call made during the test. You can call self.flushLoggedErrors() at the end of tests to stop it from doing that

jrconlin: @pjenvey Ah, cool. I'll try that.

PR #1294, opened 2018-09-25
feat: add dockerflow callbacks for endpoint

### LINKED ISSUE #1293: Add __version__ to endpoint
Right now we can only see /__version__ on the rust connection instances, it would be nice for QA and OPs to be able to see the currently deployed python version as well.

jrconlin:
Closes #1293

### DIFF
--- a/autopush/web/dockerflow.py
+++ b/autopush/web/dockerflow.py
@@ -1,0 +1,48 @@
+"""DockerFlow endpoints"""
+
+import cyclone.web
+
+from twisted.internet.threads import deferToThread
+from autopush.web.base import BaseWebHandler
+
+
+class VersionHandler(BaseWebHandler):
+
+    def _get_version(self):
+        self.set_header("Content-Type", "application/json")
+        try:
+            with open("version.json") as vers:
+                self.write(vers.read())
+        except IOError as ex:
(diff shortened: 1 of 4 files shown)


autopush/web/dockerflow.py
```diff
@@ -0,0 +1,42 @@
...
+        d.addBoth(self.finish)
+        d.addErrback(self._error)
+        return d
+
+
+class LBHeartbeatHandler(BaseWebHandler):
+
+    @cyclone.web.asynchronous
+    def get(self):
+        self.set_status(200)
```
pjenvey: nit: default's 200 so this line can be killed

autopush/web/dockerflow.py
```diff
@@ -0,0 +1,42 @@
+"""DockerFlow endpoints"""
+
+import cyclone.web
+
+from twisted.internet.threads import deferToThread
+from autopush.web.base import BaseWebHandler
+
+
+class VersionHandler(BaseWebHandler):
```
pjenvey: let's have VersionHandler call self.set_header("Content-Type", "application/json")

Also let's have both new handlers override authenticate_peer_cert to do nothing (skipping auth checks) like the handlers in web.health/memusage

PR #1296, opened 2018-10-01
feat: make gcm calls use async callbacks

### LINKED ISSUE #1291: Upgrade GCM/ADM bridges to use twisted async http
Older code is using the old thread post model. We should standardize from `requests.post` where possible.

jrconlin:
Closes #1291

### DIFF
--- a/autopush/router/gcm.py
+++ b/autopush/router/gcm.py
@@ -1,9 +1,8 @@
 """GCM Router"""
 from typing import Any  # noqa
 
-from requests.exceptions import ConnectionError, Timeout
-from twisted.internet.threads import deferToThread
 from twisted.logger import Logger
+from twisted.internet.error import ConnectError, TimeoutError
 
 from autopush.exceptions import RouterException
 from autopush.metrics import make_tags
@@ -28,7 +27,7 @@ class GCMRouter(object):
         self.dryRun = router_conf.get("dryrun", False)
         self.collapseKey = router_conf.get("collapseKey")
         timeout = router_conf.get("timeout", 10)
-        self.gcm = {}
+        self.gcmclients = {}
         self.senderIDs = {}
         # Flatten the SenderID list from human readable and init gcmclient
         if not router_conf.get("senderIDs"):
@@ -36,7 +35,8 @@ class GCMRouter(object):
         for sid in router_conf.get("senderIDs"):
             auth = router_conf.get("senderIDs").get(sid).get("auth")
             self.senderIDs[sid] = auth
-            self.gcm[sid] = gcmclient.GCM(auth, timeout=timeout)
+            self.gcmclients[sid] = gcmclient.GCM(auth, timeout=timeout,
+                                                 logger=self.log)
         self._base_tags = ["platform:gcm"]
         self.log.debug("Starting GCM router...")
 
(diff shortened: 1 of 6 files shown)


jrconlin:
`inlineCallbacks`, or at least a better implementation may improve things, however when I first tried them the threads were returning too soon, where stacking the deferreds were not. I fully expect that months from now, when I understand this better, I'll swear about how horrible it all is and rewrite it. 

For now, it should be good enough to validate if deferred processing will reduce CPU lock, and I can implement a better version for the pending FCM changeover.

autopush/router/gcmclient.py
```diff
@@ -30,14 +33,12 @@ def __init__(self, message, response):
...
+            response.headers.getRawHeaders('Retry-After') or [None])[0]
 
-        if response.status_code != 200:
-            self.retry_message = message
-        else:
-            self._parse_response(message, response.content)
-
-    def _parse_response(self, message, content):
+    def parse_response(self, content, code, message):
+        if code in (400, 401, 404):
```
pjenvey: I think the 401 case should be removed from here just to help clarify as it would have already been handled in process()

autopush/router/gcm.py
```diff
@@ -111,53 +111,61 @@ def _route(self, notification, uaid_data):
...
+        return failure
 
     def _error(self, err, status, **kwargs):
         """Error handler that raises the RouterException"""
         self.log.debug(err, **kwargs)
         return RouterException(err, status_code=status, response_body=err,
                                **kwargs)
 
-    def _process_reply(self, reply, uaid_data, ttl, notification):
+    def _process_reply(self, client, uaid_data, ttl, notification):
```
pjenvey: why do we get back a client now? (looks like all we need is its response, didn't see it utilized in tests but maybe I missed it)

autopush/router/gcmclient.py
```diff
@@ -124,9 +126,36 @@ def __init__(self,
         self._endpoint = "https://{}".format(endpoint)
         self._api_key = api_key
         self.metrics = metrics
-        self.log = logger
+        self.log = logger or Logger()
         self._options = options
-        self._sender = requests.post
+        self._sender = treq.request
```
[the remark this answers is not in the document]
jrconlin: Heh, I wanted to be explicit rather than rely on a wrapper. Not worth fighting over.

autopush/router/gcm.py
jrconlin: You're correct, we only really need `client.response`. It just felt odd to me that we were making a request of GCM and returning something else. Not a big deal, reverting.

autopush/web/webpush.py
```diff
@@ -527,6 +527,9 @@ def _router_completed(self, response, uaid_data, warning="",
                           router_type=None, vapid=None):
         """Called after router has completed successfully"""
         # Log the time taken for routing
+        print("router_complete response: ", response)
```
pjenvey: looks good except is this leftover debugging?

autopush/web/webpush.py
```diff
@@ -527,6 +527,8 @@ def _router_completed(self, response, uaid_data, warning="",
                           router_type=None, vapid=None):
         """Called after router has completed successfully"""
         # Log the time taken for routing
+        if response is None:
```
pjenvey: is this needed as well? I figured it was also stray debugging, seems like route_notification still always returns something

jrconlin: I had originally included it because the router is technically a module that may fail. I might revisit this in the future, since you're right that we should still capture metric info off of it, but for now, it's unused code.

PR #1308, opened 2018-12-13
feat: Use FCM HTTPv1 protocol with twisted async

### LINKED ISSUE #1291: Upgrade GCM/ADM bridges to use twisted async http

jrconlin:
With the coming switch to FCM, it makes sense to also switch to the new
FCM HTTPv1 protocol.

Closes #1291

### DIFF
--- a/autopush/router/fcm_v1.py
+++ b/autopush/router/fcm_v1.py
@@ -58,7 +58,6 @@ class FCMv1Router(FCMRouter):
     def route_notification(self, notification, uaid_data):
         """Start the FCM notification routing, returns a deferred"""
         router_data = uaid_data["router_data"]
-        # Kick the entire notification routing off to a thread
         return self._route(notification, router_data)
 
     def _route(self, notification, router_data):


autopush/router/fcmv1client.py
```diff
@@ -0,0 +1,142 @@
...
+            raise FCMAuthenticationError("Authentication Error")
+
+        result = Result(response)
+
+        d = response.text()
+        d.addCallback(result.parse_response)
+        return d
+
+    def error(self, failure):
+        if isinstance(failure.value, FCMAuthenticationError) or \
```
pjenvey: nit
```suggestion
        if isinstance(failure.value, (FCMAuthenticationError, TimeoutError, ConnectError):
```

autopush/router/fcmv1client.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def send(self, token, payload):
+        atoken = self._get_access_token()
+        headers = {
+            'Authorization': 'Bearer {}'.format(atoken.access_token),
+            'Content-Type': 'application/json; UTF-8'
+        }
+        message = self._build_message(token, payload)
+        if 'timeout' not in self._options:
+            self._options['timeout'] = 3
```
pjenvey: gcmclient also defaults to 3 but then gcm.py overwrites it with its own default of 10 (should we set this to 10 somewhere as too?)

autopush/router/fcm_v1.py
```diff
@@ -0,0 +1,168 @@
...
+        # type: (str, JSONDict, str, *Any, **Any) -> None
+        """Validate that the FCM Instance Token is in the ``router_data``"""
+        senderid = app_id
+        # "token" is the FCM token generated by the client.
+        if "token" not in router_data:
+            raise self._error("connect info missing FCM Instance 'token'",
+                              status=401,
+                              uri=kwargs.get('uri'),
+                              senderid=repr(senderid))
+        if not (senderid == self.senderID):
```
pjenvey: nit
```suggestion
         if senderid != self.senderID:
```

autopush/router/fcm_v1.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _error(self, err, status, **kwargs):
+        """Error handler that raises the RouterException"""
+        self.log.debug(err, **kwargs)
+        return RouterException(err, status_code=status, response_body=err,
+                               **kwargs)
+
+    def _process_reply(self, reply, notification, router_data, ttl):
+        """Process FCM send reply"""
+        # acks:
```
pjenvey: this comment's outdated (left over from gcm.py?)

autopush/router/fcmv1client.py
jrconlin: Well, GCM is different, and the value is only set to 10 if it's not already present, so it's more a backup than anything else. That said, I should be better about having a "global base timeout" value.

autopush/router/fcm_v1.py
```diff
@@ -0,0 +1,165 @@
...
+                              senderid=repr(senderid))
+        if senderid != self.senderID:
+            raise self._error("Invalid SenderID", status=410, errno=105)
+        # Assign a senderid
+        router_data["creds"] = {"senderID": self.senderID}
+
+    def route_notification(self, notification, uaid_data):
+        """Start the FCM notification routing, returns a deferred"""
+        router_data = uaid_data["router_data"]
+        # Kick the entire notification routing off to a thread
```
bbangert: Thread?

jrconlin: bad comment, thanks!

PR #1314, opened 2019-01-05
bug: Allow "dh=" parameter for Crypto-Key with 8188 encoding

### LINKED ISSUE #1313: Don't reject Crypto-Key headers with "dh=" component
Crypto-Key should still contain "dh=" for the client.

jrconlin:
The client requires the Crypto-Key to specify the dh value for
Content-Type: "aes128gcm".

Closes #1313


jrconlin:
invalid

PR #1318, opened 2019-02-19
chore: tag 1.53.1

pjenvey:


jrconlin:
not sure why, but flake8 is now complaining about https://github.com/mozilla-services/autopush/blob/master/autopush/tests/test_router.py#L66 being over-indented. 

I'd apologize, but that's code from 2 years ago.

pjenvey:
tox installs the latest flake8 for this check, so a recent release starting catching this one. I'll fix it in another PR

PR #1319, opened 2019-02-19
chore: flake8

pjenvey:
and pin the versions used


tox.ini
```diff
@@ -14,7 +14,7 @@ basepython = pypy
 
 [testenv:flake8]
 commands = flake8 autopush
-deps = flake8
+deps = flake8==3.7.6
```
jrconlin: 👍 
(More consistency, less Zoolander)

PR #1321, opened 2019-03-13
Added Strict-Transport-Security header to 404s

psiinon:


psiinon:
@jrconlin any idea when this will go to production?

jrconlin:
We don't often push changes out, mostly because of limited QA time. If this is more urgent, I can see about doing a limited pull and get this started.

psiinon:
it would be great if this could be pushed out before the end of the month, Q1 goals and all that ;)

jrconlin:
Sadly, just missed the two week push cycle we have, however, since this is
now out of your hands and we've got the bug for pushing this live, you can
note this as done. ;) Not your problem anymore and folks can rightly yell
at me.

https://bugzilla.mozilla.org/show_bug.cgi?id=1536536

Deployment to production should happen Apr. 03 or so.

psiinon:
OK no worries ;)

PR #1331, opened 2019-04-25
bug: Fix all_channels to match registration UAID format

### LINKED ISSUE #1330: db Message.all_channels() does not use properly formed uaid
the channel scan function does not properly encode the string representation of the UUID to the same format as the key is stored. This results in no channels being returned for valid UAIDs.

jrconlin:
* Fix tests to match how UAID is presented to functions.

Closes #1330

### DIFF
--- a/autopush/db.py
+++ b/autopush/db.py
@@ -557,7 +557,7 @@ class Message(object):
         # update that list as well using register_channel()
         result = self.table.get_item(
             Key={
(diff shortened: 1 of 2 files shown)


bbangert:
The tests don't seem to pass?

jrconlin:
Re-running the failed test. It's the same odd hangup that we normally get that causes the travis-ci/pr failures.

Issue #1383, opened 2020-05-06
Spotty push notifications delivery on iOS


eoger:
I've noticed some Push notifications sent from the FxA servers never seem to arrive on iOS on a regular basis. We've investigated on the FxA side and it looks like Autopush is returning an error (we *think* a 502) calling the Push endpoint, but we didn't get more information.

From the slack conversion on #push, some errors were logged but couldn't get serialized to a string:
`"message": "Uncaught exception in write_error: "`

jrconlin:
Will contact you offline, but I'm going to need a few things to diagnose what's going on here. 

If possible, could you provide:
Sample message content for a notification.
if this is prod or stage
potentially some APNs credentials so I can test this outside of some sandbox type application, etc.

eoger:
This is happening on prod. I've sent you my test app by email. Thanks for looking into this!

scheinercc:
Could this be related? https://github.com/mozilla-mobile/fenix/issues/8582

jrconlin:
@scheinercc I don't believe so, since the bridge system for FCM is different than for iOS, we generally detect unregistered users better. Still #1415 may help address any similar problem if it's present.

PR #1385, opened 2020-05-06
bug: do not require `encryption` header for APNs `aes128gcm` encoded messages

### LINKED ISSUE #1384: Push callback should not require the Encryption header for aes128gcm
Right now I get a 500 error using `web-push` 3.4.3, because it does not include this header for `aes128gcm` requests.

jrconlin:
## Description

Includes better logging for uncaught APNs errors, since those were
getting dropped by cyclone.

## Testing

Tests included in integration suite.

## Issue(s)

Addresses Issue #1383
Closes #1384

### DIFF
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -66,10 +66,12 @@ class BaseHandler(cyclone.web.RequestHandler):
                     client_info=self._client_info)
             else:
                 self.log.failure("Error in handler: %s" % code,
-                                    client_info=self._client_info)
+                                 client_info=self._client_info)
             self.finish()
         except Exception as ex:
-            self.log.failure("error in write_error: {} while printing {}".format(ex, kwargs))
+            self.log.failure(
+                "error in write_error: {} while printing {}".format(
+                    ex, kwargs))
 
     def authenticate_peer_cert(self):
         """Authenticate the client per the configured client_certs.
--- a/autopush/base.py
+++ b/autopush/base.py
@@ -57,16 +57,19 @@ class BaseHandler(cyclone.web.RequestHandler):
         websocket.
 
         """
-        self.set_status(code)
-        if 'exc_info' in kwargs:
-            self.log.failure(
-                format=kwargs.get('format', "Exception"),
-                failure=failure.Failure(*kwargs['exc_info']),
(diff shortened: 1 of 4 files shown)


autopush/router/apnsrouter.py
```diff
@@ -129,9 +129,18 @@ def _route(self, notification, router_data):
         }
         if notification.data:
             payload["body"] = notification.data
-            payload["con"] = notification.headers["encoding"]
-            payload["enc"] = notification.headers["encryption"]
-
+            try:
+                payload["con"] = notification.headers["encoding"]
+                if payload["con"] != "aes128gcm":
+                    payload["enc"] = notification.headers["encryption"]
```
pjenvey, line 135: Checking for "aes128gcm" here might be more accurate but the other routers don't do it: they just add an 'enc' to payload if 'encryption' is in the header. Can we keep it consistent across the routers?

autopush/base.py
```diff
@@ -57,16 +57,21 @@ def write_error(self, code, **kwargs):
...
+        except Exception as ex:
             self.log.failure(
-                format=kwargs.get('format', "Exception"),
-                failure=failure.Failure(*kwargs['exc_info']),
-                client_info=self._client_info)
-        else:
-            self.log.failure("Error in handler: %s" % code,
-                             client_info=self._client_info)
-        self.finish()
+                "error in write_error: {} while printing {}".format(
```
pjenvey, line 73: Can you include code in here too? (this "Uncaught exception in write_error: " error is really weird..)

pjenvey:
Approved if you wanna get it out ASAP but I'd like to address those couple of things

autopush/base.py
jrconlin: Will do. No idea if the exception is just not returning a value from `str(e)` or if there's some other weirdness going on, thus why I added the original data. And, yeah, it is.

autopush/router/apnsrouter.py
jrconlin: Sure. It's probably more consistent to just include the header if it's present since we have other checks in place to see if it's required.

PR #1386, opened 2020-05-07
chore: tag 1.55.2

jrconlin:
Release 1.55.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,67 @@
+<a name="1.55.2"></a>
+## 1.55.2 (2020-05-07)
+
+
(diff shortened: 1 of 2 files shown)


pjenvey:
Guess I forgot to merge release/1.55 because this changelog is inaccurate

jrconlin:
Yeah, it's weird. I updated the changelog to include the differences between the last "full" update so things should be covered. I also included the breaking `fcm_creds` info in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1636171), although I think we already made that change prior. 

Heh, and I signed off on the previous deploys, so I wasn't paying attention either. yay...

PR #1394, opened 2020-05-16
bug: Add connection timeouts for APNS

### LINKED ISSUE #1393: Add connection timeouts for HTTP2 connections to APNs
We're seeing increased connection errors after 7 days. Current theory is that our servers aren't disconnecting during idle and that's being tagged as abusive. Need to add in a reaper for idle hyper connections. 

This is currently impacting iOS service.

jrconlin:
Closes #1393

## Description

I don't believe that these connections are being closed, and eventually we wind up with old, mostly idle connections that Apple frowns upon. Also, APNS is not really "heavy use" so we can probably turn off the pre-connection pool and just connect when need be.

TODO: 
[find a better reaper](https://github.com/mozilla-services/autopush/issues/1404) than just calling out to Thread.

## Testing

There are unit tests. With debugging enabled, idle connections should time out after about 5 min. 

## Issue(s)

Closes #1393.


autopush/router/apns2.py
```diff
@@ -166,19 +178,52 @@ def send(self, router_token, payload, apns_id,
...
     def _get_connection(self):
-        try:
-            connection = self.connections.pop()
-            return connection
-        except IndexError:
+        self._in_use_connections += 1
+        if self.log:
+            self.log.debug("Got New APNS connection")
+        if self._in_use_connections > self._max_connections:
             raise RouterException(
```
Mcat12, line 189: This exception should happen before we increment `_in_use_connections`, because it currently will leave that value in an invalid state. Probably should gate the exception on `if self._in_use_connections == self._max_connections`.

autopush/router/apns2.py
```diff
@@ -36,13 +39,21 @@ class APNSException(Exception):
     pass
 
 
+class APNSConnection:
+    def __init__(self, conn):
+        self._conn = conn
+        self._last_used = time.time()
```
Mcat12, line 45: It doesn't look like `_last_use` is ever modified. Maybe that should happen in `_get_connection`?

autopush/router/apns2.py
```diff
@@ -166,19 +178,52 @@ def send(self, router_token, payload, apns_id,
...
 
     def _return_connection(self, connection):
-        self.connections.appendleft(connection)
+        self._in_use_connections -= 1
+        if self.log:
+            self.log.debug("Done with APNS connection")
+        conn = APNSConnection(connection)
+        self.connections.appendleft(conn)
+
+    def _reap(self):
```
Mcat12, line 214: This function doesn't loop, so the thread will exit after only one reap-cycle.

autopush/router/apns2.py
jrconlin: Going to add commentary around this. APNSConnection is a wrapper for "idle" connections, so `_last_used` is only set when a connection is returned to the idle pool. When a connection becomes active again, it's removed from the `collections` set. This prevents `_reap` from ever force closing an active connection.

jrconlin:
NOTE: going to ice this PR for the time being. 

After discussions with @pjenvey, this hack may work, but what's really needed is more visibility into whatever error may actually be occurring. The HTTP2 library we use is fairly robust and should be able to handle any invalid state naturally. 

Keeping this PR and branch in reserve in case it's needed.

PR #1401, opened 2020-06-22
feat: Add tags back to metrics

### LINKED ISSUE #1400: Add metric handler for grafana

jrconlin:
## Description
 feat: Add tags back to metrics.

per ops request, this switches metrics to using markus. This also
changes the DatadogMetrics library to TaggedMetrics, and becomes
the default metrics library if the config value `statsd_host` is set.

## Testing
Unit tests provided.

## Issue(s)
Closes #1400


autopush/metrics.py
```diff
@@ -79,59 +87,50 @@ def timing(self, name, duration, **kwargs):
...
+                    'statsd_namespace': namespace,
+                }}])
+        self._client = markus.get_metrics(namespace)
         self._host = hostname
         self._namespace = namespace
 
     def _prefix_name(self, name):
-        return "%s.%s" % (self._namespace, name)
+        return name
+        # return "%s.%s" % (self._namespace, name)
```
Mcat12, line 107: If we don't need this commented out code it can be removed (applies generally to the PR).

autopush/tests/test_integration.py
```diff
@@ -1547,6 +1547,7 @@ def endpoint_kwargs(self):
 
     @inlineCallbacks
     def test_client_cert_webpush(self):
+        raise SkipTest("test cert no longer valid")
```
Mcat12, line 1550: These skipped tests seem unrelated to metrics, why are they now invalid?

autopush/metrics.py
```diff
@@ -62,9 +67,12 @@ def timing(self, name, duration, **kwargs):
 class TwistedMetrics(object):
```
Mcat12, line 67: Why do these implementations of `IMetrics` not inherit `IMetrics`?

autopush/metrics.py
```diff
@@ -62,9 +67,12 @@ def timing(self, name, duration, **kwargs):
...
     """Twisted implementation of statsd output"""
     def __init__(self, statsd_host="localhost", statsd_port=8125):
-        self.client = TwistedStatsDClient.create(statsd_host, statsd_port)
+        self.client = TwistedStatsDClient(statsd_host, statsd_port)
         self._metric = Metrics(connection=self.client, namespace="autopush")
 
+    def make_tags(self, base=None, **kwargs):
+        return kwargs
```
Mcat12, line 74: This throws away the base tags, is that expected?

autopush/tests/test_integration.py
jrconlin: The associated test certificate is no longer valid which causes the test to fail. (CA MD is too weak) Since this is a "band-aid" patch, generating a new, more valid test cert was going to complicate things. I'm not sure we still have the original cert material and I was already changing enough code as is.

autopush/metrics.py
jrconlin: Heh, to be fair, I really should just remove TwistedMetrics since it's no longer active code (aside from unit testing)

autopush/metrics.py
jrconlin: good catch.

autopush/metrics.py
jrconlin: Partly kept it for sanity reasons, but agree it's not super important.

autopush/metrics.py
```diff
@@ -34,6 +31,9 @@ def __init__(self, *args, **kwargs):
     def start(self):
         """Start any connection needed for metric transmission"""
 
+    def make_tags(self, base=None, **kwargs):
```
Mcat12, line 34: I don't see any real implementations of this method, but it's used a lot. Maybe it should have a default implementation which just passes on the data as-is?

autopush/metrics.py
Mcat12: There's more commented-out code in [`test_router`](https://github.com/mozilla-services/autopush/pull/1401/files#diff-fd4bdb0e2090991d9609f72d3920c0edR609). Is that expected to be added back soon (or modified) like the skipped tests, or can it also be removed?

jrconlin: Ah, no. missed that. Thanks!

autopush/metrics.py
jrconlin: yeah, it's a bit of legacy. Originally, we had to convert the tags from a dict to a list. Thanks for spotting, because it reminded me that I needed to make sure that I did the opposite, and reverted things back to a dictionary. If we ever need to use DatadogMetrics again, we can use this to bottleneck metrics back into lists.

jrconlin:
Closing in favor of https://github.com/mozilla-services/autopush/pull/1406

PR #1402, opened 2020-06-23
bug: Add integration tests to verify that updates do not alter reg info

### LINKED ISSUE #1397: Ensure that update-token does not delete or alter existing subscriptions for UAID
Getting a report that one of the possible reasons for FxA Send Tab failures happening on android may be that 

1. FxA reports a push publish error
2. Push on android deletes the native FCM token, and requests a new one.
3. The new FCM token is sent to the Push Component via `PushManager(...).update(...)`
4. UA eventually calls `PushManager(...).verify_connection(...)` which results in the server returning no subscriptions for the user. 

We should ensure that UAID is not altered and the subscriptions are not dropped on a FCM token update. We may also need to file an additional bug against the Android Components Push Module if we think the problem might be in the hand-off.

jrconlin:
## Description

The original unit tests left questions as to whether or not an `update()` token change altered other data associated with the registered endpoint. Added integration tests to verify that the data remains consistent after a token update request is processed.

also fixed a typo, and due to a change in ssl handling, added some skip tests if an older cert is used. 

## Testing

This is a test.

## Issue(s)

Closes #1397

### DIFF
--- a/autopush/tests/test_integration.py
+++ b/autopush/tests/test_integration.py
@@ -1990,6 +1990,10 @@ class TestFCMBridgeIntegration(IntegrationBase):
 
     @inlineCallbacks
     def test_registration_update(self):
+        """Ensure that a client bridge token update does not alter other
+        required elements of the registration data.
+
+        """
         self._add_router()
         # get the senderid
         url = "{}/v1/{}/{}/registration".format(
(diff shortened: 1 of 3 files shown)


autopush/tests/test_integration.py
```diff
@@ -1977,6 +1988,42 @@ def test_registration(self):
         assert ca_data['enc'] == salt
         assert ca_data['body'] == base64url_encode(data)
 
+    @inlineCallbacks
+    def test_registration_update(self):
```
Mcat12, line 1992: Can you add comments to these tests explaining what they're testing?

PR #1405, opened 2020-06-25
bug: Add additional logging around APNS HTTP2 connectivity

jrconlin:
## Description

Issue #1393 notes, sending push messages across APNS seems to work
great after a deploy, but then degrades after a week or so. #1394 is
a possible work-around (by double-pooling connections and using a
dedicated connection terminator), but it's messy and hacky.

What's really needed is a bit more visibility into what may be
happening, and that will involve logging all APNS communication
exceptions reliably.

## Testing

No change in function, only additional logging added.

## Issue(s)

Issue #1393


autopush/router/apnsrouter.py
```diff
@@ -147,29 +147,32 @@ def _route(self, notification, router_data):
...
+                reason = "unknown"
             self.metrics.increment("notification.bridge.connection.error",
                                    tags=make_tags(self._base_tags,
                                                   application=rel_channel,
-                                                  reason="http2_error"))
-        if not success:
+                                                  reason=reason))
             raise RouterException(
-                "Server error",
+                str(e),
```
pjenvey, line 171: I know you said you wanted to add a prefix here

jrconlin: Yeah, dropped that. There's a prefix that's already appended so it became redundant. I noticed it when I was fixing the tests.

jdragojevic:
@jrconlin when / where will we be able to see the result of these changes?

pjenvey:
@jdragojevic We're aiming to deploy this next Tues. (with a small chance the following Tues depending on QA availability).

The release will add verbose APNS error logging to [sentry](https://sentry.prod.mozaws.net/operations/autopush-prod/) and fixup some of our related Grafana metrics.

This should give us more insight into potential APNS failures that we don't currently have.

tublitzed:
@jdragojevic - this went into the [1.56.0](https://github.com/mozilla-services/autopush/releases/tag/1.56.0) release which is now [live in prod](https://push.services.mozilla.com/__version__). 

@jrconlin can you add some more detailed information about where Janet and team should be looking for the changes you've made here?

jrconlin:
These errors should now be reported up to [Sentry](https://sentry.prod.mozaws.net/operations/autopush-prod/), when they start happening.  Since this seems to be an error that occurs some time after deploy, i expect there to be no unusual errors until at least Jul 10. Sentry notifies us on new error types.

jrconlin:
We expect that after 1.56.1 goes to prod, that the metric tagging will be fixed for autopush endpoint and that our metric reporting will improve greatly. We expect 1.56.1 to go to production in the next day or two, depending on ops load.

PR #1406, opened 2020-06-26
feat: Add tags back to metrics.

### LINKED ISSUE #1400: Add metric handler for grafana

jrconlin:
## Description

While this package does not use the officially sanctioned "markus"
metrics package, it does use the same Datadog library.

*Reviewer Note*: this replaces #1401, which does a lot more (replaces metrics libraries with `markus`, changes base tag handling, etc.) This PR carries less risk for a service that's being rewritten anyway.

## Testing

Unit tests included.

## Issue(s)

Closes #1400

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -7,7 +7,7 @@ from typing import (  # noqa
 )
 
 from twisted.internet import reactor
-from txstatsd.metrics.metrics import Metrics
+from txstatsd.metrics.metrics import Metrics  # noqa
 
 import mock
 import datadog
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -10,6 +10,7 @@ from twisted.internet import reactor
 from txstatsd.client import StatsDClientProtocol, TwistedStatsDClient
 from txstatsd.metrics.metrics import Metrics
 
+import mock
 import datadog
 from datadog import ThreadStats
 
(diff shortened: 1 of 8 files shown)


autopush/metrics.py
```diff
@@ -7,9 +7,9 @@
 )
 
 from twisted.internet import reactor
-from txstatsd.client import StatsDClientProtocol, TwistedStatsDClient
-from txstatsd.metrics.metrics import Metrics
+from txstatsd.metrics.metrics import Metrics  # noqa
```
pjenvey, line 10: nit: we can kill this and change the one import for it in `autopush/web/webpush.py` to `autopush.metrics.IMetrics`, then remove the txstatsd dep entirely

autopush/metrics.py
```diff
@@ -122,16 +108,16 @@ def timing(self, name, duration, **kwargs):
 def from_config(conf):
     # type: (AutopushConfig) -> IMetrics
     """Create an IMetrics from the given config"""
-    if conf.datadog_api_key:
+    if conf.statsd_host and not isinstance(conf.statsd_host, mock.Mock):
```
pjenvey, line 111: nit: I think my branch got away without this Mock check with this:

```
diff --git a/autopush/tests/test_db.py b/autopush/tests/test_db.py
index e814f69..2206d45 100644
--- a/autopush/tests/test_db.py
+++ b/autopush/tests/test_db.py
@@ -91,6 +91,7 @@ class DatabaseManagerTest(unittest.TestCase):
         fake_conf.message_table.tablename = "message_int_test"
         fake_conf.message_table.read_throughput = 5
         fake_conf.message_table.write_throughput = 5
+        fake_conf.statsd_port = 8125
         dm = DatabaseManager.from_config(
             fake_conf,
             resource=autopush.tests.boto_resource)
@@ -107,6 +108,7 @@ class DatabaseManagerTest(unittest.TestCase):
         fake_conf.message_table.tablename = "message_bogus"
         fake_conf.message_table.read_throughput = 5
         fake_conf.message_table.write_throughput = 5
+        fake_conf.statsd_port = 8125
         dm = DatabaseManager.from_config(
             fake_conf,
             resource=autopush.tests.boto_resource)
```

jrconlin: Huh, perhaps. I do the mock check since I wasn't sure how many tests might not set that value.

jrconlin: a better (?) solution might be to have a default `fake_conf` that we alter for local testing.

PR #1407, opened 2020-06-27
feat: fixup metrics + tags

### LINKED ISSUE #1400: Add metric handler for grafana

pjenvey:
## Description

remove datadog's ThreadStats (as it only speaks datadog's http protocol)
w/ markus backed by the plain statsd compatible Datadog backend

upgrade datadog to avoid its own telemetry metrics producing garbage
tags

Note:

This is simplified from #1401 and has a couple other fixes:

- markus is actually expecting the original datadog style tags arguments we currently use e.g. `['foo:bar', 'baz:quux']`, and the PR had a few straggler uses of the old style anyway, so kept what we have
- fixes passes them along via the `tags=` kwarg it expects
- doesn't pass the namespace option to Datadog (as markus inserts it for us)

## Testing

Running against a local statsd collector produces expected metrics + tags

## Issue(s)

Closes #1400

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -8,8 +8,7 @@ from typing import (  # noqa
 
 from twisted.internet import reactor
 
-import datadog
-from datadog import ThreadStats
+import markus
 
 from autopush import logging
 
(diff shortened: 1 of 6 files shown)


sciurus:
LGTM!

Issue #1408, opened 2020-07-01
Defer metric calls to improve latency


jrconlin:
We noticed a sharp update in latency and CPU with 1.56.1. It was suggested that we may benefit by putting markus metric recording into a deferred handler in order to reduce latency.

pjenvey:
Reopening: there's a bug in the PR: we should be calling `deferToThread` with a function and arguments, this passes the result of the call

PR #1410, opened 2020-07-07
bug: regression: wrap metric calls in thread handler

jrconlin:
## Description
The metric calls were blocking. It was recommended that they be wrapped
by a discrete thread handler.

## Testing

How should reviewers test?

## Issue(s)

Issue: #1408

### DIFF
--- a/autopush/metrics.py
+++ b/autopush/metrics.py
@@ -6,7 +6,7 @@ from typing import (  # noqa
     Sequence
 )
 
-from twisted.internet import (reactor, threads)
+from twisted.internet import reactor
 
 import markus
 
@@ -92,7 +92,7 @@ class TaggedMetrics(IMetrics):
         return tags
 
     def increment(self, name, count=1, tags=None, **kwargs):
-        threads.deferToThread(
+        reactor.callInThread(
             self._client.incr,
             self._prefix_name(name),
             count,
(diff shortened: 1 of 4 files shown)


autopush/metrics.py
```diff
@@ -92,21 +92,27 @@ def _make_tags(self, tags):
         return tags
 
     def increment(self, name, count=1, tags=None, **kwargs):
-        threads.deferToThread(
-            self._client.incr(self._prefix_name(name), count,
-                              tags=self._make_tags(tags))
+        reactor.callInThread(
```
Mcat12, line 95: Why is `deferToThread` replaced with `callInThread`? `deferToThread` can be used to call a method in the threadpool according to its docs: https://twistedmatrix.com/documents/13.1.0/api/twisted.internet.threads.deferToThread.html

fzzzy: Well, on inspection of the source, they both look like they do almost exactly the same thing. However, callInThread has no return value, while deferToThread returns a Deferred (basically a Promise or Future) which allows retrieving the called-in-thread function's return value.

In this case we don't need the return value, since we're just doing a fire and forget on some metrics. So we might as well not incur the gc overhead of creating/destroying those Deferreds since there's no return value from the metrics function.

jrconlin: Pretty much what @fzzzy said. Plus `callInThread` also allowed more tests to pass without odd timing failures.

PR #1420, opened 2020-07-21
bug: fix header_cleanup for aesgcm

### LINKED ISSUE #1419: Clean up padding and quotes in headers
Client reports issues with quotes and padding appearing in values passed from Encryption and Crypto-Key headers. 

These are legacy "aesgcm" encryption formats that are still in use. Server should probably clean these values up (remove quotes and padding) before relaying.

Client may also take steps to resolve this issue. 

image

jrconlin:
## Description

* Remove embedded " and = from Encryption and Crypto-Key header values

## Testing

testing included. 
## Issue(s)

Closes #1419

### DIFF
--- a/autopush/utils.py
+++ b/autopush/utils.py
@@ -375,9 +375,11 @@ class WebPushNotification(object):
         headers = self.headers
         # Strip crypto/encryption headers down
         for hdr in ["crypto-key", "encryption"]:
-            if STRIP_PADDING.search(headers.get(hdr, "")):
+            if hdr in headers:
                 head = headers[hdr].replace('"', '')
-                headers[hdr] = STRIP_PADDING.sub("", head)
+                head = STRIP_PADDING.sub("", head)
+                head.decode('ascii')
+                headers[hdr] = head
 
         # content-encoding header may already be stored as "encoding",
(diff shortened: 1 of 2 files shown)


autopush/utils.py
```diff
@@ -375,9 +375,11 @@ def cleanup_headers(self):
         headers = self.headers
         # Strip crypto/encryption headers down
         for hdr in ["crypto-key", "encryption"]:
-            if STRIP_PADDING.search(headers.get(hdr, "")):
+            if hdr in headers:
                 head = headers[hdr].replace('"', '')
-                headers[hdr] = STRIP_PADDING.sub("", head)
+                head = STRIP_PADDING.sub("", head)
+                head.decode('ascii')
```
Mcat12, line 381: `decode` returns a string (it doesn't mutate), but the output is not used here?

jrconlin: It's more a check to see if the header response is ascii and bailing if it's not. While it's very unlikely that anyone were to send a header that is `salt="💩"`, this would catch it. I'm also fine dropping it, but I tend to be paranoid and if we're doing checks to safeguard the client, figured I might as well add it.

PR #1423, opened 2020-07-28
feat: add metric for all mobile bridge "unregistered" user drops

### LINKED ISSUE #1421: Emit detailed metrics for new drop_user on bridge 404s
We should emit a metric with detailed tag information about the bridge and its response code that caused the drop.

jrconlin:
* normalized metric name and tags

Closes #1421

## Description
* made the APNS user disconnect metric more generic across all mobile platforms
* normalized some tag information

## Testing

Tests included. 

## Issue(s)

Closes #1421.

### DIFF
--- a/autopush/router/adm.py
+++ b/autopush/router/adm.py
@@ -191,7 +191,7 @@ class ADMRouter(object):
             self.metrics.increment("notification.bridge.error",
                                    tags=make_tags(
                                        self._base_tags,
-                                       reason="connection_unavailable"))
+                                       reason="connection unavailable"))
             raise RouterException("Server error", status_code=502,
                                   errno=902,
                                   log_exception=False)
--- a/autopush/web/base.py
+++ b/autopush/web/base.py
@@ -335,13 +335,13 @@ class BaseWebHandler(BaseHandler):
         # reporting a not found and they're mobile.
         if exc.status_code in [404, 410] and router_type in [
                 'apns', 'fcm', 'adm']:
-            tags = self._base_tags.update({
+            self._base_tags.update({
                 "platform": router_type,
                 "reason": "unregistered",
             })
             self.metrics.increment(
                     "notification.bridge.error",
-                    tags=tags)
+                    tags=self._base_tags)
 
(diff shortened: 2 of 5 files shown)


autopush/web/base.py
```diff
@@ -335,6 +335,14 @@ def _router_fail_err(self, fail, router_type=None, vapid=False, uaid=None):
         # reporting a not found and they're mobile.
         if exc.status_code in [404, 410] and router_type in [
                 'apns', 'fcm', 'adm']:
+            tags = self._base_tags.update({
```
Mcat12, line 338: `update` mutates the dictionary and returns `None`:
https://docs.python.org/3/library/stdtypes.html#dict.update

This is probably not the behavior you're expecting?

jrconlin: crap. You're right. I always get that wrong. Thanks!

autopush/web/base.py
```diff
@@ -335,6 +335,14 @@ def _router_fail_err(self, fail, router_type=None, vapid=False, uaid=None):
         # reporting a not found and they're mobile.
         if exc.status_code in [404, 410] and router_type in [
                 'apns', 'fcm', 'adm']:
+            self._base_tags.update({
+                "platform": router_type,
+                "reason": "unregistered",
+            })
```
Mcat12, line 341: I'm concerned that mutating the base tags here might leak into other usages. Are these tags reused after this point, or are they recreated?

jrconlin: It's part of `_router_fail_err` which is an error handler and is one of the last stages for the request Handler.
Mutating the base tags here is fine, and expected.

autopush/router/adm.py
```diff
@@ -191,7 +191,7 @@ def _route(self, notification, uaid_data):
             self.metrics.increment("notification.bridge.error",
                                    tags=make_tags(
                                        self._base_tags,
-                                       reason="connection_unavailable"))
+                                       reason="connection unavailable"))
```
pjenvey, line 194: FYI this *denormalized* the reason, it's an underscore everywhere else

Issue #1373, opened 2020-03-30
RouterException: Server error: INVALID_ARGUMENT: Request contains an invalid argument.


jrconlin:
Staring since around Mar. 16

https://sentry.prod.mozaws.net/operations/autopush-prod/issues/6072684/

```
RouterException: Server error: INVALID_ARGUMENT: Request contains an invalid argument.
  File "twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "autopush/router/fcmv1client.py", line 129, in error
    raise RouterException("Server error: {}".format(failure.value))

RouterException: Server error: INVALID_ARGUMENT: Request contains an invalid argument.
```

jrconlin:
Not really seeing where the error might be from [the docs](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages)

Mcat12:
The response should contain more information about what was invalid: https://firebase.google.com/docs/reference/fcm/rest/v1/ErrorCode#ENUM_VALUES.INVALID_ARGUMENT

jrconlin:
Right, but the question is whether or not the error handler for `treq` allows the response body to be read. I think so, but I'd have to play around a bit to make sure.

Adrianotiger:
I have the same error message, when I try to send a push notification to a mobile device from a php-script. 
It works on Firefox for desktop, but for Firefox on mobile (it works on Edge and Chrome), I get this message:

_Server error: INVALID_ARGUMENT: Request contains an invalid argument._

"endpoint": https://updates.push.services.mozilla.com/wpush/v1/gAAAAABfgDHFI3uKHq48tJLjFHD1Osft2o_4YSJgHeswI1XAPVODFc-ZYlqj2wm...
"code": 500, 
"errno": 999, 
"error": "Internal Server Error", 
"more_info": "http://autopush.readthedocs.io/en/latest/http.html#error-codes"

How should I know what is wrong, if the answer is a errno999?

Mcat12:
@Adrianotiger The cause of this error was due to message size limitations. Autopush was allowing larger message sizes than what Firebase would allow. Check the size of your messages, though that error is generic enough that you may have a different problem. The fix for this specific issue is targeted for release 1.58.2: https://github.com/mozilla-services/autopush/pull/1437

jrconlin:
Thanks Mark!

Yeah, the error is fairly generic and we're going to be fixing that
[...]

PR #1430, opened 2020-08-07
bug: report details of an FCM "INVALID_ARGUMENT" error

jrconlin:
Issue #1373

## Description

This is a really quick patch to try and capture the relevant details returned from FCM when it returns an INVALID_ARGUMENT error. I had thought it was related to the TTL, but we already trim it down to an appropriate max value. 

## Testing

Honestly, if I could figure out how to trigger this in dev, I'd be thrilled. As it is, to test this you'll need a valid FCM routing number, which means compiling an emulated FCM client app (the sample one works fine) which means updating Android Studio to 5, and fetching the latest updates, before altering the code paths and java code to match the FCM project you declared so that the downloaded google-services.json file tied to the sample app you declared in FCM match then use a combination of registration and send scripts to try to trigger an INVALID_ARGUMENT error that gets dumped out to sentry.

All for a patch to see what's going on, and not the real fix.

## Issue(s)

Issue #1373

### DIFF
--- a/autopush/router/fcm_v1.py
+++ b/autopush/router/fcm_v1.py
@@ -1,4 +1,7 @@
 """FCM v1 HTTP Router"""
+import json
+from math import ceil
+
 from typing import Any  # noqa
 
 from twisted.internet.error import ConnectError, TimeoutError
@@ -14,6 +17,13 @@ from autopush.router.fcmv1client import (
     FCMNotFoundError
 )
 from autopush.types import JSONDict  # noqa
+
+# the universal default for this is 4096, which is far too
+# large for FCM. The final payload size of the encoded data, plus
+# encryption headers must fit in the 4096 byte FCM payload.
+# Since the body is re-encoded base64, we reduce the message
+# size accordingly
+MAX_FCM_DATA = 3015
 
 
 class FCMv1Router(FCMRouter):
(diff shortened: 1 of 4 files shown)


Mcat12:
I have the necessary testing setup ready, I'll try to reproduce the issue with this PR.

Mcat12:
I think the error is due to incorrectly accounting for FCM's (total) message size limit, a la https://github.com/mozilla-services/autopush-rs/issues/203.

I was able to trigger the error by passing through a message which _just_ passed under our size validation but when wrapped in the FCM payload was over 4096 bytes.

jrconlin:
Huh, that explains why we're not seeing additional details. Interesting. I'll try a few test runs as well and see if I can duplicate. Thanks!

autopush/router/fcm_v1.py
```diff
@@ -15,6 +18,13 @@
 )
 from autopush.types import JSONDict  # noqa
 
+# the universal default for this is 4096, which is far too
+# large for FCM. The final payload size of the encoded data, plus
+# encryption headers must fit in the 4096 byte FCM payload.
+# Since the body is re-encoded base64, we reduce the message
+# size accordingly
+MAX_FCM_DATA = 3015
```
Mcat12, line 26: It is more straightforward to implement the max data check in the FCM client instead of the router (See https://github.com/mozilla-services/autopush-rs/pull/212). The client has the full payload available. Then there's no need to use any limit other than 4096.

jrconlin: I debated that, but part of the thing was I wanted to be helpful to the original sender. To do that, I'd need to have the original message data. I could deconstruct it from inside the sender client, or I could do it while I was composing the message to send through. I figured it was clearer to do it while composing. 

I could see an argument about putting the final message size check inside of the sender, and check the length of `payload.data_message` but it's just as straight forward to do it in the caller, and doesn't include a stack shift.

Mcat12: It's not very clear, but from testing it seems that FCM only considers the `data` field when calculating the size of the payload, so this code is accurate enough.

PR #1435, opened 2020-10-02
bug: compare VAPID aud to endpoint_hostname 

### LINKED ISSUE #1434: Have the vapid `aud` check compare against endpoint_hostname
while `hostname` and `endpoint_hostname` are usually the same, it's a better idea to check against `endpoint_hostname`.

jrconlin:
## Description

use `endpoint_url` instead of `hostname` for VAPID aud checks.

## Testing

Provide a different `endpoint_hostname` instead of `hostname` (e.g. `localhost` instead of `127.0.01`) and verify that only audiences using the matching host name are allowed.

## Issue(s)

Closes #1434.

### DIFF
--- a/autopush/web/webpush.py
+++ b/autopush/web/webpush.py
@@ -432,9 +432,10 @@ class WebPushRequestSchema(Schema):
         if jwt['aud'] != "{}://{}".format(
                 self.context["conf"].endpoint_scheme or "http",
                 self.context["conf"].hostname):
-            raise InvalidRequest("Invalid bearer token: Invalid Audience Specified",
-                                 status_code=401, errno=109,
-                                 headers={"www-authenticate": PREF_SCHEME})
+            raise InvalidRequest(
+                "Invalid bearer token: Invalid Audience Specified",
+                status_code=401, errno=109,
+                headers={"www-authenticate": PREF_SCHEME})
         if "exp" not in jwt:
             raise InvalidRequest("Invalid bearer token: No expiration",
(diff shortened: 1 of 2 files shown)


autopush/web/webpush.py
```diff
@@ -431,7 +431,7 @@ def validate_auth(self, d):
                                  headers={"www-authenticate": PREF_SCHEME})
         if jwt['aud'] != "{}://{}".format(
                 self.context["conf"].endpoint_scheme or "http",
```
pjenvey, line 433: Oh `endpoint_url` might be more correct here vs endpoint_scheme + hostname. Assuming `aud` would naturally come from the subscription info created from `make_endpoint` (which uses `endpoint_url`). endpoint_url also includes the port (on non standard ports)

jrconlin: yeah, that's the trickier part. It doesn't. "aud" comes from whoever is creating the VAPID signature, so it should match whatever the publicly visible push endpoint would be, non-standard port and all. It also should only be the scheme, host (and non-standard port) so any additional path stuff needs to be dropped.

pjenvey: Sure it comes from whoever but they would *likely* look there e.g. https://developers.google.com/web/updates/2016/07/web-push-interop-wins#payload and `endpoint_url` seem to be exactly this (I think).

pjenvey (dismissed):
just need to update the commit message (endpoint_hostname -> url)

pjenvey (approved):

PR #1462, opened 2022-06-27
bug: Fix metric calls

jrconlin:
* Be explicit about args
* handle AttributeError gracefully

Closes: #CONSVC-1842

### DIFF
--- a/autopush/router/apnsrouter.py
+++ b/autopush/router/apnsrouter.py
@@ -163,7 +163,10 @@ class APNSRouter(object):
             elif isinstance(e, (HTTP20Error, socket.error)):
                 reason = "http2_error"
             else:
-                reason = e.extra.get("reason", "unknown")
+                try:
+                    reason = e.extra.get("reason", "unknown")
+                except AttributeError:
+                    reason = "unknown"
             if isinstance(e, RouterException) and e.status_code in [404, 410]:
                 raise RouterException(
                     str(e),


autopush/router/apnsrouter.py
```diff
@@ -163,7 +163,10 @@ def _route(self, notification, router_data):
             elif isinstance(e, (HTTP20Error, socket.error)):
                 reason = "http2_error"
             else:
-                reason = e.extra.get("reason", "unknown")
+                try:
+                    reason = e.extra.get("reason", "unknown")
+                except AttributeError:
+                    reason = format("Unknown: {}", e)
```
pjenvey, line 169: `reason` here is a metrics tag value which shouldn't have whitespace:
```suggestion
                    reason = "unknown"
```

Sentry's complaining about a `hyper.common.exceptions.ConnectionResetError`,  could include it alongside `ConnectionError` handling above -- `if isinstance(e, (ConnectionError, ConnectionResetError)):` but probably not even worth the effort.

pjenvey (changes requested):

pjenvey (changes requested):
The last update removed the fix for the [AttributeError](https://sentry.io/organizations/mozilla/issues/3368428331/?project=6326250&query=is%3Aunresolved) entirely

jrconlin:
I hate "helpful" merges.

pjenvey (approved):

